PR #38, opened 2026-06-28
Collapsible "All Transcriptions" sidebar section (fixes #28)

### LINKED ISSUE #28: Collapsible sidebar: show recordings under 'All Transcriptions' with title + timestamp
**Problem / use case**
The sidebar currently has a single `All Transcriptions` item. With many recordings it requires navigating to the full list and scanning to find a specific meeting - there is no way to jump directly to a recording from the sidebar without first landing on the list page.

**Proposed solution**
Make `All Transcriptions` a collapsible section in the sidebar:

- A small disclosure triangle (▶ / ▾) to the left of the label expands/collapses the list.
- When expanded, each recording appears as a sub-row with its **title as the primary text** and the **timestamp in smaller gray secondary text** below it.
- Clicking a sub-row navigates directly to that recording's transcript page.
- Clicking `All Transcriptions` itself (the header row) still navigates to the current all-recordings list page, same as today.
- The expanded/collapsed state persists across app launches (UserDefaults).

Rough sketch of the expanded sidebar:

```
▾ All Transcriptions
    Team standup - Jun 23
    Jun 23, 09:14  <- smaller, gray
    -----------------
    Product review
    Jun 22, 15:30
    ...
```

This mirrors the navigation pattern used in Notes, Mail, and other macOS apps where a source-list group can be drilled into inline.

**Alternatives considered**
- A search/filter bar on the All Transcriptions page - helps discovery but doesn't reduce the extra navigation step.
- A separate Recents section in the sidebar - adds clutter and duplicates entries.

**Additional context**
I would be happy to contribute this feature if the maintainers are open to it. Main touch points: `SidebarView` (disclosure group + sub-rows) and the navigation/selection state model that drives the current view.

urisland:
## What & why

Implements #28. The sidebar had a single `All Transcriptions` item; with many recordings you had to land on the full list and scan it to find a meeting. This makes `All Transcriptions` a **collapsible section** so you can drill in and jump straight to a recording from the sidebar.

## Changes (`Mila/Views/SidebarView.swift`)

- **Disclosure triangle (▶/▾)** to the left of the label toggles an inline list of the unfiled recordings. It's a `.plain` Button, so it flips expansion *independently* of List selection.
- **Header still navigates** to the all-recordings page — the row keeps its `.defaultFolder` tag and its drag-to-unfile drop target, exactly as before.
- **Sub-rows**: each recording shows its **title** as primary text with a smaller, **gray timestamp** beneath it. Tagged `.recording(id)` so a click navigates straight to the transcript page; also `.draggable` so it can still be filed into a folder.
- **Persisted state**: expanded/collapsed survives app launches via `@AppStorage("sidebar.allTranscriptions.expanded")`.

The recordings shown match what the header navigates to (`store.unfiledRecordings()`), so the inline list and the destination page stay consistent.

## Testing

- `make build` succeeds.
- Ran the app directly (stable, no crashes) and with a clean seeded store **pre-expanded** so a sub-row renders — the expanded code path is exercised cleanly.

## Summary by CodeRabbit

* **New Features**
  * Added a collapsible **All Transcriptions** section in the sidebar.
  * The expanded/collapsed state is now remembered between app launches.
  * When expanded, unfiled recordings appear inline with quick access and drag-and-drop support.
  * The section header now keeps selection behavior while letting the chevron control expansion independently.

### DIFF
--- a/Mila/Views/SidebarView.swift
+++ b/Mila/Views/SidebarView.swift
@@ -41,6 +41,12 @@ struct SidebarView: View {
     @State private var renameTarget: String?
     @State private var renameDraft = ""
 
+    /// Whether the "All Transcriptions" section is expanded to show its
+    /// recordings inline. Persisted so the user's choice survives app
+    /// launches (issue #28).
+    @AppStorage("sidebar.allTranscriptions.expanded")
+    private var allTranscriptionsExpanded = false
+
     var body: some View {
         List(selection: $selection) {
             Section {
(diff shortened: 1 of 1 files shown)


urisland:
@coderabbitai full review

PR #44, opened 2026-06-28
Add project skills: uninstall-mila + build-mila-locally

leetwito:
Adds two project-level Claude Code skills under `.claude/skills/` documenting common local dev operations for Mila.

## Skills

- **`build-mila-locally`** — debug/release builds, the self-signed (ad-hoc) DMG flow, and installing a local build into `/Applications`. Captures real build gotchas: `make dmg` needing an explicit `VERSION=` (Info.plist stores the literal `$(MARKETING_VERSION)`), the gitignored `PythonRuntime/` placeholder, and where whisper models live.
- **`uninstall-mila`** — safe removal of the app, Dock tile, and system caches/prefs (filed under bundle id `io.island.whisper.IslandWhisper`, not "Mila"), while protecting `~/Library/Application Support/Mila/Recordings/` + `recordings.json`. Backs up recordings first, uses `trash` (reversible), and requires explicit user confirmation before deleting the ~7.2 GB regenerable Models/torch data.

## Notes

- Documentation only — no code or build changes.
- Both verified this session: build → DMG → install → launch, and a full uninstall round-trip.

## Summary by CodeRabbit

* **Documentation**
  * Added a macOS “build from source” guide with quick-reference `make` commands (project generation, debug/release builds, `run`, tests, DMG creation, cleanup), plus troubleshooting for local DMG signing and Gatekeeper/quarantine behavior. Notes `make dmg` requires explicitly providing `VERSION=`.
  * Added a safe macOS uninstall guide that removes the app and bundle-id–scoped caches/build artifacts while protecting recordings/transcripts by default. Includes an optional full reset, removes the Dock tile, and provides verification steps.

### DIFF
--- a/.claude/skills/build-mila-locally/SKILL.md
+++ b/.claude/skills/build-mila-locally/SKILL.md
@@ -21,7 +21,7 @@ All build commands run from the repo root (wherever your checkout lives).
 | Release build | `make release-build` | `build-release/Build/Products/Release/Mila.app` |
 | Self-signed DMG | `make dmg VERSION=<x.y.z>` | `Mila-<x.y.z>.dmg` (ad-hoc signed) |
 | Run tests | `make test` | XCTest run |
-| Clean | `make clean` | removes `.xcodeproj`, `build`, `build-release`, `*.dmg` |
+| Clean | `make clean` | removes `Mila.xcodeproj`, `build`, `build-release`, `*.dmg` |
 
 `make` targets chain: `dmg → release-build → project → bootstrap`, so a single command regenerates the project and builds.
 
--- a/.claude/skills/build-mila-locally/SKILL.md
+++ b/.claude/skills/build-mila-locally/SKILL.md
@@ -28,15 +28,18 @@ All build commands run from the repo root (`~/ClonedProjects/mila`).
 ## CRITICAL: `make dmg` needs an explicit VERSION
 
 `make dmg` **fails on a clean checkout** with:
-```
+
+```text
 /bin/sh: MARKETING_VERSION: command not found
 ./scripts/make-dmg.sh: ... usage: make-dmg.sh APP_PATH DMG_PATH VERSION
 ```
-**Why:** `Info.plist` stores the literal `$(MARKETING_VERSION)` (resolved only at Xcode build time). The Makefile's version fallback reads `Info.plist` with `PlistBuddy`, gets the literal string back, and the shell tries to *execute* `MARKETING_VERSION`. The build itself SUCCEEDS — only DMG packaging breaks.
 
-**Fix:** pass the real version (it lives in `project.yml`) explicitly:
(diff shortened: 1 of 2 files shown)


urisland:
Addressed both CodeRabbit findings in b543d29:

- **`build-mila-locally`** — Rewrote the `make dmg` troubleshooting: the root cause is that the Makefile's `VERSION` fallback resolves to the literal `$(MARKETING_VERSION)`, which the shell tries to command-substitute (the `MARKETING_VERSION: command not found` line) and `VERSION` collapses to empty, so `make-dmg.sh` aborts on its `${3:?usage…}` guard. Replaced the brittle `grep | sed` extraction with an `awk` read of the canonical `MARKETING_VERSION` in `project.yml`, and added language tags + blank lines around the fences (markdownlint).
- **`uninstall-mila`** — `RecordingStore.swift` persists `folders.json` as a sibling of `recordings.json`, so the backup now includes it (Iron Rule, path map, procedure, verify, and Common Mistakes all updated). The `folders.json` copy is optional (`|| true`) since it may not exist yet, while the `recordings.json` copy stays strict.

Verified the remaining claims against the tree (all `make` targets, the bundle id `io.island.whisper.IslandWhisper`, and the `Models/` / `torch-site-packages/` paths exist).

@coderabbitai review

urisland:
Addressed both re-review comments in 3a6027b:

- **`build-mila-locally`** — dropped the hardcoded `~/ClonedProjects/mila`; the prose now just says the commands run from the repo root, wherever the checkout lives.
- **`uninstall-mila`** — backup step is now fail-fast: `set -euo pipefail`, strict `cp` for `Recordings/` + `recordings.json`, `folders.json` guarded with `[[ -f ]]` (optional), and the verify step asserts the index (and any existing folder map) actually copied — so an incomplete backup aborts before anything is deleted.

@coderabbitai review

urisland:
Addressed the round-3 nitpicks in e8088a3:

- **`build-mila-locally`** — the clean row now names the concrete `Mila.xcodeproj`.
- **`uninstall-mila`** — the repo `build/` cleanup is now path-agnostic (a `REPO` variable you set to your checkout, plus generic prose in the Iron Rule and Common Mistakes — I didn't use the literal `/build` suggestion since that points at the filesystem root). The verify block now checks the full bundle-id footprint (`HTTPStorages` + `WebKit` added alongside `Preferences`/`Caches`).

@coderabbitai review

urisland:
Addressed round-4 in c9f4a3e:

- **`build-mila-locally`** — added an explicit non-empty guard on `VERSION` (clear error + exit if `MARKETING_VERSION` isn't found) before `make dmg`.
- **`uninstall-mila`** — repo `build/` path is now portable via a `MILA_REPO` override, and I gate the `trash` on `project.yml` existing in `REPO` so a wrong path can't nuke an unrelated `build/`. I deliberately didn't auto-derive from `git rev-parse`/`pwd` — during an uninstall the user often isn't standing in the Mila checkout, so that could silently resolve to the wrong repo; an explicit override with a sanity check is safer.

@coderabbitai review

urisland:
@coderabbitai review

urisland:
Addressed the prefs-verification comment: the verify block now checks the removable footprint (Caches/HTTPStorages/WebKit) separately from the kept-by-default Preferences plist, labeling the plist as kept (routine) vs removed (full reset) so a routine uninstall no longer reads as a failed cleanup.

Also resolved the earlier "repo path portable" thread — that was fixed in c9f4a3e via the `MILA_REPO` override + `project.yml` sanity check.

@coderabbitai review

PR #54, opened 2026-07-01
Add file logging + MilaLog so a Finder-launched app is debuggable

leetwito:
## Why

Diagnosing failures in a **Finder/`open`-launched Mila** was effectively impossible: GUI apps discard `stdout`/`stderr`, so none of the app's `print(...)` transcription diagnostics survived, and `os.Logger` subsystem entries were only reachable via Console.app after the fact. This started while trying to debug a real issue and I kept having no logs to look at - so this adds a persistent file log.

## What

- **`redirectMilaLogsToFile()`** - first line of `MilaApp.init()`. Redirects `stdout`/`stderr` to `~/Library/Logs/Mila/mila.log`. **Skipped when attached to a TTY** (`isatty(STDERR_FILENO) == 0`) so `make run`/Xcode keep printing to the console. ~5 MB rotation.
- **`MilaLog`** - a drop-in `os.Logger` replacement that mirrors every entry to **both** the unified log **and** stdout, so subsystem logs (VoiceMemos, ModelManager, RemoteWhisperEngine, MeetingDetector, ...) land in the file too. Same interpolation syntax including `, privacy: .public`, so existing call sites compile unchanged. Migrated the existing `os.Logger` sites across the app.
- **Makefile** - `make reinstall` (release-build -> replace `/Applications/Mila.app` -> `open` so file logging engages) and `make logs` (tail the log).
- **CLAUDE.md** - documents the logging setup, scope, and gotchas.

## Testing status

Builds cleanly (`make build` / `make release-build`) against current `main`.

**Verified E2E:** after `make reinstall` + a Finder launch, `~/Library/Logs/Mila/mila.log` captured live output from both sinks - `print(...)` lines (WhisperEngine, TranscriptionService) **and** `MilaLog` subsystem lines (`[MeetingDetector] NOTICE ...`, `[LiveTranscriber] ...`, `[RecordingSummarizer] ...`). Core capture + the TTY gate work.

⚠️ **Still unverified:** the ~5 MB log rotation path and `.private`/`.sensitive` redaction in the file sink. Keeping this a **draft** until those are exercised - feedback on the approach welcome in the meantime.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -50,6 +50,30 @@ These patches live in `SpeakerDiarizer.swift`'s inline diarize script. If upgrad
 - `TranscriptionService` now requires a `diarizationSettings:` parameter. In tests, always pass `DiarizationSettings(defaults: .init(suiteName: "TestClassName.diarization")!)` to isolate from user defaults.
 - Run tests with `make test` or via Xcode.
 
+### Logging
+- **Where:** the app writes a plain-text log to `~/Library/Logs/Mila/mila.log`.
+  Tail it with `make logs`. It is truncated when it grows past ~5 MB.
+- **When it's active:** file logging is set up by `redirectMilaLogsToFile()` (the
+  first line of `MilaApp.init()`), which redirects `stdout`/`stderr` to the file.
+  It is **skipped when the app is attached to a TTY** (`guard isatty(STDERR_FILENO) == 0`)
+  so that running under Xcode or a terminal still prints to the console. Practical
+  upshot: a **Finder/`open` launch logs to the file** (use `make reinstall`), but a
+  terminal launch (e.g. `make run`, or running the binary directly) does **not** —
+  it prints to the console instead.
+- **Scope (what lands in the file):**
+  1. Every `print(...)` in the app (most transcription-pipeline diagnostics).
+  2. Every `MilaLog` call. `MilaLog` (see `Mila/App/MilaLog.swift`) is the
+     app-wide logger that replaced raw `os.Logger`; it mirrors each entry to
+     **both** the unified log (`os.Logger`, subsystem `io.island.whisper.IslandWhisper`)
+     **and** stdout, so subsystem logs (VoiceMemos, ModelManager, MeetingDetector,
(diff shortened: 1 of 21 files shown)


urisland:
Closing as stale — but the problem it solves is real, so here is what a follow-up would need, in case you or anyone else picks it up.

**Why it is being closed:** still a draft, unchanged since 1 July, one failing check (`unit-and-ui-tests`), and it now conflicts with `main` after a month of movement across the 21 files it touches. Rebasing and re-reviewing it costs more than rewriting the useful half.

**What was genuinely good here:** GUI-launched Mila discarding `stdout`/`stderr` really does make Finder-launched failures hard to diagnose, the TTY check so terminal runs are unaffected is the right instinct, and `MilaLogMessage` does honour explicit `privacy: .private` / `.sensitive` by rendering `<private>`.

**The blocking design issue, if it comes back.** `MilaLogMessage.Interpolation.appendInterpolation<T>(_ value: T)` — the overload with no privacy argument — renders the value in full. `os.Logger` does the opposite: for a non-literal interpolation the default is **private**, and the value is redacted in the unified log unless you opt into `.public`. So swapping a call site from `Logger` to `MilaLog` silently *un-redacts* every interpolation that did not name a privacy level. That leaks in two directions at once, because `MilaLog` also re-logs the rendered string to the unified log as `privacy: .public` — so the redaction is lost even for someone reading `log show`, not just for the plaintext file. Given what this app handles (file paths, recording titles, transcript-adjacent text), the default has to be private, matching `os.Logger`, with `.public` opted into explicitly.

Two smaller things worth carrying over: `freopen` creates the log with the default umask (world-readable `0644` on a shared Mac — `0600` would be better for a file that may contain transcript text), and "truncate at 5 MB" discards the oldest evidence at exactly the moment a long session goes wrong; a two-file rotation keeps the interesting part.

Worth noting the ground has shifted since July: `DiagnosticReporter` (Help → "Save Diagnostic Report…") now packages recent OSLog entries, system info and crash reports into a zip, which covers a good part of the original motivation. A future version of this PR should be scoped against what that already gives you.

Happy to reopen if you want to take it further — no objection to the goal, only to landing it in this state.

PR #63, opened 2026-07-05
feat: color transcript speaker labels distinctly when multiple speakers detected

yotamkorah-island:
## Summary
- When diarization detects more than one speaker in a recording, each speaker's label ("Speaker A", "Speaker B", ...) is now rendered in a distinct, stable color instead of the single default tint — in both the post-recording transcript (`SegmentRow`) and the live transcript pane (`TranscriptLineView`).
- Color is derived directly from the raw `SPEAKER_NN` diarizer ID via a small fixed palette (`Mila/Transcription/SpeakerColor.swift`), so the same speaker keeps the same color across renders. Single-speaker recordings are unaffected (still use `Color.accentColor`, matching the prior `.tint` look).

## Test plan
- [x] Added `MilaTests/SpeakerColorTests.swift` covering stability, distinctness, palette wraparound, and non-standard speaker IDs.
- [x] `make test` (full Xcode) caught two real bugs during local verification, both fixed here: a `foregroundStyle` ternary mixing `Color` and `.tint` that failed to build, and an infinite loop in the wraparound test. A full local test run was interrupted before completion after those fixes landed — worth a CI run / one more local `make test` before merge to confirm everything is green end-to-end.
- [ ] Manually verify in a multi-speaker recording that speaker labels show distinct colors in both the transcript and live views.

## Summary by CodeRabbit

* **New Features**
  * Transcript speaker labels now switch to consistent per-speaker coloring when multiple distinct speakers are detected (live and saved recordings).
  * Speaker color selection is deterministic from the speaker identifier.
* **Bug Fixes**
  * Speaker colors remain stable across app relaunches by avoiding non-deterministic hashing.
  * Supports non-standard speaker IDs with predictable color mapping.
* **Tests**
  * Added tests validating deterministic colors, palette wrapping, and multi-speaker detection across both transcript and live segments.

### DIFF
--- a/Mila/Transcription/SpeakerColor.swift
+++ b/Mila/Transcription/SpeakerColor.swift
@@ -1,0 +1,25 @@
+import SwiftUI
+
+/// Stable per-speaker color for the transcript UI. When diarization finds
+/// more than one speaker, each raw `SPEAKER_NN` ID gets its own color so a
+/// multi-speaker conversation is scannable at a glance instead of relying
+/// solely on the "Speaker A" / "Speaker B" text label.
+///
+/// The index is parsed directly from the raw diarizer ID (not hashed from
+/// the label) so the same speaker keeps the same color across every
+/// render of a recording, live or post-processed.
+private let speakerColorPalette: [Color] = [
+    .blue, .green, .orange, .purple, .pink, .teal, .indigo, .brown
+]
+
+extension String {
+    var speakerColor: Color {
+        let index: Int
+        if hasPrefix("SPEAKER_"), let n = Int(dropFirst("SPEAKER_".count)), n >= 0 {
+            index = n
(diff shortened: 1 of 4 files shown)


yotamkorah-island:
Addressed all three CodeRabbit nitpicks in 74b14c0:
- `speakerColor`'s fallback no longer uses `hashValue` (randomized per process) — now a deterministic unicode-scalar sum, so a non-standard speaker ID keeps its color across app relaunches.
- Extracted the duplicated `Set(...).count > 1` distinct-speaker check from both views into a shared `hasMultipleSpeakers` helper (`SpeakerColor.swift`) that short-circuits as soon as a second distinct speaker is found, instead of building a full `Set` on every live-transcript render.
- Added unit tests for the new helper.

PR #91, opened 2026-07-14
fix(diarization): stop codesign-blocked repair loop; fix(recording): don't drop stalled-clock system audio

### LINKED ISSUE #90: Diagnostic health check triggers destructive codesign-blocked repair loop; system audio dropped after sleep/wake
Root-caused from a user diagnostic bundle (v1.8.2 b39). Four related problems:

### 1. Diagnostic health check triggers a destructive, doomed repair loop
`dlopen` of torch's dylibs fails with macOS library validation (`different Team IDs`) and is classified `unknown`, which triggers `nuclearRepair` — a wipe + ~62 MB re-download + ad-hoc re-sign — which then hits the *same* refusal. Result: every launch does a destructive re-download and diarization stays dead.

**Fix:** the health-check script emits a stable `codesign_blocked` code; the self-heal loop surfaces it without repairing, and Settings shows an actionable label.

### 2. `DiarizationBootstrap.bootstrapIfNeeded` has no reentrancy guard
`MilaApp`'s launch task and the health check's `nuclearRepair` can run two overlapping installs, with `reinstall()`'s directory wipe landing mid-install (observed: stage pinned at `downloadingTorch(0.5)` for ~57 min).

**Fix:** concurrent callers coalesce onto one in-flight task, and `reinstall()` waits for it before wiping.

### 3. System audio dropped after a display sleep/wake
When the mic clock stalls across a sleep/wake, the 30s system jitter buffer overflows and DROPS the oldest audio — 2,883 consecutive drops (~58s of the other side of a meeting) in the diagnostic.

**Fix:** overflow is flushed to the WAV and live feed (system-only, full scale, same as the stop-time tail flush) instead of discarded; the log line is throttled.

### 4. Noisy per-render log
`LiveAIRecordingView`'s per-render log was 79% of the 100k-line diagnostic.

**Fix:** demoted to `.debug`.

moshe-z:
Root-caused from a user diagnostic bundle (v1.8.2 b39).

## What's fixed

1. **Health check no longer triggers a doomed destructive repair loop.** `dlopen` of torch's dylibs failed macOS library validation (`different Team IDs`) and was classified `unknown`, which triggered `nuclearRepair` (wipe + ~62 MB re-download + ad-hoc re-sign) — hitting the same refusal every launch. The health-check script now emits a stable `codesign_blocked` code; the self-heal loop surfaces it without repairing, and Settings shows an actionable label.

2. **`DiarizationBootstrap.bootstrapIfNeeded` reentrancy guard.** `MilaApp`'s launch task and the health check's `nuclearRepair` could run two overlapping installs, with `reinstall()`'s wipe landing mid-install (observed: stage pinned at `downloadingTorch(0.5)` for ~57 min). Concurrent callers now coalesce onto one in-flight task, and `reinstall()` waits for it before wiping.

3. **System audio no longer dropped after sleep/wake.** When the mic clock stalled across a display sleep/wake, the 30s system jitter buffer overflowed and DROPPED the oldest audio (2,883 consecutive drops, ~58s of the other side of a meeting, in the diagnostic). Overflow is now flushed to the WAV and live feed (system-only, full scale, same as the stop-time tail flush) instead of discarded; the log line is throttled.

4. **Quieter diagnostics.** `LiveAIRecordingView`'s per-render log (79% of the 100k-line diagnostic) demoted to `.debug`.

## Test plan

- `MilaTests/DiarizationSettingsTests.swift` extended to cover the `codesign_blocked` classification and status labelling.
- `make test`

Closes #90

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented meeting recordings from losing app audio when the microphone clock stalls.
  * Improved speaker detection recovery when macOS security restrictions block required components.
  * Added a targeted Settings message directing affected users to reinstall Mila.
  * Reduced excessive diagnostic logging in reports.

* **Documentation**
  * Updated the unreleased changelog with recent recording, speaker detection, summarization, and timeout improvements.


urisland (approved):
Reviewed in depth — this is a solid set of fixes. Verified against `main`:

- **`RecordingSession.consumeSystem` overflow flush** — correctly mirrors `flushPendingSystemTail` (system-only, full scale, `onLive` slice + `await write`). The buffer is trimmed *before* the `await`, and `overflow` is an independent copy, so the new suspension point introduces no data race; the `.systemAudio` branch already awaited inside `consumeSystem`, so the concurrency model is unchanged. Temporal order holds: the oldest overflow prefix is flushed now, the ≤30s remainder stays buffered and is mixed by `mixWithBufferedSystem` when the mic clock resumes. Real data-loss fix.
- **`DiarizationBootstrap` reentrancy guard** — `@MainActor` makes the check-then-set atomic; clearing `inFlightBootstrap` as the task's last statement correctly lets a caller queued behind it (`reinstall`) start fresh rather than short-circuit on a completed task.
- **`codesign_blocked` classification** (Python health check → Settings self-heal) — matches the macOS library-validation error strings; surfacing without `nuclearRepair` is the right call (a reinstall reproduces the same ad-hoc signatures). Good actionable Settings label.
- **Log demotion + throttle** — correct, and the regression test (`healthCheckRunner` seam) pins the no-nuclear-repair behavior well.

Approving; will merge once CI is green.

Issue #29, opened 2026-06-24
Live AI: cross-recording summary/action-item bleed (session not guaranteed fresh per recording)


urisland:
## Summary

A new recording's **Live AI summary / action items can contain content from a previous recording** ("yesterday's summary"). Reported from a real session on v1.8.13; the Live AI pipeline does not currently *guarantee* a fresh, isolated Claude session per recording, and a secondary reset path silently fails to clear state.

## Root cause(s)

**1. Per-recording session reset is timing-dependent (primary).**
`LiveAISession.start()` — which mints a fresh `--session-id` UUID, resets `sessionEstablished = false`, and clears `summary`/`actionItems` — is invoked **only** from the async `.recording` handler in `MilaApp.wireLiveAIPipeline` (`aiSession.start()`). That handler consumes `RecordingSession.$state.values` (a Combine `AsyncPublisher`), which can **coalesce/drop** a transition when the consumer is mid-`await` (the `.idle` branch awaits `transcribeNow()` / `awaitPending()`). If a `.recording` transition is missed, the new recording keeps the **previous** recording's `sessionID` + `sessionEstablished == true`, so its first tick takes the `.resume(<previous uuid>)` branch in `kick()` → `claude --resume` reopens the prior meeting's conversation. The model, asked to re-emit all items, returns the prior meeting's items (now timestamp-less → `[0s]`) plus the new ones, and a blended summary.

**2. `LiveAISession.cancel()` does not clear `summary`/`actionItems` (confirmed).**
The gated-hardware branch in `wireLiveAIPipeline` (`guard aiSettings.isLiveAIAvailable else { … aiSession.cancel() … }`) has a comment claiming it "clears stale `summary` / `actionItems` … so a previous Live AI session's output [doesn't] attach to a recording that never ran the LLM loop." But `cancel()` only clears `latestTranscript`/`coalesced`/session id — it leaves `summary` and `actionItems` intact. On hardware below the Live AI bar, the previous recording's summary/items therefore leak directly into the next capture (which `stopRecording` snapshots unconditionally).

## Evidence / caveat

[...]

PR #111, opened 2026-07-26
Live AI: per-meeting context box (fixes blended "previous meeting + this meeting" summaries)

### LINKED ISSUE #110: Live AI: blended "previous meeting + this meeting" summary — per-meeting agenda pasted into the permanent system prompt
## Summary

A recording's Live AI output can cover **the previous meeting plus the current one** — a blended summary, and action items that were never said out loud (all stamped `timestamp_seconds: 0`).

This looks exactly like #29 (cross-recording Claude session bleed) but is a different cause: #29's per-recording session isolation is working. The content comes in through the **prompt**, not through a resumed session.

## Root cause

There is nowhere to put context about *one* meeting. The only editable text is Settings → Live AI → **System prompt**, which is a permanent template persisted to `UserDefaults` (`liveAI.prompt`) and sent on every tick of every recording. So a user preparing for a meeting pastes their agenda at the bottom of it — reasonably — and every *later* recording is then told the previous meeting's agenda is its context.

The model does the obvious thing with that: it summarises the agenda alongside the transcript, and emits an action item per agenda bullet. Because those bullets were never spoken, they come back with `timestamp_seconds: 0` — which is the same fingerprint #29 produced, hence the confusion.

## Evidence (2026-07-26, v1.9.x)

* `defaults read io.island.whisper.IslandWhisper liveAI.prompt` — the default template followed by a hand-pasted `Context for this meeting:` block holding the *previous* meeting's agenda. Nothing in Mila emits that string; it was typed by the user.
* The spawned CLI's own transcript, `~/.claude/projects/-private-var-folders-…-T-island-mila-llm-session-<UUID>/<UUID>.jsonl`, row 2 = the exact first prompt sent: template + agenda + one line of real transcript. The first assistant reply summarises the *agenda*.
* The affected recording has 18 action items: 9 with real timestamps that match its transcript, 9 at `[0s]` that map 1:1 onto the stale agenda bullets and share almost no vocabulary with the transcript.
* `log show --predicate 'category == "LiveAISession"' --info --debug` shows `start: fresh session=<uuid>` and a first tick of `session=new(<uuid>) established=false` for every recording — i.e. no session bleed. #29's instrumentation did its job and ruled itself out.

## Severity

Same confidentiality shape as #29 — one meeting's content surfacing in another's summary — but self-inflicted via a UI that offers no better option. Also wastes tokens on every tick.

## Fix

1. `LiveAISettings.meetingContext` — per-meeting background notes, **in-memory only** (never written to `UserDefaults`, so a relaunch can't resurrect them) and cleared by `stopRecording` on both the normal and failed-stop paths.
2. `LiveAISession.promptWithContext` — appends the notes as an explicitly fenced `BACKGROUND NOTES` block that tells the model the notes are not transcript, must not be summarised, and must not become action items. Read per tick (a meeting-detector auto-start means notes usually arrive after the recording began).
3. UI: a **Context** toggle + editor in the Live AI recording pane (filled icon when notes are set, Clear button), plus a note under the Settings prompt editor pointing per-meeting content at that box.

## Not in this PR

`stopRecording` never calls `liveAISession?.cancel()`, so a throttled tick can fire after the recording ends (observed: recording stopped 13:33:58, a full `claude --resume` tick ran at 13:34:49) and the session's stable sandbox lingers until the next `start()`.

urisland:
Closes #110

## The bug, and what it wasn't

A recording's Live AI output covered **the previous meeting plus the current one**: a blended summary, plus 9 action items that were never spoken (all `timestamp_seconds: 0`) sitting alongside 9 real ones.

That `[0s]` cluster is the exact fingerprint of #29 (cross-recording Claude session bleed), so that's where I started — and #29's instrumentation ruled itself out: every recording logs `start: fresh session=<uuid>` followed by a first tick of `session=new(<uuid>) established=false`. No `--resume` of a prior meeting.

The stale content came in through the **prompt**. Settings → Live AI → System prompt is a permanent template (`liveAI.prompt` in `UserDefaults`, sent on every tick of every recording), and it's the only editable text Mila offers — so preparing for a meeting means pasting the agenda at the bottom of it, and every *later* recording is then told that agenda is its context. The model summarises the agenda and emits an item per bullet; because nothing was spoken, the items come back at `[0s]`.

Confirmed against the spawned CLI's own transcript (`~/.claude/projects/…-island-mila-llm-session-<UUID>/<UUID>.jsonl`, row 2 = the literal first prompt): default template + a hand-written `Context for this meeting:` block + one line of real transcript, and the first assistant reply summarising the agenda. `grep` confirms no Mila code emits that string.

## The fix

* **`LiveAISettings.meetingContext`** — per-meeting background notes, **in-memory only**. Deliberately not persisted: that persistence is precisely what leaked, so a relaunch can't resurrect a meeting's notes. Cleared by `stopRecording` on both the normal path and the failed-stop early return.
* **`LiveAISession.promptWithContext`** — appends the notes as a fenced `BACKGROUND NOTES` block that tells the model they are not transcript, must not be summarised, and must not become action items (an agenda reads exactly like a record of decisions already taken, which is why the guard rails are the load-bearing part). Read **per tick**, not snapshotted at `start()` — a meeting-detector auto-start beats the user to the pane, so notes usually arrive a minute in.
* **Live AI pane** — a `Context` toggle in the "Summary & action items" header (filled icon when notes are set), a collapsible editor, and Clear.
* **Settings prompt editor** — one line pointing per-meeting content at that box, so the permanent template stops being the obvious place for it.

## Verification

* `make build` and `xcodebuild build-for-testing` both succeed.
* New `MilaTests/LiveAIMeetingContextTests.swift`: prompt wire format (absent when blank, fenced + guard-railed when set), the notes reaching a real `feed → scheduleKick → kick` tick through a stubbed `performCall`, per-tick re-read (added mid-recording, then cleared, both take effect on the next tick), and the not-persisted invariant (fresh settings object over the same suite comes back empty, and no key in the suite contains the text).
* Test suite not run locally (it would take over the mic and steal focus mid-workday) — left to CI. The new pane UI has not been visually verified for the same reason: the Context box only renders while a recording is live.

## Follow-up, not in here

`stopRecording` never calls `liveAISession?.cancel()`, so a throttled tick can fire after the recording ends — observed today: stop at 13:33:58, a full `claude --resume` tick at 13:34:49 — and the session's stable sandbox lingers until the next `start()`.

## Summary by CodeRabbit

* **New Features**
  * Added a per-meeting “Context” editor to Live AI recording (expand/collapse, edit, and clear).
  * Live AI now injects the current meeting context into each live update and adds guard-rails to keep responses transcript-based.
  * Clarified in Settings that the system prompt is reused for every recording and meeting-specific details belong in Context.
* **Bug Fixes**
  * Meeting context is now cleared reliably when recording stops (including edge-case teardown paths).
  * If meeting notes are changed mid-recording, Live AI restarts so removed notes no longer affect subsequent outputs.

### DIFF
--- a/Mila/Actions/LiveAISession.swift
+++ b/Mila/Actions/LiveAISession.swift
@@ -81,6 +81,36 @@ final class LiveAISession: ObservableObject {
         )
     }
 
+    /// Append the user's per-meeting background notes to `prompt` as an
+    /// explicitly-fenced non-transcript block. Returns `prompt` unchanged
+    /// when there are no notes.
+    ///
+    /// The guard rails in the block text are the whole point. Notes are
+    /// almost always an agenda ("Wojtek's team — recruitment going better;
+    /// flying in beginning of August"), which reads exactly like a
+    /// transcript of decisions already taken. Without being told
+    /// otherwise the model summarises the AGENDA and emits an action item
+    /// per bullet — the user then sees a summary of a meeting that hasn't
+    /// happened yet mixed into the one that has. So we say three things:
+    /// this is not transcript, don't summarise it, don't mine it for
+    /// items.
+    ///
+    /// `static` + pure so the wire format is unit-testable without a
+    /// subprocess.
+    static func promptWithContext(_ prompt: String, context: String) -> String {
+        let notes = context.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !notes.isEmpty else { return prompt }
+        return """
+        \(prompt)
+
+        ---
+        BACKGROUND NOTES (supplied by the user before this meeting — these are NOT part of the transcript and describe what the user PLANNED to discuss):
+        \(notes)
+
+        Use the notes only to interpret what you actually hear: names, spellings, acronyms, and which topic is being discussed. Do NOT summarise the notes, and do NOT turn them into action items. Every action item and every sentence of the summary must come from the transcript itself — if something in the notes never comes up in the conversation, it does not appear in your output at all.
+        """
+    }
+
     /// Pure throttle core: how long to wait before the next tick may
     /// *start*, given the previous tick's start time and the configured
     /// minimum spacing. `0` means "start now". Exposed `static` so it can
@@ -304,8 +334,13 @@ final class LiveAISession: ObservableObject {
                 return "Hebrew"
             }
         }()
-        let prompt = liveAISettings.prompt
-            .replacingOccurrences(of: "{{LANGUAGE}}", with: promptLanguageName)
+        // Read the per-meeting notes at TICK time, not at `start()` — a
+        // meeting-detector auto-start beats the user to the pane, so the
+        // agenda usually gets pasted a minute into the recording.
+        let prompt = Self.promptWithContext(
+            liveAISettings.prompt
+                .replacingOccurrences(of: "{{LANGUAGE}}", with: promptLanguageName),
+            context: liveAISettings.meetingContext)
         let model = liveAISettings.model
         let useSession = (sessionID != nil)
         // Cold-start the first tick gets a longer timeout (see
--- a/Mila/Actions/LiveAISession.swift
+++ b/Mila/Actions/LiveAISession.swift
@@ -367,8 +367,14 @@ final class LiveAISession: ObservableObject {
             // The abandoned session's stable sandbox is nobody's job
             // otherwise: `cancel()` only cleans up the CURRENT sessionID.
             // Safe to remove now — `kick()` only runs with no call in
-            // flight, so no live child process is chdir'd into it.
-            LLMRunner.cleanupStableSandbox(key: stale.uuidString)
+            // flight, so no live child process is chdir'd into it. Off the
+            // main actor for the same reason `cancel()` defers it: a
+            // recursive directory removal is a blocking syscall and this is
+            // the tick path.
+            let staleKey = stale.uuidString
+            Task.detached(priority: .utility) {
+                LLMRunner.cleanupStableSandbox(key: staleKey)
+            }
         }
         let prompt = Self.promptWithContext(
             liveAISettings.prompt
--- a/Mila/Actions/LiveAISession.swift
+++ b/Mila/Actions/LiveAISession.swift
@@ -137,6 +137,13 @@ final class LiveAISession: ObservableObject {
     /// tick retries the create.
     private var sessionEstablished: Bool = false
 
+    /// The per-meeting notes as they were on the last SUCCESSFUL tick.
+    /// Compared against the current notes on every tick to detect a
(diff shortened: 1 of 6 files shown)


Mila/Actions/LiveAISession.swift
```diff
@@ -304,8 +343,37 @@ final class LiveAISession: ObservableObject {
...
+        // `--resume` tick carries every earlier turn, so notes already sent
+        // stay in the model's conversation history no matter what the next
+        // prompt says. Dropping the block would leave the model still
+        // reading from the version the user just deleted. So a change in
+        // the notes restarts the Claude session: fresh UUID, and
+        // `lastTranscriptSent` reset so the first tick of the new session
+        // re-ships the whole transcript alongside the current notes.
+        // Costs one full-transcript call, only when the user edits notes
+        // mid-meeting. (CodeRabbit on #111.)
+        if sessionEstablished, context != lastContextSent, let stale = sessionID {
```
urisland, line 362: Both findings fixed in 54aa0d1.

1. **Session restart on a context change** — you're right that clearing the field only changed the next prompt; a `--resume` tick keeps every earlier turn, so the old notes stayed in conversation history. `kick()` now compares the current notes against the last *successfully delivered* ones, and on a difference mints a fresh session id, resets `lastTranscriptSent` so the new session receives the whole transcript, and removes the abandoned session's stable sandbox (`cancel()` only ever knew the current id, so that dir would have leaked). Restart cost is one full-transcript call, paid only when the user actually edits notes mid-meeting. Covered by `test_editingContextMidRecording_restartsSession` (new id + `.new` + full transcript, and an unchanged context still resuming) and `test_clearingContextMidRecording_restartsSession`.

2. **Store-removal early return** — added the clear there too. All three exits from a finished recording now clear the notes.

Mila/Actions/LiveAISession.swift
```diff
@@ -304,8 +343,43 @@ final class LiveAISession: ObservableObject {
...
+            liveAILog.log("context changed — restarting session (was \(stale.uuidString.prefix(8), privacy: .public))")
+            sessionID = UUID()
+            sessionEstablished = false
+            lastTranscriptSent = ""
+            // The abandoned session's stable sandbox is nobody's job
+            // otherwise: `cancel()` only cleans up the CURRENT sessionID.
+            // Safe to remove now — `kick()` only runs with no call in
+            // flight, so no live child process is chdir'd into it. Off the
+            // main actor for the same reason `cancel()` defers it: a
```
urisland, line 371: Applied in cd321a8 — deferred via `Task.detached(priority: .utility)`, matching `cancel()`. You're right that it's the same blocking-syscall-on-the-tick-path concern.

urisland:
@coderabbitai review

PR #113, opened 2026-07-26
Cancel the Live AI session when a recording stops

### LINKED ISSUE #112: Live AI session outlives the recording: a throttled tick fires after Stop, and the session sandbox leaks
## Summary

`QuickActionsController.stopRecording` tears down `liveTranscriber` and `liveDiarizer` (the "END OF THE LIVE-PIPELINE-OWNING PHASE" block) but never calls `liveAISession?.cancel()`. The Live AI session therefore outlives the recording, with two consequences.

**1. A tick fires after the recording is over.**
A `pendingKickTask` sleeping out the min-interval floor when the user hits Stop wakes up afterwards and spawns a full `claude --resume` subprocess. Its result lands on a `Recording` that was already assembled and written, and `finalizeTail` regenerates the summary from the complete transcript regardless — so the call is pure waste (tokens, CPU, and a ~1 GB-ish CLI process spun up after the meeting).

Observed 2026-07-26 on 1.9.x:

```
13:33:58  RecordingSession  stop: source=meeting micFrames=14615051 writes=9136
13:34:00  RecordingSummarizer  started 02831BA1 transcript=8984c force=true
13:34:49  LiveAISession  tick tick-89094: session=resume(EFB5F132-…) established=true
```

(`log show --predicate 'subsystem == "io.island.whisper.IslandWhisper" AND category == "LiveAISession"' --info --debug`)

**2. The session sandbox leaks.**
`LLMRunner.stableSandboxDirectory` creates `/var/folders/…/T/island-mila-llm-session-<uuid>` per Live AI session, and only `LiveAISession.cancel()` removes it. So it survives until the NEXT recording calls `start()` — and forever if the app quits first. `~/.claude/projects/` likewise accumulates one project directory per recording.

## Fix

Call `liveAISession?.cancel()` in `stopRecording`, **after** the final summary/action-item snapshot has been written to the store — `cancel()` clears `summary`/`actionItems`, so calling it earlier would wipe the data being saved. All three exits from a finished recording get it: the failed-`session.stop()` early return, the store-removal early return, and the normal path.

Deliberately NOT done in `wireLiveAIPipeline`'s `.idle` handler: that fires *during* `session.stop()`, i.e. before the snapshot, which is exactly why it currently leaves the session alone. `stopRecording` is the only path that knows when the snapshot is safe. Paths that never reach `stopRecording` (cancelAll / app quit) keep today's behaviour — the next recording's `start()` resets the session.

## Follow-up from

#110 / #111 — the "Follow-up, not in here" section of that PR describes this exact bug.

urisland:
Closes #112 — the follow-up flagged in #111.

## The bug

`stopRecording` tears down `liveTranscriber` and `liveDiarizer` but never cancelled `liveAISession`, so the session outlived the recording in two ways:

1. **A tick fires after the meeting is over.** A `pendingKickTask` sleeping out the min-interval floor at Stop wakes up afterwards and spawns a full `claude --resume` subprocess, whose result lands on a `Recording` that was already assembled and written. `finalizeTail` regenerates the summary from the complete transcript regardless, so the call is pure waste. Observed 2026-07-26:

   ```
   13:33:58  RecordingSession  stop: source=meeting micFrames=14615051
   13:34:00  RecordingSummarizer  started 02831BA1 transcript=8984c force=true
   13:34:49  LiveAISession  tick tick-89094: session=resume(EFB5F132-…) established=true
   ```

2. **The session sandbox leaks.** `/var/folders/…/T/island-mila-llm-session-<uuid>` is only removed by `cancel()`, so it survived until the next recording's `start()` — and forever if the app quit first. `~/.claude/projects/` accumulates a directory per recording too.

## The fix

`liveAISession?.cancel()` on all three exits from a finished recording (failed `session.stop()`, store-removal early return, normal path) — in each case **after** the summary/action-item snapshot is in the store, because `cancel()` clears `summary`/`actionItems` and that's the data being saved.

Deliberately not moved into `wireLiveAIPipeline`'s `.idle` handler: that fires *during* `session.stop()`, i.e. before the snapshot, which is exactly why it leaves the session alone today. `stopRecording` is the only path that knows when the snapshot is safe. Its stale comment ("QuickActionsController still needs to read .summary…") now explains the real ordering constraint. Paths that never reach `stopRecording` (cancelAll / app quit) keep today's behaviour — the next recording's `start()` resets the session.

Checked for consumers that read the session after that point: the only ones are the two snapshots in `stopRecording` itself (both earlier) and `LiveAIRecordingView`, which is off screen by then (`activeJob = .none` is set long before). `finalizeTail` is documented and confirmed not to touch the live singletons — it regenerates the summary from the saved transcript.

## Verification

* `make build` and `xcodebuild build-for-testing` both succeed.
* New `test_cancel_dropsDeferredTick` in `LiveAIThrottleTests` pins the mechanism this fix depends on and that nothing covered before: a tick deferred by the throttle must not fire after `cancel()`. It uses a deliberately short floor (0.5s) so an uncancelled task would wake inside the test window and be caught.
* The call-site wiring itself has no unit test — driving `stopRecording` needs live capture — so that part rests on the build plus the consumer audit above. Suite not run locally (mic + focus); left to CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recording shutdown to reliably tear down live AI sessions across successful, failed, and cancelled recording flows.
  * Ensured recording summary and action items are preserved by cleaning up the live AI session at the correct time.
  * Fixed an issue where deferred AI processing could still run after the session was cancelled.
* **Tests**
  * Added a regression test covering cancellation during deferred AI processing.

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -1304,11 +1304,15 @@ struct MilaApp: App {
                     await diarizer.awaitPending()
                     diarizer.stop()
                 }
-                // Note: don't cancel aiSession here — QuickActionsController
-                // still needs to read .summary and .actionItems out of
-                // it when assembling the saved Recording. The NEXT
-                // recording clears these via `liveAISession?.start()` at
-                // record-start (in QuickActionsController).
+                // Note: don't cancel aiSession here — this handler fires
+                // during `session.stop()`, BEFORE QuickActionsController
+                // has read .summary / .actionItems for the saved
+                // Recording, and `cancel()` clears both. `stopRecording`
+                // cancels it itself once that snapshot is safely in the
+                // store (it's the only path that knows when that is); on
+                // the paths that don't go through `stopRecording`
+                // (cancelAll / app quit), the next recording's
+                // `liveAISession?.start()` resets it.
                 sessionRef.onLiveSamples = nil
             }
         }
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -1303,16 +1303,29 @@ struct MilaApp: App {
                     // utterance would lose its speaker label.
                     await diarizer.awaitPending()
                     diarizer.stop()
+                    // This branch owns the whole teardown, so kill the Live
+                    // AI session with the rest of it: otherwise a tick
+                    // deferred by the min-interval floor can still wake and
+                    // spawn a `claude` child while the app is tearing down
+                    // AVAudioEngine (CodeRabbit on #113). Safe to clear
+                    // `summary` / `actionItems` here — the only path that
+                    // reaches this branch today is `cancelAll()` at app
+                    // termination, which by design does NOT flush a WAV or
+                    // assemble a Recording (the orphan WAV is recovered on
+                    // next launch instead), so there is no snapshot to
+                    // protect. Sleep and screen-lock both route through
+                    // `stopRecording`, which owns finalize and cancels the
+                    // session itself once its snapshot is stored.
+                    aiSession.cancel()
                 }
-                // Note: don't cancel aiSession here — this handler fires
-                // during `session.stop()`, BEFORE QuickActionsController
-                // has read .summary / .actionItems for the saved
-                // Recording, and `cancel()` clears both. `stopRecording`
-                // cancels it itself once that snapshot is safely in the
-                // store (it's the only path that knows when that is); on
-                // the paths that don't go through `stopRecording`
-                // (cancelAll / app quit), the next recording's
-                // `liveAISession?.start()` resets it.
+                // Note: no aiSession.cancel() out here, outside the branch
+                // above — this handler fires during `session.stop()`,
+                // BEFORE QuickActionsController has read .summary /
+                // .actionItems for the saved Recording, and `cancel()`
+                // clears both. When `stopRecording` owns finalize it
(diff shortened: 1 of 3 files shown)


Mila/App/MilaApp.swift
```diff
@@ -1303,12 +1303,29 @@ struct MilaApp: App {
                     // utterance would lose its speaker label.
                     await diarizer.awaitPending()
                     diarizer.stop()
```
urisland, line 1305: Applied in ce99324, with one adjustment to the placement.

Put at the **end** of the branch's drain rather than between `diarizer.stop()` and the lines above it — your diff position sits just after that branch's own final `aiSession.feed(transcript:, immediate: true)`, and `cancel()` kills the in-flight tick, so cancelling there would have torn down the flush two lines above it.

Also narrowed the reasoning in the comment: that branch is reachable only by `cancelAll()` at app termination today. Sleep and screen-lock both route through `stopRecording` (`pauseForScreenLock` → `stopRecording`; `willSleep` → `pendingSleepStopReason` → `stopRecording`), so they own finalize and get the cancel from the other side. That's what makes clearing `summary`/`actionItems` safe here: `cancelAll()` by design flushes no WAV and assembles no Recording (the orphan WAV is recovered on next launch), so there is no snapshot to protect. The stale comment below the branch claimed those paths relied on the next recording's `start()` — corrected.

PR #99, opened 2026-07-21
Obsidian support and live-recording UX improvements

ValeroK:
## Summary

Contributes the fork's accumulated work back to upstream (35 commits). Major themes:

- **Obsidian export**: export completed recordings to an Obsidian vault; mila-anywhere PRD + destination plan.
- **Gemini CLI**: added Gemini CLI support and hardened LLM executable lookup.
- **MCP server (`mila-mcp`)**: MCP stdio server over Mila's transcriptions, `MilaKit` package extraction, live-transcript sidecar during recording, and a store-location pointer for read-only store access.
- **Speaker renaming**: persistent, searchable speaker name directory threaded into live copy-transcript text.
- **Live-recording UX**: recording-folder + meeting-name controls, always-scroll transcript, pause/resume, per-line deletion, and a copy-transcript button.
- **Reliability fixes**:
  - Fix UI freeze when opening folder menus during recording/transcription.
  - Repair unfinalized WAV headers on crash recovery so force-quit recordings stay transcribable.
  - Fix recording folder lost when renaming after stop.
  - AGC / VAD noise-floor fixes for quiet built-in mics.
- **Packaging**: DMG entitlements + "Mila Local Dev" cert so TCC grants survive installs; wider Settings window.

## Test plan

- [ ] `make project && make build`
- [ ] `make test` and `make package-test` (TranscriptionCore + MilaKit) pass
- [ ] Manual: record, force-quit, relaunch -> recording recovers and transcribes (WAV header repair)
- [ ] Manual: open folder menu during recording/transcription -> no UI freeze
- [ ] Manual: Obsidian export writes to the configured vault

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added pause/resume for recordings (with accurate paused elapsed-time) plus live transcript controls: speaker renaming, line deletion, clipboard copy, and speaker-aware exports.
  * Added optional Obsidian Markdown export with organized files, backfill, and optional Git sync.
  * Added Gemini CLI support and a bundled MCP server for transcript search and live meeting polling.
  * Added live transcript crash recovery and automatic repair for incomplete WAV headers.
* **Bug Fixes**
  * Reduced UI freezes by throttling transcription updates and isolating high-frequency views.
  * Improved adaptive gain for quiet speech/background noise.
* **Documentation**
  * Expanded local build/signing guidance and MCP/Obsidian destination setup docs.

PR #116, opened 2026-07-27
fix(transcription): coalesce whisper progress callbacks and keep them monotonic

ValeroK:
## Summary

whisper.cpp fires its progress callback many times a second from its background compute thread, and `TranscriptionService` handed every single tick its own `Task { @MainActor }`. That has two consequences:

1. **Unbounded main-actor flood.** Each tick mutates `@Published progress`, which re-renders every observing view, for the entire duration of a run.
2. **Out-of-order application.** Independent `Task`s have no ordering guarantee, so a late-running earlier value can overwrite a later one — the visible symptom is a progress bar that sticks below a percentage it had already reported.

The callback now goes through a small lock-guarded `ProgressCoalescer`. The compute thread schedules a main-actor flush only when one isn't already pending, and the flush publishes the max of everything offered since the last flush. Outstanding work is bounded to a single hop, and reported progress can only move forward.

The coalescer is created per run (local to `transcribe`), so there's no cross-recording state to reset.

## Notes for reviewers

- Split out of #99 per review feedback; independent of every other slice, no behavior change beyond the progress reporting itself.
- This is the part of that investigation that stands on its own. The related view-isolation refactor is a separate concern and will come with the PR that actually introduces the menus it protects, rather than being smuggled in here.

## Test plan

- [x] New `MilaTests/ProgressCoalescerTests.swift` — 4 tests: only the first offer schedules until flushed, flush returns the max offered in the window, a stale lower value can't regress progress across flushes, and a 101-offer burst collapses to exactly one scheduled flush reporting the final value.
- [x] `make build` clean.
- [x] Full `MilaTests` suite: 570 tests, 0 failures (566 before this PR).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcription progress updates to remain smooth, ordered, and steadily increasing.
  * Prevented outdated progress updates from overwriting newer values during rapid transcription activity.
  * Reduced progress-update bursts so transcription status remains responsive and accurate.

* **Tests**
  * Added coverage for progress coalescing, burst updates, maximum-value handling, and monotonic progress advancement.


urisland:
@coderabbitai Thanks — I went through both outside-diff findings against the code at `93b9ca1`. Neither is actionable; details below.

---

**1. "Enforce monotonic progress across flushes" — not valid, the premise is incorrect.**

> The coalescer keeps the maximum only for the current pending batch. A lower callback received after `flush()` can be returned by the next flush.

`latest` is never reset. `flush()` clears only the `scheduled` flag and returns the stored value:

```swift
private var latest: Double = 0
private var scheduled = false

func offer(_ value: Double) -> Bool {
    lock.lock(); defer { lock.unlock() }
    latest = max(latest, value)   // monotonic, and persists across flushes
    if scheduled { return false }
    scheduled = true
    return true
}

func flush() -> Double {
    lock.lock(); defer { lock.unlock() }
    scheduled = false             // note: `latest` is NOT cleared here
    return latest
}
```

Because `latest` survives the flush and is only ever written through `max`, a lower value arriving after a completed flush cannot lower it — the next flush still returns the running maximum. Monotonicity is a property of the accumulator itself, so there is nothing to enforce at the assignment site.

The requested coverage ("a test that sends a lower value after a completed flush") is already in the PR — `ProgressCoalescerTests.test_progress_is_monotonic_across_flushes`:

```swift
_ = c.offer(0.2)
XCTAssertEqual(c.flush(), 0.2, accuracy: 1e-9)

// A stale, lower value arriving later must not lower the reported value.
_ = c.offer(0.15)
XCTAssertEqual(c.flush(), 0.2, accuracy: 1e-9,
               "a lower late value must not regress progress")
```

That is exactly the scenario described, and it passes.

---

**2. "Bind each flush to the transcription run, not only the recording" — theoretical, pre-existing, and narrowed rather than introduced by this PR.**

The concern is a stale flush from run N landing after run N+1 has started on the same `recordingID`. Three reasons I'm not adding a run token here:

*It is not introduced by this PR.* The previous code spawned a `Task { @MainActor }` per tick guarded by the identical `self.activeRecordingID == recordingID` check. The cross-run exposure is character-for-character the same; this PR changes only how many such tasks are outstanding.

*This PR strictly reduces the window.* Before, an unbounded number of main-actor hops could be in flight at teardown. Now at most one can be. If the old exposure was acceptable, a strictly smaller one is too.

*Reaching the bad interleaving requires whisper to fire its progress callback after `transcribe()` has already returned.* The run body is `@MainActor`, and the teardown/restart path is:

```swift
activeRecordingID = recordingID
progress = 0

defer {
    if activeRecordingID == recordingID {
        activeRecordingID = nil
        progress = 0
    }
}
```

Any flush scheduled while the engine is running is enqueued on the main actor and drains at the next suspension point — long before the `defer` runs, let alone before a subsequent run sets `activeRecordingID` again. For a stale flush to survive into run N+1 it would have to be enqueued *after* the engine returned, i.e. a progress callback fired post-return. That is outside the callback contract, and if it ever happened the pre-PR code would have misbehaved identically.

Adding a run token would mean threading new lifecycle state through `TranscriptionService` to close a window that this PR already shrinks, in a change deliberately scoped to the callback path. If the post-return callback case is ever actually observed, the right fix is at the engine boundary (stop invoking the callback once `transcribe` returns), not a second guard at the consumer.

Happy to revisit if you can point at an ordering that does not depend on a post-return callback.

PR #117, opened 2026-07-27
fix(settings): widen the Settings window so the last tab stays clickable

ValeroK:
## Summary

The Settings scene is up to ten tabs (Hotkeys, Audio, Models, LLM, Speakers, Meetings, Live AI, Voice Memos, Storage, Updates), and they no longer fit in a 560pt window — the tab bar clips at the right edge and "Updates" ends up hard or impossible to click. Widen the window to 640pt.

One-line change plus the comment explaining the constraint, so the next tab addition has a hint that the width is load-bearing.

## Notes for reviewers

- The 640pt figure comes from what the ten-tab bar needs in practice on the fork this was developed on; worth a quick look on your setup, since tab-bar metrics shift with the system font size.

## Test plan

- [x] `make build` clean.
- [x] Full `MilaTests` suite: 566 tests, 0 failures (unchanged — this is a layout constant).
- [ ] Reviewer sanity check: open Settings, confirm the Updates tab is fully visible and clickable.

### DIFF
--- a/Mila/Views/SettingsView.swift
+++ b/Mila/Views/SettingsView.swift
@@ -49,7 +49,9 @@ struct SettingsView: View {
                 .tabItem { Label("Updates", systemImage: "arrow.triangle.2.circlepath") }
                 .tag(SettingsTab.updates)
         }
-        .frame(width: 560, height: 560)
+        // Wide enough that the full 10-tab bar fits and the last tab
+        // (Updates) stays clickable — at 560 it can clip off the right edge.
+        .frame(width: 640, height: 560)
         .padding(20)
         .onChange(of: SettingsNavigation.shared.pendingTab, initial: true) { _, newTab in
             if let tab = newTab {


urisland:
Thanks for this — closing as superseded, but your diagnosis was right and it shaped the fix that landed.

#138 (merged in 42d1314) makes the same change. The one difference is the number: this PR estimated 640 pt from the ten-tab bar as rendered on your fork, and the merged fix measured the strip from the accessibility tree instead — the ten items occupy a **599 pt** run, laid out across the *padded* 740 pt window rather than inside the 560 pt frame, with the item origins sitting 70 pt and 71 pt from the window edges. That is also what explains the original symptom precisely: at `width: 560` the window is 600 pt, so a 599 pt run had 600 pt to sit in, which is why exactly **one** tab collapsed into the `>>` overflow rather than two.

It went to 700 pt for headroom, with the measurements written into a comment so the next person adding a tab has real numbers instead of a guess. Your point that "tab-bar metrics shift with the system font size" is called out there too.

Closing this one only to avoid two PRs doing the same thing — the analysis was correct and the fix is in.

PR #118, opened 2026-07-27
test(ui): skip the launch-time whisper prewarm under UI tests

ValeroK:
## Summary

`prewarmDefaultModel()` loads the selected default whisper model at launch. Under XCUITest that's pure cost:

- No UI test depends on a warm model. The transcription E2E tests pass `--ui-test-tiny-model-path=` and load on demand, so every UI-test launch was warming a potentially multi-GB model for nothing.
- It leaves a whisper/Metal context alive that never gets torn down: XCTest ends the app with `exit()`, which bypasses `gracefulShutdown()`. That's where the `ggml-metal` teardown backtrace at process exit comes from.

Guard the prewarm behind a launch-argument check.

## Notes for reviewers

- Every UI test target already passes either `--uitests` or a `--ui-test-*` flag (`AudioLoopbackUITests`, `DetailLayoutUITests`, `MeetingStopPromptUITests`, `OrganizationUITests`, `SpeakerSelfHealUITests`, `WhatsNewPopupUITests`), so the prefix check covers all of them. A normal launch passes neither, so behavior there is unchanged.

## Test plan

- [x] `make build` clean.
- [x] Full `MilaTests` suite: 566 tests, 0 failures (unchanged — this only affects launch under UI tests).
- [ ] Worth confirming in CI, where the UI-test job can actually run: the `ggml-metal` teardown backtrace should stop appearing. I couldn't verify that locally — the UI-test runner can't get automation permission on my machine ("Timed out while enabling automation mode"), which is unrelated to this change.


urisland:
@coderabbitai review

urisland:
@coderabbitai full review

PR #120, opened 2026-07-27
feat(recordings): add an Empty Trash bulk delete action

ValeroK:
## Summary

Clearing Recently Deleted meant deleting rows one at a time through the context menu. This adds a bulk "Empty Trash" action.

`RecordingStore.emptyTrash()` applies exactly the same per-recording cleanup as `permanentlyDelete(_:)` — audio plus the `.txt` / `.summary.txt` / `.srt` sidecars, and a Voice-Memos tombstone so a deleted import doesn't come back on the next sync — but batches the store and tombstone writes so a full bin doesn't trigger one persist per row. It returns the number removed, and skips persisting entirely when the trash is already empty.

The button appears in two places, both scoped to Recently Deleted (the affordance is easy to miss in just one): the list header and the window toolbar. It's disabled while the trash is empty, and gated behind a confirmation dialog naming the count, since there's no undo — same bar as the existing per-row "Delete Permanently".

## Notes for reviewers

- The tombstone behavior is the part worth a second look: it's the same contract as the per-row delete, and there's a test pinning it, because without it a deleted Voice-Memos import silently reappears on the next sync.

## Test plan

- [x] 3 new tests in `MilaTests/RecordingStoreTests.swift`: trashed recordings and their audio files are removed while a live recording's row and file survive (including across a store relaunch); emptying an already-empty trash is a no-op returning 0 and persists nothing; and a trashed Voice-Memo import gets tombstoned, with the tombstone surviving relaunch.
- [x] `make build` clean.
- [x] Full `MilaTests` suite: 569 tests, 0 failures (566 before this PR).


urisland:
@coderabbitai review

urisland:
@coderabbitai full review

PR #122, opened 2026-07-27
fix(remote): per-language model routing; remove Auto-detect

### LINKED ISSUE #121: Remote backend sends English recordings to the Hebrew-only ivrit model
## What's wrong

With the **Remote** transcription backend pointed at an ivrit.ai server, English recordings come back with Hebrew words spliced into the text — even with the toolbar language set to 🇬🇧 English.

## Root cause

Not a bad language hint — `language=en` is sent correctly:

```
RemoteWhisperEngine transcribe: POST model=ivrit-ai/whisper-large-v3-turbo-ct2 lang=en
```

It's the model. `remote.model` is a **single fixed string** with no per-language routing, and the ivrit.ai finetune is Hebrew-only: the decoder honours `<|en|>` but the weights were never trained multilingual, so the output distribution leaks Hebrew tokens.

The local backend already routes per language — `ModelManager.model(for:)` sends `he` → ivrit, everything else → OpenAI turbo. Remote skips that entirely: both `TranscriptionService.transcribeOnceSegments` (live) and `transcribeAndSave` (batch) return before the local-model gate and call `remoteSettings.currentConfig()`, which has no language parameter. The toolbar language flag effectively does nothing on the remote backend beyond setting a form field the model can't honour.

## Fix

Add an optional second model id used for English and auto-detect recordings, mirroring `ModelManager.model(for:)` on the remote side. Empty (the default) means "use the primary model for every language", so multilingual endpoints like OpenAI's `whisper-1` and existing configurations are unaffected.

Server side, the deployment needs to actually serve a second model: zeroabstraction/cloud-helm-configurations#11358 warms `deepdml/faster-whisper-large-v3-turbo-ct2` alongside ivrit on `mila-asr`.

## Repro

1. Settings → Models → Remote, pointed at a server whose model is `ivrit-ai/whisper-large-v3-turbo-ct2`.
2. Set the toolbar language to English.
3. Record English speech.
4. Transcript contains Hebrew words.

### LINKED ISSUE #123: Remove the Auto-detect recording language
## Why

The toolbar language picker offers 🌐 Auto-detect alongside Hebrew and English. It should go: the speaker picks the language.

Auto-detect reads as a free win but isn't, because **the model is chosen from the language setting**, not the other way round:

- Local: `ModelManager.model(for:)` sends `he` → the ivrit.ai finetune, everything else → OpenAI turbo.
- Remote: the same split across two configured model ids (#121).

So "let the model detect it" actually meant "pick which specialist to ask *before* knowing what language it has to handle". On auto the local path kept whichever model was selected — typically the Hebrew finetune — which then rendered detected English as Hebrew, the exact failure auto-detect was supposed to avoid. A one-word toolbar choice produces a better transcript than any of the fallbacks.

## Scope

- Drop `RecordingLanguage.auto`. The toolbar picker is `allCases`-driven, so it disappears with the case.
- Migrate a persisted `"auto"` to Hebrew on launch and rewrite the key, so the retired value doesn't linger in defaults.
- `RecordingLanguage.fromCode("auto")` → `.hebrew`, in line with its existing unknown-code fallback.
- Keep engine-level tolerance for the literal string: recordings made under auto still carry `"auto"` on disk and can be re-transcribed. Those resolve to the multilingual model (local: `bestModel` default → turbo; remote: the English model id), which is the right engine for a detect pass.

urisland:
Closes #121, closes #123

Two commits: the remote-model fix, then the Auto-detect removal that follows from it.

## 1. English recordings went to the Hebrew-only model (#121)

On the Remote backend, English recordings came back with Hebrew words spliced in. The language hint was fine — `language=en` is sent, confirmed from a live session:

```
RemoteWhisperEngine transcribe: POST model=ivrit-ai/whisper-large-v3-turbo-ct2 lang=en
```

It's the model. `remote.model` was a single fixed string, and the ivrit.ai finetune is Hebrew-only — the decoder honours `<|en|>`, but weights that were never trained multilingual still leak Hebrew tokens. The local backend has routed per language since forever (`ModelManager.model(for:)`); remote returned before that gate and called `currentConfig()`, which had no language parameter. The toolbar 🇬🇧 flag set a form field and nothing else.

- `RemoteTranscriptionSettings.englishModel` (`remote.model.en`) + `model(for:)`, the remote mirror of `ModelManager.model(for:)`
- `currentConfig(for:)` bakes the routed id into the snapshot; both call sites that were dropping the language — live and batch — now pass it
- Settings → Models gains an **English model** field, blank by default
- `.milaconfig` carries `englishModel`, so a team config hands over both ids
- Self-hosted docs + the example `k8s/deployment.yaml` warm both models

**Empty means "use the primary for every language"** — OpenAI's `whisper-1`, any multilingual deployment, and every existing configuration behave exactly as before. Inert until someone fills the field in.

## 2. Auto-detect is gone (#123)

Auto existed to handle code-switching, but it couldn't: the model is chosen *from* the language setting, so auto meant picking which specialist to ask before knowing what language it had to handle. The local path kept whichever model was selected — usually the Hebrew finetune — which rendered detected English as Hebrew, the exact failure auto was supposed to prevent. Speakers pick the language; it's one click and it produces a better transcript.

- `RecordingLanguage.auto` dropped. The toolbar picker is `allCases`-driven, so it disappears with the case, and both model-routing switches simplify.
- A persisted `"auto"` migrates to Hebrew *and* the key is rewritten, so the retired value doesn't linger in defaults (`current`'s `didSet` doesn't fire from `init`).
- `fromCode("auto")` → `.hebrew` via the existing unknown-code fallback.
- Recordings made under auto keep `"auto"` on disk and can still be re-transcribed, so the engines keep tolerating the literal string. Those resolve to the multilingual model — local `bestModel` default → turbo, remote → the English model id — which is the right engine for a detect pass.

## Server side

Needs [zeroabstraction/cloud-helm-configurations#11358](https://github.com/zeroabstraction/cloud-helm-configurations/pull/11358), which warms `deepdml/faster-whisper-large-v3-turbo-ct2` alongside ivrit on `mila-asr`. Order doesn't matter — this PR is a no-op until the field is set, and that PR is a no-op until a client asks for the second model.

Why that model: speaches serves faster-whisper/CT2 only, so NeMo models (Parakeet, Canary) that top the Open ASR leaderboard aren't an option without a different serving stack. Turbo over full `large-v3` because the live path posts one VAD utterance at a time — round-trip is ~0.8 s today and non-turbo (32 decoder layers vs 4) would visibly lag the live pane for a marginal English WER difference.

## Testing

- `make build` and `xcodebuild build-for-testing` — both clean, so no stale `.auto` references anywhere including the test and UI-test targets.
- 5 new tests in `RemoteTranscriptionTests`: empty-english-model preserves single-model behaviour across all languages, `he`/`iw`/`en`/`en-US`/legacy-`auto` routing, whitespace trimming (whitespace-only == unset), `currentConfig(for:)` baking in the right id, persistence across instances.
- 4 new tests in `RecordingLanguageSettingsTests`: `allCases == [.hebrew, .english]` (also guards the picker's contents), `fromCode("auto")` → Hebrew, persisted `"auto"` migrates *and* rewrites the key, and a fresh install doesn't write a language at all.
- `ModelManagerTests`' `bestModel(for: "auto") == .openaiTurbo` still holds — that's now the default branch rather than a special case.
- Tests not run locally: the host app was mid-recording and a run would launch a second instance. CI is the first real execution.
- End-to-end verification of the English path is blocked until the CHC PR lands, since the server only serves ivrit today.

## Summary by CodeRabbit

- **New Features**
  - Added an optional **English model** override for remote transcription, with language-aware routing (English/auto → English model when set; otherwise uses the primary model).
  - Exposed the English model setting in configuration import/export and the remote transcription settings UI.
- **Bug Fixes**
  - Remote transcription now selects the correct backend model based on the recording language.
  - Retired/unknown language values now reliably resolve to Hebrew.
- **Documentation**
  - Updated self-hosted server docs and examples for a two-model deployment (Model + English model), including configuration and warm-up behavior.
- **Tests**
  - Added per-language routing and legacy language decoding/migration coverage for remote transcription and language settings.

### DIFF
--- a/Mila/Models/MilaConfig.swift
+++ b/Mila/Models/MilaConfig.swift
@@ -26,7 +26,8 @@ struct MilaConfig: Codable, Equatable {
     var version: Int
 
     var remoteTranscription: RemoteTranscription?
-    /// ISO-style language code for new recordings: `"he"`, `"en"`, or `"auto"`.
+    /// ISO-style language code for new recordings: `"he"` or `"en"`. Anything
+    /// else — including the retired `"auto"` — is read as Hebrew.
     var recordingLanguage: String?
     var liveAI: LiveAI?
     var diarization: Toggle?
--- a/Mila/Models/MilaConfig.swift
+++ b/Mila/Models/MilaConfig.swift
@@ -43,6 +43,10 @@ struct MilaConfig: Codable, Equatable {
         var endpoint: String?
         /// Model id the server expects, e.g. `ivrit-ai/whisper-large-v3-turbo-ct2`.
         var model: String?
+        /// Optional model id for English and auto-detect recordings, for servers
+        /// whose main model is language-specific (the ivrit.ai finetune is
+        /// Hebrew-only). Absent or empty means `model` handles every language.
+        var englishModel: String?
         /// Bearer token. Stored in the Keychain on apply, never in UserDefaults.
         var apiKey: String?
     }
--- a/Mila/Models/MilaConfigImporter.swift
+++ b/Mila/Models/MilaConfigImporter.swift
@@ -117,6 +117,10 @@ final class MilaConfigImporter: ObservableObject {
             if let model = r.model, model != remote.model {
                 out.append(.init(label: "Model", value: model))
             }
+            if let englishModel = r.englishModel, englishModel != remote.englishModel {
+                out.append(.init(label: "English model",
+                                 value: englishModel.isEmpty ? "Same as Model" : englishModel))
+            }
             if let apiKey = r.apiKey, apiKey != remote.apiKey {
                 out.append(.init(label: "API key", value: Self.mask(apiKey)))
             }
@@ -156,6 +160,7 @@ final class MilaConfigImporter: ObservableObject {
             // stale stored one.
             if let endpoint = r.endpoint { remote.endpoint = endpoint }
             if let model = r.model { remote.model = model }
+            if let englishModel = r.englishModel { remote.englishModel = englishModel }
             if let apiKey = r.apiKey { remote.apiKey = apiKey }
             if let enabled = r.enabled { remote.backend = enabled ? .remote : .local }
         }
--- a/Mila/Models/RecordingLanguageSettings.swift
+++ b/Mila/Models/RecordingLanguageSettings.swift
@@ -53,9 +53,17 @@ enum RecordingLanguage: String, CaseIterable, Identifiable, Codable {
     /// before the rename) and for the retired `"auto"`.
     static func fromCode(_ code: String) -> RecordingLanguage {
         let normalized = code.lowercased()
-        if normalized == "iw" || normalized.hasPrefix("he") { return .hebrew }
-        if normalized.hasPrefix("en") { return .english }
+        if normalized == "iw" || isVariant(normalized, of: "he") { return .hebrew }
+        if isVariant(normalized, of: "en") { return .english }
         return .hebrew
+    }
+
+    /// Whether `code` is `base` or a regional form of it (`en-US`, `en_GB`).
+    /// A bare `hasPrefix` would also swallow unrelated codes that merely start
+    /// with those letters — `"enochian"` must fall back to Hebrew like any
+    /// other unrecognised value, not pick the English model.
+    private static func isVariant(_ code: String, of base: String) -> Bool {
+        code == base || code.hasPrefix("\(base)-") || code.hasPrefix("\(base)_")
     }
 }
 
--- a/Mila/Models/RecordingLanguageSettings.swift
+++ b/Mila/Models/RecordingLanguageSettings.swift
@@ -7,18 +7,17 @@ import Combine
 /// digging into Settings.
 ///
 /// Persisted to `UserDefaults` so the choice sticks across launches.
+/// The language is always an explicit choice — there is deliberately no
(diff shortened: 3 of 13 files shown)


urisland:
Thanks — all four actionable findings fixed in 98986c3:

- **`STT_MODEL_TTL` missing** (Major): correct, and the comment was the thing lying. The example manifest now sets `STT_MODEL_TTL=-1`. Worth keeping rather than dropping the claim: the model cache is an `emptyDir`, so an evicted model *re-downloads* rather than reloading from disk.
- **Unbounded `postStart` curls** (Major): fixed, all three now carry `--connect-timeout 5` plus `--max-time` (10 for the health poll, 600 for each model register). Matches what the internal chart already does.
- **`hasPrefix("en")` over-matching** (Minor): real — `"enochian"` resolved to English and would have picked the English model id, against the documented unrecognised → Hebrew contract. Now matches `en`, `en-US`, `en_GB` only, same tightening for `he`, with a regression test.
- **Empty English model fallback** (Minor): stated explicitly in the quickstart.

The remaining note about non-English languages other than Hebrew selecting **Model** — you flagged it as addressed by b46bf12, and it is: with Auto-detect gone `RecordingLanguage` is Hebrew or English only, and every unrecognised code resolves to Hebrew, so there's no third case for an operator to reason about.

urisland:
@coderabbitai review

PR #125, opened 2026-07-28
feat(live): delete individual lines from the live transcript

ValeroK:
Part of the #99 split (roadmap in that PR). Independent of the other live-recording slices; touches `LiveTranscriber` and one view.

## What

Each line in the live transcript pane gets a delete affordance — a trash button revealed on hover, plus a right-click "Delete line" context menu item.

## Why

Whisper occasionally emits a garbage line mid-meeting (a cough transcribed as words, a hallucinated phrase during silence, a snippet of someone else's conversation). Right now the only fix is to wait until the recording ends and edit the saved transcript. Removing it live keeps the pane readable, and — because the live transcript is what gets fed onward — keeps the junk out of the saved transcript and anything downstream of it.

## How

`LiveTranscriber.removeSegment(id:)` records the segment's absolute time range in `deletedRanges`, removes it from `segments`, and recomputes `fullText`. Incoming whisper output that overlaps a deleted range is then skipped.

That suppression is the non-obvious part and the reason this isn't just an array removal: the **fixed-window path re-transcribes a rolling buffer on every tick**, so without it the deleted line simply reappears a second later. The VAD path emits each utterance once, so there it's a safety net rather than load-bearing.

Because `segments` is `@Published`, the deletion propagates for free — the app's feed loop picks it up, and the Stop-time snapshot excludes the deleted line, so it never reaches the saved transcript.

One layout detail worth flagging: the delete button forces its own `.leftToRight` environment so the trash icon stays on the trailing edge for both RTL and LTR lines, and it's always in the layout (just transparent until hover) so rows don't jump as the pointer moves across them.

## Tests

Three cases in `LiveTranscriberTests`:

- removal drops the line and recomputes `fullText`
- an unknown id is a no-op
- **a deleted line stays gone across a subsequent re-transcription tick** — the regression this feature is really about

18 tests pass in the suite, and `make build` succeeds.

## Note on verification

I could not run the **UI test** target locally — the XCUITest runner fails to launch on my machine with "Timed out while enabling automation mode", which reproduces on a clean checkout of `main`, so it looks environmental rather than caused by this change. A reviewer sanity-check of the hover interaction would be welcome.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete individual lines from live transcripts using a hover-revealed button or context menu.
  - Deleted transcript lines are removed from the full transcript and remain suppressed if matching audio is processed again.
  - Deleted-line tracking resets when a new recording starts.

- **Bug Fixes**
  - Prevented deleted transcript lines from reappearing during subsequent transcription updates.
  - Corrected transcript timing so deleting one segment does not suppress adjacent speech.
  - Safely handles deletion requests for unavailable transcript lines.

### DIFF
--- a/Mila/Audio/LiveTranscriber.swift
+++ b/Mila/Audio/LiveTranscriber.swift
@@ -112,6 +112,13 @@ final class LiveTranscriber: ObservableObject {
     /// amount so `windowAbsoluteStart` in `runOnce` keeps producing
     /// correct absolute timestamps for the segment merge.
     private var samplesDropped: Int = 0
+    /// Absolute time ranges of segments the user deleted from the live
+    /// pane. Incoming whisper output overlapping any of these is dropped so
+    /// a just-deleted line can't reappear on a later tick — the fixed-window
+    /// path re-transcribes a rolling buffer and would otherwise re-emit the
+    /// deleted content. The VAD path emits each utterance once, so this is
+    /// mostly a safety net there. Cleared on `start()`.
+    private var deletedRanges: [(start: Double, end: Double)] = []
     private var inFlight: Task<Void, Never>?
     private var tickTask: Task<Void, Never>?
     private var language: String = "he"
@@ -140,6 +147,7 @@ final class LiveTranscriber: ObservableObject {
         self.language = language
         self.buffer.removeAll(keepingCapacity: true)
         self.samplesDropped = 0
+        self.deletedRanges = []
         self.fullText = ""
         self.segments = []
         self.speakerNames = [:]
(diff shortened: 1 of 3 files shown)


ValeroK:
@urisland Rebased onto current `main` (resolved conflicts with the pause/resume work) and re-ran the relevant tests — `LiveTranscriberTests` is green (19/19). Ready for another look when you have a moment.

ValeroK:
Rebased onto current \`main\` and addressed the open review comments:

- **VAD pad clamp (CodeRabbit major):** incoming VAD segment ends are clamped to the unpadded utterance before suppression/storage, with a regression test that a padded-end delete does not drop the next utterance.
- **Trash placement (CodeRabbit):** outer HStack stays LTR so the control is always the physical trailing child; inner stack owns RTL for speaker/text.
- **Hover hit-testing (Copilot):** button stays in layout but uses \`allowsHitTesting(hovering)\` and \`accessibilityHidden(!hovering)\`.
- **Tests (CodeRabbit):** \`try XCTUnwrap\` instead of \`try!\`.

\`LiveTranscriberTests\` is green (20/20).

ValeroK:
@urisland Rebased onto current `main` and addressed the review comments. Tests are green — waiting on you to merge this to `main` when you're ready.

ValeroK:
@urisland Rebased onto current `main` again. The earlier review comments (VAD pad clamp, hover hit-testing, RTL trash placement, `try XCTUnwrap`) are already in this branch. `LiveTranscriberTests` is green. Ready to merge when you are.

urisland:
Thank you for this, @ValeroK — and sorry it sat for as long as it did.

This is being relayed to an in-repo branch as #229, tracked by #222. The reason is purely mechanical and has nothing to do with the work: a fork PR can't get an automated review under the contributor's own quota, and its CI needs per-run approval on every single run, so a fork branch can't converge to green here. The same thing was done for #129 → #164 and #198 → #204.

**Both of your commits are cherry-picked unmodified, so you are the `Author` on each of them** — the cherry-pick onto today's `main` was clean, zero conflicts, and the relayed diff is byte-identical to this PR's. Nothing was rewritten or reimplemented. Everything added on top sits in separate commits and is a guard, a protocol fix and tests, not the feature; those commits carry a `Co-authored-by` trailer for you as well.

Two things the review turned up that are worth reading, both downstream of your change rather than in it:

1. With auto-segment (VAD) **off**, `finalizeTail` enqueues the WAV for a full batch pass that overwrites the saved segments wholesale — so in that mode a deleted line came back into `recordings.json`, the `.txt` sidecar and the `.srt`. The relay hides the affordance in that mode; making the deletion survive re-transcription instead is noted as a follow-up.
2. `get_live_transcript`'s incremental cursor is append-only by construction (a count plus a "re-read the last entry" rule), so an MCP client that had already fetched the deleted line was never told it went away, and its cursor drifted by one. The relay adds a removal marker to the live snapshot so that poll gets the full corrected set.

Neither is a criticism of the diff — your suppression logic and the `absEnd` clamp are the hard parts and they're right. They're the sort of thing only a trace through the persistence and MCP paths surfaces.

Closing this as superseded by #229. Please do take a look there; the credit line at the top is yours, and if you disagree with the mode gate or want the batch-pass follow-up, that's the place to say so.


---

PR #126, opened 2026-07-28
fix(live): keep the transcript pinned to the newest words

ValeroK:
Part of the #99 split (roadmap in that PR). Smallest of the live-recording slices — one view, no API changes.

## The bug

During a long recording the live transcript drifts out of view and you have to keep scrolling by hand to see what's being said.

Two separate causes, both in the same `onChange`:

1. **It only fires on a new segment.** The scroll is keyed on `transcriber.segments.count`, but whisper extends the *in-progress* last segment before closing it — so during the stretch where the newest line is growing in place, the count never changes and no scroll fires at all. That's the common case, not the edge case.

2. **Even when it fires, it undershoots.** `scrollTo` is called synchronously inside `onChange`, so it runs in the same update cycle as the change that triggered it and targets the **pre-update** content height. It lands just above the newest content. Under `LazyVStack` this is worse, because the row that was just appended hasn't been measured at that instant.

## The fix

- Also observe `transcriber.fullText`, which is `@Published` and recomputed as the in-progress segment expands — that covers the grow-in-place case.
- Route both triggers through a `scrollToBottom(_:)` helper that hops to the next runloop via `DispatchQueue.main.async`, so the changed content is laid out **before** we scroll.

The existing `bottom-anchor` (a 1pt `Color.clear` placed after the `LazyVStack`, so it's non-lazy and always measured) remains the scroll target — it's a stable target once layout settles.

## Verification

`make build` succeeds and the unit suite is unaffected (no logic outside the view).

This one is inherently visual, and I could not run the **UI test** target locally — the XCUITest runner fails on my machine with "Timed out while enabling automation mode", which reproduces on a clean checkout of `main`, so it looks environmental. Worth a reviewer eyeballing a real recording to confirm the pane tracks the newest words without fighting manual scrolling.


ValeroK:
Pushed a follow-up commit that debounces the scroll.

While preparing the folder-controls PR I hit the reason the naive version isn't quite enough: `fullText` changes on *every* partial-transcript tick, many per second as the in-progress line grows. Now that the scroll is keyed on it, a fresh `withAnimation` per tick stacks animations and adds main-thread work to the transcript re-render. It's coalesced to one scroll per 100 ms window instead, which also still gives the changed content a chance to lay out before we scroll.

urisland:
@coderabbitai review

urisland:
@coderabbitai full review

PR #127, opened 2026-07-28
feat(recording): choose a folder and meeting name for a recording

ValeroK:
Part of the #99 split (roadmap in that PR). Largest of the live-recording UX slices, but self-contained.

## What

Destination controls in the three places you'd naturally name or file a recording:

- **Home** — a "Save to" folder picker and an optional meeting name, so both are set *before* you start.
- **Live recording header** — the same pair, editable mid-recording. Both are applied when the recording is saved, so changing either at any point up to Stop takes effect.
- **Recording detail view** — an inline folder menu in the meta line, mirroring the sidebar's context-menu assignment but reachable from the open recording.
- **Post-stop rename sheet** — a folder picker with an inline "New Folder…" field.

It also fixes a bug: **the folder was lost when a recording was renamed after stop.**

## Deliberate design choices, in case you'd decide differently

**The folder and meeting name are per-session, not persisted.** `nextRecordingFolder` / `nextRecordingTitle` reset on launch and the title clears after each save. A folder picked for one meeting silently becoming the default for the next felt like the worse failure mode — you'd only notice weeks of recordings landed in the wrong place after the fact. There's a test pinning this (`test_nextRecordingFolder_defaults_to_all_transcriptions`) so it can't regress by accident. Easy to flip if you'd rather it were sticky.

**`resolvedRecordingTitle` is pure and `static`** so the fallback-to-date-stamped-default logic is unit-testable without standing up a recording session.

## The isolation, and why it's in this PR

A SwiftUI `Menu` is backed by an NSMenu that runs a **nested modal event-tracking runloop** while it's open. If that menu lives in a view body that re-renders at transcript or progress cadence, the menu host gets reconciled mid-tracking and the whole app beachballs.

Every one of the folder menus above sits in exactly such a body — the live recording screen, the detail view during an active transcription, the rename sheet while the batch transcribe runs. So adding the feature naively *introduces* a hang. The controls are pulled into `Mila/Views/RecordingFolderControls.swift` as leaves that observe only `RecordingStore` and `QuickActionsController` — never `LiveTranscriber` / `TranscriptionService` / `LiveAISession`. Same rationale as the existing `RecordingElapsedLabel` isolating `RecordingSession`.

For the same reason the rename sheet no longer holds `@EnvironmentObject transcription`. The status row is extracted into a private `RenameTranscriptionStatusRow`, so progress ticks re-render that small leaf instead of the entire sheet (including its folder `Menu`).

I'd rather not split the isolation out into its own PR — on its own it's a refactor with no visible motivation, and landing the feature without it ships a known hang.

## The rename bug

Two causes, both fixed here:

1. SwiftUI reuses the sheet's `@State` across presentations, so the `init`-seeded folder went stale. The sheet now mirrors the stored recording's folder until the user picks one in the sheet, at which point mirroring stops so their choice isn't clobbered.
2. `applyFolder()` now runs on **every** exit path (plain Save and Send-to-LLM), not just some.

## Tests

`QuickActionsControllerTests` gains coverage for the non-persistence default and both branches of `resolvedRecordingTitle` (empty/whitespace falls back, non-empty wins and is trimmed). Suite passes, and `make build` succeeds.

## Note on verification

I could not run the **UI test** target locally — the XCUITest runner fails on my machine with "Timed out while enabling automation mode", which reproduces on a clean checkout of `main`, so it looks environmental. The beachball fix in particular is worth a reviewer confirming by hand: open the folder menu on the live recording screen while a transcription is running.

## Summary by CodeRabbit

- **New Features**
  - Choose a destination folder for the next recording.
  - Add an optional meeting name before or during recording.
  - Move existing recordings between folders or create a new folder from recording details.
  - Select a folder while renaming recordings, including when saving or sending.
  - View and edit recording name and destination details during live recording.

- **Bug Fixes**
  - Recording names now ignore surrounding whitespace and use an automatic title when left blank.
  - User-provided recording names are preserved without automatic replacement.

### DIFF
--- a/Mila/Actions/ObsidianExporter.swift
+++ b/Mila/Actions/ObsidianExporter.swift
@@ -69,6 +69,51 @@ final class ObsidianExporter: ObservableObject {
         return write.written
     }
 
+    /// Undo the export for `recording`: drop it from the pending gate so a
+    /// summary still in flight can't file it later, and delete the note if one
+    /// was already written. Returns the removed file, or nil when there was
+    /// nothing in the vault for this recording.
+    ///
+    /// Called when the user throws a recording away. Export is driven off the
+    /// summarizer's completion hook, which can fire either side of a discard,
+    /// so clearing the gate alone would still leave a note behind whenever the
+    /// summary happened to land first — and "I discarded it and it's in my
+    /// vault anyway" reads as the app ignoring the discard.
+    @discardableResult
+    func discardNote(for recording: Recording) -> URL? {
+        clearPending(recording.id)
+        guard let vault = settings.vaultURL else { return nil }
+        var index = writtenIndex()
+        let key = Self.indexKey(vault: vault, id: recording.id)
+        migrateLegacyIndexEntry(&index, key: key, vault: vault, id: recording.id)
+        guard let relative = index[key] else {
+            // The migration above may have dropped a legacy entry.
+            setWrittenIndex(index)
+            return nil
+        }
+        index[key] = nil
+        setWrittenIndex(index)
+
+        let note = vault.appendingPathComponent(relative)
+        // Same rule the write path follows: never resolve a stored path back
+        // out of the vault. A path that no longer sits inside it is one we
+        // must not delete, whatever the index says.
+        guard ObsidianPathSanitizer.isContained(note, in: vault, fileManager: fileManager),
+              fileManager.fileExists(atPath: note.path) else { return nil }
+        do {
+            try fileManager.removeItem(at: note)
+        } catch {
+            print("ObsidianExporter: failed to remove \(note.path): \(error)")
+            return nil
+        }
+        if settings.gitSyncEnabled {
+            let title = Self.singleLine(recording.title)
+            let message = "Remove discarded transcript: \(title.isEmpty ? "Untitled recording" : title)"
+            kickGitSync(vault: vault, changedPaths: [note], commitMessage: message)
+        }
+        return note
+    }
+
     /// Backfill: write every provided recording into the vault, then kick a
     /// single git sync covering all of them. Used by "Sync existing
     /// transcripts". Returns the count of notes actually written.
--- a/Mila/Actions/PostRecordingCoordinator.swift
+++ b/Mila/Actions/PostRecordingCoordinator.swift
@@ -120,7 +120,15 @@ final class PostRecordingCoordinator: ObservableObject {
     /// the sheet is already showing a different recording the new one is
     /// dropped (we don't queue — concurrent voice memos are rare and a
     /// stack of sheets is worse UX than just naming the first one).
-    func present(_ recording: Recording) {
+    ///
+    /// `titleWasUserProvided` reports that the recording was saved under a
+    /// meeting name the user typed before / during recording. The background
+    /// auto-title job can only protect a user's title by comparing against the
+    /// default it expects to replace, and here that default IS the user's
+    /// title — so the guard cannot see the difference and the CLI suggestion
+    /// would overwrite the typed name. Skip the job entirely in that case; the
+    /// Suggest button in the sheet still offers it on demand.
+    func present(_ recording: Recording, titleWasUserProvided: Bool = false) {
         // UI-TEST: the record-while-finalizing E2E drives two back-to-back
         // recordings through the real `stopRecording`; a modal rename sheet
         // after each Stop would sit over Home and block the next Record tap
@@ -130,6 +138,7 @@ final class PostRecordingCoordinator: ObservableObject {
         if CommandLine.arguments.contains("--ui-test-finalize-regression") { return }
         guard pending == nil else { return }
         pending = recording
+        guard !titleWasUserProvided else { return }
         armAutoSuggestTitle(for: recording)
     }
 
--- a/Mila/Actions/PostRecordingCoordinator.swift
+++ b/Mila/Actions/PostRecordingCoordinator.swift
@@ -31,6 +31,12 @@ final class PostRecordingCoordinator: ObservableObject {
     private let transcription: TranscriptionService
     private let llm: LLMSettings
 
+    /// Late-bound by `MilaApp`, like the exporter wiring on
+    /// `QuickActionsController`. Discard has to reach the vault: export runs
+    /// off the summarizer's completion hook, which can land either side of the
+    /// user throwing the recording away.
+    var obsidianExporter: ObsidianExporter?
+
     /// Injectable LLM-run seam so tests can assert what `PostRecordingCoordinator`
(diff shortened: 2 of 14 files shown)


ValeroK:
@urisland Rebased onto current `main` (resolved conflicts with the pause/resume work — kept both the pause button and the folder/meeting-name controls) and re-ran the relevant tests — `QuickActionsControllerTests` is green (23/23). Ready for another look when you have a moment.

ValeroK:
Rebased onto current \`main\` and addressed the open review comments:

- **Clear one-shot folder (CodeRabbit major):** capture \`nextRecordingFolder\` into a local, apply it, then set it to \`nil\` with the title so the next recording in the same launch does not inherit it.
- **Rename sheet reuse (CodeRabbit major):** \`onChange(of: initialRecording.id)\` resets \`userEditedFolder\` and reseeds \`selectedFolder\`.
- **Redundant persist (Copilot):** \`store.assign\` only runs when the name is not already in \`store.folders\` (new folder or case mismatch).
- **UserDefaults in tests (CodeRabbit + Copilot):** dropped the \`UserDefaults.standard\` mutation; a fresh controller is enough.

Added tests that stop clears the folder/title and that a brand-new folder still gets registered. \`QuickActionsControllerTests\` is green (25/25).

ValeroK:
@urisland Rebased onto current `main` and addressed the review comments. Tests are green — waiting on you to merge this to `main` when you're ready.

ValeroK:
@urisland Rebased onto current `main` (resolved a conflict with the MCP live-sidecar drain from #183). The earlier review comments (one-shot folder clear, rename-sheet reuse, skip redundant `assign`) are already in this branch. `QuickActionsControllerTests` is green. Ready to merge when you are.

ValeroK:
Addressed the new CodeRabbit note: a failed `session.stop()` (nil URL) now clears `nextRecordingTitle` and `nextRecordingFolder` the same way the successful path does, so they cannot leak into the next recording.

ValeroK:
Fixed a bug found while dogfooding this branch: a meeting name typed before recording was being overwritten by the LLM auto-title.

`PostRecordingCoordinator.armAutoSuggestTitle` protects a user's title by only applying the suggestion when the recording *still* carries the default title it captured as `baselineTitle`. With this feature, a user-typed name IS the title at `store.add` time, so it becomes the baseline — the guard sees no divergence and the CLI suggestion replaces the name the user chose.

`present(_:titleWasUserProvided:)` now skips arming the job when the title came from the user. The Suggest button in the rename sheet still offers it on demand.

Two regression tests, both verified to fail without the guard (they write "An LLM Title" over "Weekly Sync"):
- `PostRecordingCoordinatorSendTests.test_present_skips_auto_title_when_the_user_named_the_recording` — the coordinator seam.
- `QuickActionsControllerTests.test_stopRecording_does_not_auto_title_a_user_named_recording` — end-to-end through `stopRecording`, guarding the wiring.

ValeroK:
Added the Discard/recovery fix found while testing this PR (`74fe137`).

What looked like Discard saving the recording was the launch recovery sweep resurrecting empty captures: AVAudioFile's empty WAV is 4096 bytes, while the old guard only skipped files below 512 bytes. Empty captures now stay discarded, while non-empty crashed recordings remain recoverable.

Discard also now clears a pending Obsidian export and removes an already-written note, covering both sides of the summarizer/export race.

Verification: `RecordingStoreTests`, `ObsidianExportSequencingTests`, and `PostRecordingCoordinatorSendTests` — 72/72 passing. The regression tests cover empty vs non-empty orphan WAVs, discard before/after export, and coordinator wiring.

urisland:
Thank you @ValeroK — this is a genuinely good piece of work, and all of it is
landing. The folder + meeting-name controls, the NSMenu isolation (including the
diagnosis that a `Menu` in a transcript-cadence body beachballs the app), the
auto-title interaction, and the folder-lost-on-rename fix are all yours,
unchanged.

I'm relaying it rather than merging this PR directly, which is purely mechanical
and nothing to do with the change: fork PRs can't be reviewed under a
contributor's own quota, and fork CI needs per-run approval on every single run,
which makes iterating on them painful for both of us. The established pattern
here is to recreate the commits on an in-repo branch with authorship preserved
(same as #198 → #204 and #129 → #164).

- Relay PR: #230
- Issue: #223 (CI requires every PR to link one — see
  `.claude/rules/pull-requests.md`; that's the only reason it exists, and it
  credits you)

Your four commits cherry-picked onto today's `main` **without a single
conflict**, and the resulting diff is byte-identical to this one — nothing had
to be reimplemented.

**Your late commit is relayed too.** You pushed `74fe137`
("stop discarded and empty captures from coming back") after I'd cut the relay
branch from `b4bce8b`, so I re-fetched this PR's head before closing it and
cherry-picked that one as well — authorship intact, the seven files byte-identical
to yours. It's an unrelated fix riding in the same PR, so #230 now carries both;
say the word if you'd rather see it split out and I'll move it to its own PR.
For the record, I read it to the same standard as the rest and it holds up: the
4 KB `AVAudioFile` header-block reservation is the right explanation for why the
512-byte guard never fired, and keeping the threshold a *byte count* rather than
asking `AVAudioFile` for a duration is the subtle part — an unfinalized header
claims zero frames, so a frame count would have discarded exactly the crashed
recordings the sweep exists to save. Your two `RecordingStoreTests` pin both
sides of that. Same for routing discard through the exporter: the note really can
be in the vault before the discard lands, and `discardNote` reusing the write
path's containment check (never resolve a stored path back out of the vault) is
the right instinct.

On top of everything I added two commits of my own:

- Three negative controls the destination edges were missing (a meeting name
  containing `..` / `/` can't reach a filesystem path; a folder deleted
  mid-recording doesn't cost the user the recording; a folder chosen at record
  time stays resolvable through `MilaStoreReader`, the reader `mila-mcp` uses),
  and a rewrite of the comment on the `store.assign` call in `stopRecording` — it
  described a case the picker can't actually produce, so it now names the one the
  UI does reach and the invariant the call keeps. No behaviour change.
- One tiny cleanup on your late commit: `emptyWAVByteCount` landed between
  `recoverOrphanRecordings`' doc comment and the function, which silently swapped
  the two `///` blocks (the crash-recovery prose became the constant's docs and
  vice versa). Moved the constant above. Comments only.

Two notes back to you, both credit rather than criticism:

- Your call to make `nextRecordingFolder` / `nextRecordingTitle` per-session and
  not persisted was the right one, and I've kept it. I've noted in the relay PR
  the one edge it leaves — deleting the chosen folder mid-recording recreates it
  — and argued for keeping your behaviour, since "Save to <name>" was on screen
  right up to Stop.
- The XCUITest failure you hit ("Timed out while enabling automation mode") does
  reproduce on clean checkouts; it isn't yours. CI scopes those out for the same
  reason.

Closing this in favour of #230. Thanks again — and please keep them coming.


---

PR #128, opened 2026-07-28
feat(mcp): mila-mcp — an MCP stdio server over Mila's transcriptions

ValeroK:
Part of the #99 split (roadmap in that PR). Kept as one unit per your note that the MCP server and the MilaKit extraction are best reviewed together — but the first two commits are a self-contained refactor and I'm happy to split them out if you'd rather land those first.

## What

An MCP stdio server, `mila-mcp`, embedded at `Mila.app/Contents/MacOS/mila-mcp`. It gives an agent read-only access to your transcriptions: list, search, fetch a transcript (speaker-named), and read the transcript of a recording **that's still in progress**.

Setup and the full tool reference are in `docs/mcp.md`. `skills/mila-meetings/SKILL.md` is an agent skill built on top of it.

## Commits, in review order

1. **Extract MilaKit; move TranscriptFormatter behind `SpeakerTextSegment`** — pure refactor, no behaviour change. Formatting logic moves into the new package behind a protocol so both the app's `TranscriptSegment` and the helper's `StoredRecording.Segment` can use it, instead of the helper reimplementing it and drifting.
2. **Store-location pointer + read-only store access** — `RecordingStore` writes `store-location.json`; `MilaStoreReader` reads recordings.json without the app running.
3. **Live-transcript sidecar** — mirrors the in-progress transcript to `live/current.json` during recording.
4. **The `mila-mcp` executable** — thin stdio loop; all tool logic lives in `MilaMCPToolHandlers`.
5. **Docs, agent skill, Makefile target, CLAUDE.md notes.**

## Design constraints worth knowing about

**MilaKit is dependency-free, deliberately.** It must not pull in TranscriptionCore, because it links into `mila-mcp` and that helper must not drag the whisper xcframework into its binary. This is why `TranscriptFormatter` moved behind a protocol rather than just being made public where it was.

**Tool logic lives in `MilaMCPToolHandlers`, not in the executable.** The handlers are pure JSON-in/JSON-out, so they're testable without spawning a process — `MilaMCPToolHandlersTests` covers them directly.

**The two cross-process contracts live at the DEFAULT app-support root and deliberately do not travel with a relocated recordings folder.** If they moved with the store, the helper would have no fixed place to start looking. `store-location.json` is the indirection that makes relocation work at all.

**`StoredRecording` is a mirror of recordings.json and can drift.** Anything `Recording.encode(to:)` writes has to be mirrored there. `StoredRecordingDriftTests` is the tripwire for that, and `StoredRecording` decodes leniently (`decodeIfPresent` + defaults) so an older helper survives a newer app's schema rather than failing to parse.

## Tests

- **MilaKit package: 33 tests** — `MilaStoreReaderTests` (pointer resolution and relocation, sorting, filters, search relevance and snippets, speaker-name resolution, sidecar-vs-segments transcript fallback), `MilaMCPToolHandlersTests`, `TranscriptFormatterTests`.
- **App: 61 tests** across `LiveTranscriptSidecarWriterTests`, `StoredRecordingDriftTests`, `RecordingStoreTests`, `TranscriptFormatterTests`.

Both `make build` and `make package-test` pass.

## One note on the diff

The sidecar's per-tick call maps `LiveSegment -> TranscriptSegment` inline in `MilaApp`. In our fork that mapping was centralised as `LiveTranscriber.transcriptSegments`, but that property is @DanielSolomon's work from #85, so I've left it out rather than duplicate his commit here. If #85 lands first, this call site should switch to `transcriber.transcriptSegments` and the inline map goes away.

## Note on verification

I could not run the **UI test** target locally — the XCUITest runner fails on my machine with "Timed out while enabling automation mode", which reproduces on a clean checkout of `main`, so it looks environmental. Unit and package tests are verified.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,6 +15,8 @@ A macOS (Swift/SwiftUI) local transcription app built on whisper.cpp, with optio
   - `Mila/Resources/DiarizationModels/` — bundled pyannote speaker diarization model weights (~31 MB)
   - `MilaTests/` — unit tests
   - `Packages/TranscriptionCore/` — cross-platform Swift package: WhisperEngine (whisper.cpp bindings), WAVReader, WER calculator, and E2E transcription test fixtures
+  - `Packages/MilaKit/` — zero-dependency Swift package shared by the app and the `mila-mcp` helper: TranscriptFormatter, the read-only `StoredRecording` mirror of recordings.json, `MilaStoreReader`, the live-transcript snapshot schema, and the MCP tool handlers
+  - `MilaMCP/` — the `mila-mcp` executable (MCP stdio server over Mila's transcriptions), embedded at `Mila.app/Contents/MacOS/mila-mcp`; see `docs/mcp.md`
   - `scripts/` — release/build scripts (make-dmg.sh, etc.)
 
 ## Conventions
@@ -46,9 +48,14 @@ These patches live in `SpeakerDiarizer.swift`'s inline diarize script. If upgrad
 - For verification/setup state that should survive app restarts, persist a `verified` flag alongside the verified parameter values (path). On launch, restore only if current values match the persisted ones.
 - Computed `status` properties must check `verificationStatus` before `lastVerifyResult` -- the persisted verified state should take precedence over nil in-memory verify results on launch.
 
+### MCP server (mila-mcp) and MilaKit
+- The app persists two cross-process contracts for the embedded MCP helper: `store-location.json` (written by `RecordingStore` on init + relocate — where recordings.json currently lives) and `live/current.json` (the live-transcript sidecar written during recording by `LiveTranscriptSidecarWriter`). Both live at the DEFAULT app-support root and deliberately do not travel with a relocated recordings folder.
+- **Any change to what `Recording.encode(to:)` writes into recordings.json must be mirrored in MilaKit's `StoredRecording`** — `StoredRecordingDriftTests` in MilaTests is the tripwire. Keep `StoredRecording` decoding lenient (`decodeIfPresent` + defaults) so an older helper survives a newer app's schema.
+- MilaKit must stay dependency-free (in particular: no TranscriptionCore) — it links into `mila-mcp`, which must not drag in the whisper xcframework. Tool logic lives in `MilaMCPToolHandlers` (pure JSON), not in the executable.
+
 ### Tests
 - `TranscriptionService` now requires a `diarizationSettings:` parameter. In tests, always pass `DiarizationSettings(defaults: .init(suiteName: "TestClassName.diarization")!)` to isolate from user defaults.
-- Run tests with `make test` or via Xcode.
+- Run tests with `make test` or via Xcode. Package tests: `make package-test` (TranscriptionCore + MilaKit).
 
 ## Release Process
 - **Release notes are REQUIRED, first.** Every release must add
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ help:
 	@echo "  models-coreml-tiny - Download ggml-tiny + sibling -encoder.mlmodelc into ~/.cache/whisper-coreml-test/ (for CI ANE verification test)"
 	@echo "  dmg           - Build a release DMG (VERSION=<x.y.z>) suitable for upload"
 	@echo "  e2e           - Run E2E transcription tests (requires ggml-tiny.bin)"
-	@echo "  package-test  - Run TranscriptionCore package unit tests"
+	@echo "  package-test  - Run TranscriptionCore + MilaKit package unit tests"
 	@echo "  bundle-diarization - Produce the bundled Python + pyannote.audio runtime under Mila/Resources/PythonRuntime/"
 	@echo "  clean         - Remove generated project and build artifacts"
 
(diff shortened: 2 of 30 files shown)


urisland:
**Who's writing:** I'm Claude, working with @urisland on the repo. He asked me to post the current state of thinking here so it's on the record rather than sitting in a chat. To be explicit about attribution: **the architecture proposal below is his idea**, the audit findings and the pros/cons are my analysis, and **the decision is still open — he wants to think it over and will come back to it soon.** Nothing here is a request for you to change anything yet.

## Where this PR stands

@ValeroK — the feature is **approved in principle**, with two conditions he set: it must be **off by default**, and when on it must be **local-only**. What's parked is not *whether*, it's *which architecture*.

His question, in his words: *why is it not an MCP stdio server that uses an API that Mila will have that listens locally? Then we enable/disable an API, and whoever wants to use an MCP binary can do it regardless — and if Mila upgrades, the older MCP can still work.*

The point being that as written, `mila-mcp` reads `recordings.json` directly and ships inside the app bundle, so it's version-locked to Mila's internal file formats by construction. Anyone else's client breaks when the schema moves.

## Audit findings (facts, useful regardless of the outcome)

I went through the SDK and the layering. Four things worth having in writing:

1. **No listener exists anywhere today.** `NWListener` has zero matches in `modelcontextprotocol/swift-sdk` 0.12.1. Of its five transports, none owns a socket — both `Stateful`/`StatelessHTTPServerTransport` are *parsing adapters* (`handleRequest(HTTPRequest) -> HTTPResponse`), so a host must bring its own HTTP server. `MilaMCPMain.swift` constructs `StdioTransport()` and nothing else. The current design is stricter than "localhost only": it's parent-process only.
2. **NIO is not linked into `mila-mcp`.** It's a dependency of the package's conformance *executable*, not of the `MCP` library target — the `project.yml` comment saying the SDK's NIO deps are the reason for the pin is inaccurate and worth correcting whenever this lands.
3. **Your layering holds up well.** `MilaMCPToolHandlers` is decoupled from SDK types as you said. It isn't decoupled from the filesystem — but the coupling is 5 call sites in the handlers and 3 disk touches in `MilaStoreReader`. An API-backed source would need to serve exactly four things: the recording list, one transcript, the live snapshot, and location resolution (which disappears entirely under an API). Extract a data-source protocol and the handler bodies don't change; the 33 MilaKit tests keep passing against a file-backed implementation. **Credit where due — that's why switching architecture is a swap, not a rewrite.**
4. **No entitlement needed** for a listener — the app isn't sandboxed. But binding TCP can raise the macOS "accept incoming connections?" prompt; loopback-only and unix sockets avoid it.

## The four options

**A. Keep as-is — stdio + direct file reads, plus a consent gate.**
*Pros:* zero new surface; works when Mila is quit; parent-process-only reachability is the strongest isolation of the four; already written.
*Cons:* stays version-locked to `recordings.json` — the objection that started this. Keeps both cross-process files on disk (`store-location.json`, `live/current.json`) and their drift-mirror tripwire. "Off" means a binary politely declining to read files it can still read.

**B. Mila exposes a local API over a unix domain socket; `mila-mcp` becomes a thin client.**
*Pros:* the proposed architecture with the safest transport — `~/Library/Application Support/Mila/api.sock` at `0600`, unreachable from browser JavaScript, no port to scan, no firewall prompt, access control is filesystem permissions. **Deletes both cross-process contracts** and the whole older-helper-vs-newer-app schema problem this PR spends real effort defending against. Third parties get a versioned contract.
*Cons:* the app must be running. `curl` needs `--unix-socket`. Mila owns a socket it doesn't own today.

**C. Local API over `127.0.0.1` TCP + bearer token.**
*Pros:* same decoupling as B, plus ordinary HTTP clients (curl, Raycast, Shortcuts) work with no ceremony. The SDK already ships `OriginValidator.localhost` for DNS-rebinding, and its conformance server defaults to `127.0.0.1`.
*Cons:* any local process — including a web page you happen to be visiting — can reach the port, so the token becomes load-bearing security rather than convenience. Possible firewall prompt. More surface than B for a convenience most users won't exercise.

**D. Mila hosts MCP itself over Streamable HTTP; no separate binary.**
*Pros:* any MCP client connects by URL, nothing to embed or ship. `StatefulHTTPServerTransport` implements the full spec (session IDs, SSE).
*Cons:* by far the most work — the SDK's own reference server is **432 lines of NIO glue**, or you take a third-party HTTP dependency. Inherits every con of C, and re-couples client to app version, undoing the decoupling that motivated the change.

**My lean (Claude's, not his):** B, with C as a later opt-in if HTTP clients ever matter. It's the only option that *removes* code as well as adding it, and it buys the decoupling at the lowest security cost. The tradeoff it accepts, and the thing I'd most want a human to weigh: **transcripts become unreadable while Mila isn't running**, which today they are.

## Two housekeeping notes for you

- **CodeRabbit never actually reviewed this PR** — it returned *"Review limit reached — @ValeroK, you've reached your PR review limit"*. There are zero inline comments and zero reviews on it, so the single green check is the absence of a signal rather than a pass. Worth re-triggering with `@coderabbitai review` when this moves.
- The branch is **9 commits behind `main`** but I confirmed it **rebases cleanly**, no conflicts.

**Please don't rework anything on the strength of this comment.** It's a status marker so the reasoning isn't lost. @urisland will decide the architecture and say so here.

ValeroK:
Thanks @urisland — acknowledging the architecture note and the follow-up.

Deferring to #183 (file-backed stdio + opt-in consent gate). Closing this PR as superseded; the original commits are cherry-picked there with authorship intact.

ValeroK:
Superseded by #183.

PR #143, opened 2026-08-01
fix(audio): recover capture when the input device changes mid-recording

### LINKED ISSUE #142: Recording silently stops capturing when the input device changes mid-session (26-minute meeting saved as 24 seconds)
## Symptom

A user's 26-minute Zoom meeting was saved as a recording that **plays for ~24 seconds**. The sidebar row showed the full `26:24`, the transcript simply stopped after a few sentences, and nothing anywhere said a word about it. Reported via a diagnostic bundle (`Mila-DiagnosticReport-2026-07-30T12-32-14Z`, Mila 1.9.1 build 54, macOS 15.5, Mac16,12).

Two of the three meetings recorded that day were truncated this way.

## Root cause

The `AVAudioEngine` input tap dies when CoreAudio re-points the input device under a running engine, and nothing in the app notices.

From the bundle's `logs/process.log`, recording `AAC8F64B`:

```
12:02:03  MicrophoneRecorder  input device: EarPods Microphone … uid=AppleUSBAudioEngine:…:EarPods:G9JK9YD399:2
12:02:03  MicrophoneRecorder  native input format: 44100.0Hz 1ch
12:02:03  MicrophoneRecorder  first mic buffer delivered — capture is live
12:02:25  LiveTranscriber     utterance … startSec=18.69          ← last sign of life
12:02:28  AUHAL   DeviceListener: Device 860 died! Looking for usable device...
12:02:28  AUHAL   FindUsableDevice: Found usable input device 87   ← built-in mic, 48kHz
12:02:28  ddagg   HAL notification: default input device changed → Built valid aggregate 872
12:02:28  AUHAL   SetProperty: caller requesting device change from 87 to 872
          …26 minutes with not one line from any Mila audio category…
12:28:27  MicrophoneRecorder  mic stopping: buffers=247 frames=394348 conversionFailures=0
```

`394348 / 16000 Hz = 24.6 s` captured across 26 minutes of wall clock. The app was alive throughout (AppKit window events at 12:06 and 12:13) — it just recorded nothing and reported nothing.

This is documented `AVAudioEngine` behaviour, from `AVAudioEngine.h`:

> When the engine's I/O unit observes a change to the audio input or output hardware's channel count or sample rate, the engine stops itself […] and issues this notification. The nodes remain attached and connected with previously set formats.

The EarPods (44.1kHz) died and CoreAudio substituted the built-in mic (48kHz) — a sample-rate change, so the engine stopped itself and posted `AVAudioEngineConfigurationChange`. **Nothing in the codebase observes that notification** (zero references repo-wide), and there is no in-session check that buffers are still arriving: `MicFrameStats` is only inspected at `stop()`, and only for the `frames == 0` case, so 24 seconds out of 26 minutes looks perfectly healthy.

### Second instance, same day, same cause

Session `09:07:18–09:29:42` (22 min wall clock) captured `frames=4508066` = **4m42s** — the recording listed as 281.8s. Trigger at `09:11:30`: `ddagg  HAL notification: default output device changed` → new aggregate `711`, and the SCStream was torn down at the same instant. The only meeting that survived intact that day (11:08) was the one on the built-in mic with no device event.

### Why meeting mode loses everything, not just the mic

In meeting mode the mic is the master clock — `consumeMic` drives both `write()` and the live-transcription feed (`RecordingSession.swift:252`). When the tap dies, nothing is written even if the system-audio leg were still delivering. The existing "mic clock stalled" safety valve in `consumeSystem` (`RecordingSession.swift:280`) never fired here because the SCK leg went quiet at the same moment, so `pendingSystem` never reached its 30s cap.

### Two smaller problems this exposed

- **The duration lies.** `stopRecording` used `max(durationBeforeStop, audioDuration(at:))` (`QuickActionsController.swift:565`), so the wall-clock timer always won and the row claimed 26:24 for a 24-second file.
- **A lost system-audio leg leaves no trace.** `SystemAudioRecorder.didStopWithError` logs via `print()`, not `Logger`, so it never reaches OSLog and never appears in a diagnostic bundle. We cannot tell from this report whether SCK reported an error at 09:11:30 or 12:02:28.

## Intended fix

1. Observe `AVAudioEngineConfigurationChange` and rebuild engine + tap in place, keeping the session's `AsyncStream`, gain controller and frame counters so recording continues into the same WAV.
2. A capture watchdog that rebuilds when the frame count flatlines, as a backstop for stalls that post no notification.
3. Restart the SCStream after an SCK-side death (bounded, and never for permission failures), and move that logging to `Logger`.
4. Store the real audio duration instead of the wall clock, and warn the user when captured audio falls far short of the time they spent recording.

## Repro

With a USB mic or wired headset selected as input, start a meeting recording, then unplug the device ~20s in. Before the fix: capture stops permanently, the timer keeps counting, and the saved recording contains only the audio from before the unplug.

urisland:
Closes #142

## The bug

A user's 26-minute Zoom meeting was saved as a recording that plays for **24 seconds**. The row showed the full `26:24`, the transcript just stopped, and nothing said a word about it. Two of their three meetings that day were truncated the same way.

From their diagnostic bundle (Mila 1.9.1 build 54, macOS 15.5):

```
12:02:03  MicrophoneRecorder  input device: EarPods Microphone … 44100.0Hz 1ch
12:02:25  LiveTranscriber     utterance … startSec=18.69         ← last sign of life
12:02:28  AUHAL   DeviceListener: Device 860 died! Looking for usable device...
12:02:28  AUHAL   FindUsableDevice: Found usable input device 87  ← built-in mic, 48kHz
          …26 minutes with not one line from any Mila audio category…
12:28:27  MicrophoneRecorder  mic stopping: buffers=247 frames=394348
```

`394348 / 16000 = 24.6s` captured across 26 minutes. The app was alive the whole time — it just recorded nothing and reported nothing.

This is documented `AVAudioEngine` behaviour (`AVAudioEngine.h`): when the I/O unit sees the input hardware change sample rate, *"the engine stops itself"* and posts `AVAudioEngineConfigurationChange`, while *"the nodes remain attached and connected with previously set formats."* The 44.1kHz EarPods were replaced by the 48kHz built-in mic, so the engine shut itself down. Nothing in the app observed that notification (zero references repo-wide), and nothing checked mid-session that buffers were still arriving — `MicFrameStats` was only read at `stop()`, and only for the `frames == 0` case, so 24s of a 26-minute meeting looked perfectly healthy.

## What changed

**`MicrophoneRecorder` — two independent safety nets**

1. `AVAudioEngineConfigurationChange` → rebuild engine + tap immediately. No "is it really broken?" probe: the engine has already stopped itself by the time the notification lands, so there's nothing to preserve by waiting.
2. A `CaptureStallDetector` watchdog → rebuild when the frame count stops growing. This is the backstop for stalls that post no notification, which is what the user's *other* truncated session looks like (`default output device changed` → new aggregate at the same rate).

Both keep the session's `AsyncStream`, gain controller and frame counters, so recording continues into the same WAV instead of ending — the stream has to survive, since `RecordingSession.micTask` is iterating it. Each rebuild builds a brand-new `AVAudioEngine`, which is mandatory for the reason already documented on the class (reusing one leaves the input node permanently silent).

Rebuilds are capped per session, and the watchdog stops watching once the budget is spent — otherwise a permanently dead input would re-log the same stall every 4s for the length of a meeting, the same trap `consumeSystem`'s overflow log already had to throttle.

**`SystemAudioRecorder`**

- Restarts its `SCStream` after an SCK-side death (bounded; never for permission failures, since a revoked grant fails identically every time and just buries the real reason).
- `Logger` instead of `print`. `print` never reaches OSLog, which is why this user's bundle contains no trace of the app-audio leg dying — only ScreenCaptureKit's own internal lines, which don't say who asked or why. Worth calling out: **the system-leg root cause is not fully confirmed** by this bundle. The mic side is proven; this logging change is what will settle the SCK side next time.

**`QuickActionsController`**

- Duration now comes from the audio file rather than `max(wallClock, fileDuration)`. The timer always won, which is why the row claimed 26:24 for a 24-second file.
- Capture that dies part way through now warns the user instead of looking like a bad transcription. Exempt for `.systemAudio`, where SCK legitimately delivers nothing while no audio is playing.

## Testing

19 new tests. The stall detector, the short-capture policy and the SCK restart policy are all pure functions with injected clocks/inputs, so they cover the decisions exactly — no audio device, no permission grants, no sleeping. Watchdog wiring (rebuild happens, stream survives it, cap is respected, `stop()` cancels it) is driven through the existing `bringUpOverride` seam.

```
Executed 39 tests, with 0 failures     # new + the two suites this touches
```

Local `make build` and `build-for-testing` are clean; the only warnings left in the touched files are pre-existing.

Not verifiable in CI: the real device-yank path. Worth a manual check on the beta — start a meeting recording on a USB mic or wired headset, unplug it ~20s in, and confirm the recording keeps going (the log should show `mic capture rebuilt`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recording duration accuracy by prioritizing the captured audio file’s length.
  * Added detection and user-facing alerts for substantial microphone or meeting capture loss.
  * Improved microphone recovery when audio capture stalls or system configuration changes.
  * Added automatic, limited recovery for temporary system-audio stream failures.
  * Prevented misleading warnings for short recordings, system-audio-only captures, and empty files.

### DIFF
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -567,9 +567,12 @@ final class QuickActionsController: ObservableObject {
         // the wall clock is a claim the user can't act on — the diagnostic
         // report that prompted this showed a 26:24 row whose audio was 24
         // seconds long, because `max()` always let the timer win. Fall back to
-        // the wall clock only when the file can't be read at all.
-        let capturedDuration = audioDuration(at: outputURL)
-        let duration = capturedDuration > 0 ? capturedDuration : durationBeforeStop
+        // the wall clock only when the file can't be decoded at all — a
+        // readable file reporting zero frames is a real answer, not a failure
+        // to measure, and must not be re-inflated to the full wall clock.
+        let decodedDuration = audioDuration(at: outputURL)
+        let capturedDuration = decodedDuration ?? 0
+        let duration = decodedDuration ?? durationBeforeStop
         let (title, source, appName): (String, RecordingSource, String?) = {
             switch captured {
             case .recordingMic:
@@ -1203,8 +1206,16 @@ final class QuickActionsController: ObservableObject {
         return "\(prefix) · \(f.string(from: Date()))"
     }
 
-    private func audioDuration(at url: URL) -> Double {
-        guard let file = try? AVAudioFile(forReading: url) else { return 0 }
+    /// Length of the audio actually on disk, or `nil` if the file can't be
+    /// decoded at all.
+    ///
+    /// The optional matters: a readable WAV containing zero frames is a real
+    /// answer ("we captured nothing"), and must not be confused with "we
+    /// couldn't tell". Returning 0 for both let a completely empty recording
+    /// fall back to the wall clock and be saved at full length — the exact
+    /// lie this change set out to remove.
+    private func audioDuration(at url: URL) -> Double? {
+        guard let file = try? AVAudioFile(forReading: url) else { return nil }
         return Double(file.length) / file.processingFormat.sampleRate
     }
 }
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -562,7 +562,14 @@ final class QuickActionsController: ObservableObject {
             activeJob = .none
             return
         }
-        let duration = max(durationBeforeStop, audioDuration(at: outputURL))
+        // Prefer what's actually on disk over the wall clock. For a healthy
+        // recording the two agree; when they don't, the file is the truth and
+        // the wall clock is a claim the user can't act on — the diagnostic
+        // report that prompted this showed a 26:24 row whose audio was 24
+        // seconds long, because `max()` always let the timer win. Fall back to
+        // the wall clock only when the file can't be read at all.
+        let capturedDuration = audioDuration(at: outputURL)
+        let duration = capturedDuration > 0 ? capturedDuration : durationBeforeStop
         let (title, source, appName): (String, RecordingSource, String?) = {
             switch captured {
             case .recordingMic:
@@ -594,6 +601,21 @@ final class QuickActionsController: ObservableObject {
         if source == .microphone, session.lastMicFrameCount == 0 {
             transcription.lastError = "Microphone captured no audio. Check System Settings ▸ Privacy & Security ▸ Microphone, and that the input selected in Settings ▸ Audio Input isn't muted, disconnected, or in use by another app."
             quickActionsLog.error("mic-only recording captured 0 frames — surfaced audio-input guidance to the user")
+        }
+
+        // Capture that died PART way through used to be completely silent: the
+        // row showed the full wall-clock length and the transcript just stopped
+        // early, so a lost meeting looked like a bad transcription. Say it out
+        // loud instead. `transcription.lastError` is only set when nothing more
+        // specific claimed it (the 0-frame case above is strictly better
+        // guidance for its own situation).
+        if Self.capturedAudioFellShort(source: source,
+                                       wallClock: durationBeforeStop,
+                                       captured: capturedDuration) {
+            quickActionsLog.error("captured only \(capturedDuration, privacy: .public)s of audio for a \(durationBeforeStop, privacy: .public)s recording (source=\(source.rawValue, privacy: .public), micRebuilds=\(self.session.mic.restartCount, privacy: .public)) — capture died mid-session")
+            if transcription.lastError == nil {
+                transcription.lastError = "Only \(formatDuration(capturedDuration)) of audio was captured for a \(formatDuration(durationBeforeStop)) recording — capture stopped early. This usually means the input device changed or went away mid-recording (headphones unplugged, a USB mic removed, or another app taking over the device). Pinning a specific input in Settings ▸ Audio Input makes it less likely."
+            }
         }
 
         // ---- IMMEDIATE: build a tentative Recording from whatever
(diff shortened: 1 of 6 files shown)


urisland:
Thanks — all five taken. Summary of what changed in c9978c9:

**Orphaned capture resources after `stop()`** (mic + SCK) — agreed, and the most serious of the five. Both bring-ups suspend for seconds and `stop()` runs freely on the main actor meanwhile, which is exactly the sequence when a user hits Stop *because* their headphones just died. Both paths now re-check session intent after the await and release the orphan (tap removed + engine stopped off-main; `stop()` for the stream).

**Strong `self` in the watchdog loop** — agreed. Re-bound weakly per iteration; this also matches the shape `RecordingSession.timerTask` already uses.

**Detector not re-armed after a notification-driven rebuild** — agreed. The window is narrower than the comment implies (a successful rebuild resumes frames within milliseconds, so the next tick usually sees growth and re-arms), but it is real whenever a rebuild runs longer than `captureStallTimeout`, and the invariant is worth making explicit. The watchdog now re-arms on any change to `restartCount`.

**`audioDuration` conflating undecodable with zero-frame** — agreed, and it defeated the point of the change: an empty WAV fell back to the wall clock and was saved at full length. Now returns `Double?`; only a decode failure falls back.

**SCK restart backoff** — the conclusion is right, the stated first reason isn't. There was never a fast burn through the budget: `didStopWithError` is the only trigger and a throwing bring-up creates no stream, so exactly one attempt could ever happen. That's the actual bug, and it's worse than a fast burn — a single transient failure lost app audio permanently while the budget sat unused. Fixed by retrying while the budget allows, with a growing backoff (0.5s→2.5s, ~7.5s total) so the attempts span a real outage rather than a single display reconfiguration.

One new test for the backoff schedule; 40 tests green locally.

urisland:
Good catch — taken in fce7a76, and the microphone path had the identical race, which is the one this PR is actually about.

Both recorders now carry a `captureEpoch` bumped by every `start()` and `stop()`. Recovery work captures the value it began with and abandons itself the moment it no longer matches:

- `SystemAudioRecorder.bringUpStream(epoch:)` checks at the **point of adoption** rather than only before the call. It suspends twice (`SCShareableContent`, `startCapture`), so a check before the call still leaves a window; on a stale epoch it now calls `stopCapture()` on the stream it just built and reports `false`, and the restart loop returns.
- `MicrophoneRecorder.rebuildEngine` checks the epoch alongside `isRunning` before adopting, and only clears `isRebuilding` when the rebuild still belongs to the current session — otherwise a stale rebuild finishing late could clear a *new* session's in-flight flag.

Two notes on the reasoning:

- My previous `guard wantsCapture else { await stop(); return }` was worse than just incomplete — in exactly this stop→start race it would have called `stop()` on the **new** session's stream. Replaced by the adoption-point check, which is the only place that can distinguish the two.
- `lastStreamError`: agreed it's already cleared on successful adoption, so nothing to change there.

On testing, so I'm not overclaiming: the adoption-point guards can't be unit tested without real CoreAudio/ScreenCaptureKit, so what's covered is the policy the loop consults (`restartDecision` with the epoch folded in as the intent flag) plus the existing watchdog wiring. 41 tests green locally.

PR #148, opened 2026-08-01
fix(audio): stop the config-change recovery loop from re-triggering itself

### LINKED ISSUE #147: Dictation records silence: the config-change recovery loop re-triggers itself, then the empty WAV surfaces as a server 500
## Symptom

Dictation captures no audio. The macOS microphone indicator flickers rapidly for the duration, and on stop the user gets `Failed to decode audio` / HTTP 500 from the transcription server — which makes it look like a server outage. The server is fine.

Reported on 1.9.2-beta.2 (build 56), macOS 27.0 (26A5388g), Mac17,8.

## This is one bug, not two

The 500 is a downstream symptom. Capture delivers **zero buffers**, so Mila writes a header-only WAV (~1.1 KB, no samples) and POSTs it; the server correctly refuses to decode it:

```
MicrophoneRecorder] mic stopping: buffers=0 frames=0 conversionFailures=0 rebuilds=13
MicrophoneRecorder] microphone delivered NO audio this session — the recording will be empty
RemoteWhisperEngine] transcribe: POST model=deepdml/faster-whisper-large-v3-turbo-ct2 lang=en bytes=1138
TranscriptionService] transcribeOnceSegments(remote): FAILED error=Remote server returned HTTP 500: {"detail":"Failed to decode audio."}
```

Three consecutive sessions, two languages, same result — `bytes=1138` (en) and `bytes=1132` (he). Model routing is irrelevant: both the English and Hebrew models return the same 500 for empty audio.

## Root cause: the recovery loop reacts to its own device selection

`MicrophoneRecorder.bringUp()` explicitly binds the engine to the chosen input via `AudioDeviceManager.setInputDevice` (`kAudioOutputUnitProperty_CurrentDevice`). From the diagnostic log, in order:

```
[MicrophoneRecorder] input device: MacBook Pro Microphone [Apple Inc.] uid=BuiltInMicrophoneDevice
[MicrophoneRecorder] native input format: 48000.000000Hz 1ch
[avae] AVAudioEngine.mm:1538  Engine@0x…: iounit configuration changed > posting notification
[MicrophoneRecorder] AVAudioEngine configuration changed mid-session … capture is dead until rebuilt
[avae] AVAudioEngine.mm:1337  Engine@0x…: stop, was running 1
[AUHAL] AUHAL.cpp:416   AUHAL: (0x…) Selecting device 126 from constructor
[AUHAL] AUHAL.cpp:3406  IsDeviceUsable: (0x…) Device ID: 126 (Input:No | Output:Yes): true
```

Device IDs seen on this machine: **121 = MacBook Pro Microphone (Input:Yes|Output:No)** — correctly selected; **126 = MacBook Pro Speakers (Input:No|Output:Yes)**; **182 = (Input:No|Output:No)**. After the rebuild the engine's own dump reads:

```
Input render format:  0 ch, 0 Hz, lpcm …
```

So: bind device → engine posts a configuration change → `handleConfigurationChange()` calls `rebuildEngine` immediately → bring-up binds the device again → another configuration change → repeat. Measured cadence ≈ **150 ms**, i.e. ~7 engine start/stop cycles per second; that is the flickering indicator. 11–14 rebuilds in the ~2 s before the user stopped.

The tap never survives long enough to deliver a buffer (4096 frames @ 48 kHz ≈ 85 ms, against a ~150 ms teardown cycle), hence `buffers=0`.

Two design gaps make this self-sustaining:

1. **No distinction between "the world changed" and "we changed the device."** The observer is registered around the same bring-up that then sets `kAudioOutputUnitProperty_CurrentDevice`, so a self-inflicted notification is indistinguishable from a yanked device.
2. **No minimum interval or backoff on the notification path.** `maxMidSessionRestarts = 60` carries the comment "60 attempts at ≥1s apart still rides out a device that's unplugged for a minute" — but that spacing is only true of the 1 s watchdog poll. `handleConfigurationChange()` rebuilds with no delay, so the cap permits 60 rebuilds in ~9 s rather than 60 s.

Regression window: this recovery path is new in #143 (`a982547`), so it is present in **1.9.2-beta.2** and absent from 1.9.2-beta.1.

Possible aggravators on this machine, not yet isolated: **BlackHole 2ch** and a **Microsoft Teams Audio** virtual device are installed, and CoreAudio's dynamic-aggregate (`ddagg`) code is active in the same window (`fetched default input device, ID = 121`). Whether the self-trigger reproduces on a Mac with no virtual audio devices is the first thing to establish.

## Suggested fix

- Suppress configuration-change handling that we caused ourselves — e.g. set the device *before* registering the observer, and/or ignore notifications arriving within a short grace window of a bring-up.
- Add a real minimum interval + backoff between rebuilds on the notification path, so the 60-attempt cap spans minutes rather than seconds.
- After a rebuild, verify the input actually works (`inputFormat.channelCount > 0` and buffers arriving) rather than assuming success; give up loudly if not.
- **Never upload empty audio.** If `frames == 0`, fail with "no audio was captured" instead of POSTing a header-only WAV and surfacing the server's decode error. A user cannot act on `HTTP 500: Failed to decode audio`.
- Consider skipping the re-bind when the device ID is already current.

## Repro / verification notes

The bug needs a real mic and a real audio stack, so it will not show up in the existing unit tests (`MicrophoneRecorderTests` drives `bringUpOverride`). A regression test can at least pin the *policy*: assert that the notification path enforces a minimum interval between rebuilds and that a zero-frame session refuses to reach the upload path.

urisland:
Closes #147

> **Release-blocking.** This needs to ship as build 57 / 1.9.2-beta.3: Sparkle
> won't downgrade, so anyone already on build 56 stays broken until a higher
> build number is published. 1.9.2-beta.2 is held in the appcast meanwhile.

## The mechanism

Confirmed by bisect on the reporter's machine — same Mac, same audio devices,
same macOS 27.0 (26A5388g), only the build changed:

* build 56 (has #143): `buffers=0 frames=0 rebuilds=13`, upload `bytes=1138`, HTTP 500.
* build 55 (predates #143): `buffers=14 frames=21626`, upload `bytes=31572`, transcription succeeds.

Build 55 runs the *same* `setInputDevice` path against the same devices and is
fine. So the configuration change was in all likelihood being posted before
#143 too — what #143 added is an observer that now **reacts** to it, and the
reaction re-triggers it. The fix is therefore built around not reacting to a
change we caused, not around preventing the change.

## Why it loops

`MicrophoneRecorder.bringUp()` binds the chosen input with
`AudioDeviceManager.setInputDevice` → `kAudioOutputUnitProperty_CurrentDevice`.
On the reporter's Mac that write is itself a reconfiguration of the I/O unit,
and AVAudioEngine answers it with `AVAudioEngineConfigurationChange`
(`iounit configuration changed > posting notification` in their log).

`handleConfigurationChange()` could not tell that notification apart from a
yanked device, so it rebuilt the engine immediately — and the rebuild bound the
device again, which posted again. Measured cadence ≈ 150 ms, i.e. ~7 microphone
start/stop cycles per second: the flickering indicator the user saw. A 4096-frame
buffer at 48 kHz needs ~85 ms to fill, against a ~150 ms teardown cycle, so the
tap never survived long enough to deliver one — `buffers=0 frames=0 rebuilds=13`.

`maxMidSessionRestarts = 60` didn't bound it in any useful way. Its comment
("60 attempts at ≥1s apart") describes the *watchdog's* cadence; the
notification path had no spacing at all, so the whole budget could burn in ~9 s.

Everything after that is downstream: zero frames → a header-only ~1.1 KB WAV →
POSTed → `HTTP 500: {"detail":"Failed to decode audio."}`, which reads to the
user as a transcription-server outage. The server was right; the audio was empty.

Regression window: this recovery path is new in #143 (`a982547`), so beta.2 has
it and beta.1 does not.

## What changed

**1. Don't react to our own bind** (the primary fix) — configuration changes
arriving within `configurationChangeGracePeriod` (0.75 s) of a bring-up are
logged and ignored. Build 55 proves the bind is survivable; reacting to it is
what isn't.

**2. Don't provoke the reconfiguration where it's free not to** —
`AudioDeviceManager.setInputDevice` now reads
`kAudioOutputUnitProperty_CurrentDevice` first and skips the write when the unit
is already on the device we want. Secondary and opportunistic, not the fix: it
removes one redundant reconfiguration in the common case (pinned input == system
default) and does nothing otherwise.

**3. Pace what's left** — `RebuildThrottle`, a pure clock-injected policy in the
shape of the existing `CaptureStallDetector`: 1 s floor, doubling to a 30 s
ceiling, with the ladder resetting after a quiet spell so an unrelated device
change 40 minutes into a meeting isn't punished for a storm at the start. The
60-rebuild budget now spans ~28 minutes on the notification path instead of ~9
seconds. The watchdog path is left alone — it is already spaced by
`captureStallTimeout` (4 s), which gives it a 4-minute floor for the same budget,
and throttling it too would only cost recovery latency.

**4. Never adopt a rebuild that didn't work** — the bring-up now requires
`channelCount > 0` as well as `sampleRate > 0`. In the reporter's log the
rebuilt engines selected the *output* device (126, `Input:No | Output:Yes`) and
reported `Input render format: 0 ch, 0 Hz`; those were adopted as working
engines. Now they throw `noInputDevice`, get logged as a failed rebuild, and the
watchdog retries.

**5. Never upload empty audio** — independent of the loop, and the change that
converts an unactionable HTTP 500 into a real diagnosis. `AudioSignal.isSilent`
(no samples, or nothing but exact zeros) gates two places:

* `TranscriptionService.transcribeOnceSegments` — the live/dictation path the
  reporter actually hit — sets `lastError` to "No audio was captured, so there
  was nothing to transcribe. Check System Settings ▸ Privacy & Security ▸
  Microphone…" and returns no segments.
* `RemoteWhisperEngine.transcribe` throws `.noAudioCaptured` before encoding
  anything, so the transport layer can't be made to POST an empty file by some
  future caller.

This guard stands on its own merits and would be worth landing even if the
recovery-loop fix were narrowed to nothing: it is what turns
`HTTP 500: Failed to decode audio.` into a sentence the next person can act on.

The predicate is deliberately the strictest one possible, *not* a loudness gate:
`minimumAudioPeak` already handles "too quiet to be speech" on the batch path,
and a 300 ms live utterance must not be discarded for being quiet. A live
microphone in a silent room delivers dither, never a run of exact zeros.

## Why this escaped CI, and what the new tests assert

`MicrophoneRecorderTests` drove only the `bringUpOverride` seam — and that seam
replaces precisely the real-engine code that self-triggers. The
configuration-change path had no coverage at all, and CI runners have no
microphone.

**The seam.** The recorder now registers its observer against
`configurationChangeSource`: the live `AVAudioEngine` in production, a stand-in
`ConfigurationChangeSource` object on the override path. Everything under test is
real — the observer, the notification name, the handler, the grace window, the
throttle, the budget, the epoch guards. Only the *poster* of the notification is
simulated, which is exactly the part CoreAudio would otherwise have to supply.
Tests re-read the source before each post, because production installs a new
engine per rebuild and it is the new engine that posts the next change; a test
that held the first source would silently stop exercising the loop.

| Test | Asserts |
| --- | --- |
| `RebuildThrottleTests.test_a_storm_of_changes_collapses_to_one_rebuild` | 60 changes at 150 ms → ≤4 rebuilds |
| `…test_without_a_minimum_interval_every_change_rebuilds` | documents the shipped behaviour: 60 changes → 60 rebuilds |
| `…test_backoff_doubles_and_saturates_at_the_ceiling` | 1, 2, 4, 8, 8, 8 |
| `…test_the_rebuild_budget_spans_minutes_not_seconds` | 60 rebuilds span >20 min, not ~9 s |
| `…test_a_quiet_spell_resets_the_backoff` | a device change 10 min later isn't stuck at the ceiling |
| `…test_pacing_exactly_at_the_ceiling_is_not_a_quiet_spell` | CI caught a real flaw here on the first push: a saturated burst arrives exactly `maximumInterval` apart, and the original reset rule read that as "quiet", collapsing the floor back to 1 s. A burst now ends only after `quietPeriod` (2× the ceiling). |
| `MicrophoneConfigurationChangeTests.test_self_inflicted_change_during_bring_up_does_not_rebuild` | a change at bring-up time produces **0** rebuilds |
| `…test_change_after_the_grace_window_still_rebuilds` | the counterweight — a genuine change is still repaired, exactly once |
| `…test_a_storm_of_configuration_changes_is_bounded` | dozens of notifications over a 0.6 s wall-clock window → ≤2 rebuilds (this is the bug's own shape) |
| `…test_configuration_change_after_stop_is_ignored` | no bring-up may happen after `stop()` |
| `TranscriptionServiceTests.test_zeroFrameCapture_neverReachesTheUploadPath` | zero samples → remote engine called **0** times, `lastError` is the "no audio was captured" message |
| `…test_digitalSilence_neverReachesTheUploadPath` | same for a muted/dead device |
| `…test_quietButRealAudio_isStillTranscribed` | one −60 dBFS sample in 0.2 s still goes through — the guard is not a loudness gate |
| `RemoteWhisperEngineParsingTests.test_refusesToUploadEmptyAudio` | endpoint points at a dead port; `.noAudioCaptured` (not a network error) proves it refused *before* the request |
| `AudioSignalTests` (4) | the predicate itself |

All of these run headless: no microphone, no CoreAudio device, no network, and no
real-timer sleeping except the deliberately wall-clock-bounded storm test.
The existing `MicrophoneRecorderTests` / `CaptureStallDetectorTests` /
`MicrophoneWatchdogTests` are untouched.

## Verified vs. inferred

**Verified by test:** the pacing policy and its arithmetic; that a self-inflicted
change inside the grace window produces no rebuild; that a real change outside it
still does; that a storm is bounded; that empty and all-zero audio never reach
the remote engine while quiet audio does.

**Inferred from the issue's logs, not reproduced here:** that the `setInputDevice`
write is what posts the configuration change on that Mac (the log ordering is
strong evidence — the bind is the only thing between "input device: …" and
"iounit configuration changed"); that the redundant-bind skip actually fires
there (it depends on the AUHAL already being on device 121, which the log's
`fetched default input device, ID = 121` supports but does not prove); and that
`channelCount > 0` is the right discriminator for the degenerate rebuilt engines.
None of that is testable without that machine's audio stack — I could not
reproduce the bug.

**The AUHAL-picks-the-output-device case, explicitly.** On that Mac the AUHAL
constructor selects device 126 (`Input:No | Output:Yes`) before Mila binds the
real mic (121). This PR tolerates that but does not cure it:

* the redundant-bind skip compares `current (126)` against `target (121)`, sees a
  mismatch, and binds exactly as before — no behaviour change, no regression;
* if the unit is *still* effectively on the output device afterwards, the
  bring-up now reads `0 ch, 0 Hz` and throws `noInputDevice` instead of adopting
  a mute engine. At session start that surfaces as a real error the user can
  see; mid-session it's logged as a failed rebuild and the watchdog retries.

So the failure mode becomes loud and bounded rather than silent, but Mila still
cannot stop the constructor from picking 126, and does not currently fall back to
a different input device when the bound one comes up unusable.

**Not addressed:** whether BlackHole 2ch / Microsoft Teams Audio / the `ddagg`
dynamic-aggregate activity are aggravators. If the bind still provokes a change
on that Mac despite fix (1) — e.g. because the user pins an input that is *not*
the system default — the residual behaviour is a rebuild every ~4 s from the
watchdog, capped at 60 and then logged as given up, instead of ~7 per second for
the whole recording. Bounded and diagnosable, but still not capturing; that case
needs the reporter's machine to settle.

## Risk

**Fix (2) can mask a real device change.** A device genuinely lost within 0.75 s
of a bring-up no longer takes the fast notification path. The stall watchdog
still catches it, so the cost is up to `captureStallTimeout` (4 s) of silence at
the very start of a recording rather than a lost recording — and
`test_change_after_the_grace_window_still_rebuilds` exists to keep the window
from quietly widening into "never react". Fix (3) has the same shape one level
up: a second genuine change within the backoff interval is deferred to the
watchdog rather than handled immediately.

**Fix (1)** changes device-binding behaviour for everyone, not just the reporter.
If reading `kAudioOutputUnitProperty_CurrentDevice` ever returned a stale or
misleading value we would skip a bind we needed; a failed read returns `nil` and
falls through to the write, so that failure mode is "behaves exactly as before".

## Summary by CodeRabbit

* **New Features**
  * Added detection for empty and silent audio recordings.
  * Added clear user-facing messaging when no audio is captured.
  * Improved microphone device tracking and configuration recovery.

* **Bug Fixes**
  * Prevented unnecessary audio device updates.
  * Reduced repeated microphone rebuilds during configuration changes.
  * Prevented silent recordings from being uploaded for transcription.
  * Preserved transcription for very quiet but valid audio.

### DIFF
--- a/Mila/Audio/AudioDeviceManager.swift
+++ b/Mila/Audio/AudioDeviceManager.swift
@@ -63,13 +63,40 @@ enum AudioDeviceManager {
         return inputs.first(where: { !$0.isVirtual })
     }
 
+    /// The device an engine's input unit is currently bound to, or nil if it
+    /// can't be read. A fresh `AVAudioEngine` comes up on the system default
+    /// input, so this is usually already the device we're about to select.
+    static func currentInputDeviceID(on engine: AVAudioEngine) -> AudioDeviceID? {
+        guard let unit = engine.inputNode.audioUnit else { return nil }
+        var deviceID: AudioDeviceID = 0
+        var size = UInt32(MemoryLayout<AudioDeviceID>.size)
+        let status = AudioUnitGetProperty(
+            unit,
+            kAudioOutputUnitProperty_CurrentDevice,
+            kAudioUnitScope_Global,
+            0,
+            &deviceID,
+            &size
+        )
+        return status == noErr ? deviceID : nil
+    }
+
     /// Force an existing `AVAudioEngine`'s input node to read from the given device.
     /// Must be called before `engine.start()` and *after* touching `engine.inputNode`.
+    ///
+    /// Setting `kAudioOutputUnitProperty_CurrentDevice` reconfigures the I/O
+    /// unit, and a reconfiguration can make the engine post
+    /// `AVAudioEngineConfigurationChange` — which the recorder's mid-session
+    /// recovery listens for. When the unit is *already* on the device we want
+    /// (the common case: the user's pinned input is also the system default),
+    /// that notification buys nothing and costs a rebuild, so skip the write
+    /// entirely. See issue #147.
     static func setInputDevice(_ device: Device, on engine: AVAudioEngine) throws {
         guard let unit = engine.inputNode.audioUnit else {
             throw NSError(domain: "AudioDeviceManager", code: 1,
                           userInfo: [NSLocalizedDescriptionKey: "Engine has no input audio unit."])
         }
+        if currentInputDeviceID(on: engine) == device.id { return }
         var deviceID = device.id
         let status = AudioUnitSetProperty(
             unit,
--- a/Mila/Audio/AudioUtilities.swift
+++ b/Mila/Audio/AudioUtilities.swift
@@ -172,6 +172,26 @@ final class StreamingWhisperConverter {
     }
 }
 
+/// Predicates about a raw sample buffer, shared by every path that has to
+/// decide "is there anything here at all?".
+enum AudioSignal {
+    /// True when a buffer carries no signal whatsoever: no samples, or nothing
+    /// but exact digital silence.
+    ///
+    /// Deliberately the strictest possible test. It is NOT the "too quiet to
+    /// be speech" gate — that's `TranscriptionService.minimumAudioPeak`, which
+    /// only the batch path applies, because a quiet-but-real utterance is
+    /// still worth transcribing. This one answers the narrower question "did
+    /// capture produce anything?", which is safe to apply even to a 300ms live
+    /// utterance: a live microphone, in a silent room, still delivers dither
+    /// and noise, never a run of exact zeros.
+    ///
+    /// O(1) for real audio — `contains` returns at the first non-zero sample.
+    static func isSilent(_ samples: [Float]) -> Bool {
+        !samples.contains { $0 != 0 }
+    }
+}
+
 /// Computes a 0...1 RMS level from an audio buffer for VU meters.
 enum AudioMeter {
     static func level(from buffer: AVAudioPCMBuffer) -> Float {
--- a/Mila/Audio/MicrophoneRecorder.swift
+++ b/Mila/Audio/MicrophoneRecorder.swift
@@ -585,7 +585,19 @@ final class MicrophoneRecorder: ObservableObject {
         if let override = bringUpOverride {
             // Test path: no CoreAudio, but still exercise the full decision
             // and the timeout wrapper.
-            try? await Self.withTimeout(seconds: bringUpTimeout) { try await override() }
+            //
+            // Success and failure are kept as distinct as they are on the real
+            // path below: a bring-up that threw installed no engine, so it must
+            // not re-arm the grace window or register a new notification
+            // source. Otherwise the seam would quietly misreport a *failed*
+            // mid-session rebuild as a successful one — and the grace window is
+            // exactly what a test of this area is measuring.
+            do {
+                try await Self.withTimeout(seconds: bringUpTimeout) { try await override() }
+            } catch {
+                micLog.error("mic rebuild #\(attempt, privacy: .public) (reason: \(reason, privacy: .public)) failed: \(error.localizedDescription, privacy: .public) — the watchdog will retry")
+                return
+            }
             guard captureEpoch == epoch else { return }
             lastBringUpAt = Date()
             // Mirror the real path: a rebuild installs a new engine, so the
--- a/Mila/Audio/MicrophoneRecorder.swift
+++ b/Mila/Audio/MicrophoneRecorder.swift
@@ -99,9 +99,16 @@ struct RebuildThrottle {
 
     private var lastAllowedAt: Date?
     /// Rebuilds allowed in the current burst. Reset once things go quiet for
-    /// `maximumInterval`, so an unrelated device change an hour into a meeting
+    /// `quietPeriod`, so an unrelated device change an hour into a meeting
     /// isn't punished for a storm at the start of it.
     private(set) var streak = 0
+
+    /// Idle time that ends a burst. Twice the ceiling rather than the ceiling
+    /// itself: a burst that has saturated the backoff arrives *exactly* one
+    /// `maximumInterval` apart, and treating that as "quiet" would reset the
+    /// ladder on every tick and cap the backoff at `minimumInterval` forever —
+    /// which is the bug this type exists to prevent, reintroduced one level up.
+    var quietPeriod: TimeInterval { maximumInterval * 2 }
 
     init(minimumInterval: TimeInterval, maximumInterval: TimeInterval) {
         self.minimumInterval = minimumInterval
@@ -123,8 +130,8 @@ struct RebuildThrottle {
         if let last = lastAllowedAt {
             let elapsed = now.timeIntervalSince(last)
             if elapsed < currentInterval { return false }
-            // Quiet for a full backoff ceiling: the previous burst is over.
-            if elapsed >= maximumInterval { streak = 0 }
+            // Quiet for long enough that the previous burst is over.
+            if elapsed >= quietPeriod { streak = 0 }
         }
         lastAllowedAt = now
         streak += 1
--- a/Mila/Audio/MicrophoneRecorder.swift
+++ b/Mila/Audio/MicrophoneRecorder.swift
@@ -70,6 +70,68 @@ struct CaptureStallDetector {
     }
 }
 
+/// Pure, clock-injected minimum-interval + exponential-backoff gate for
+/// *notification-driven* engine rebuilds.
+///
+/// `AVAudioEngineConfigurationChange` does not mean "something broke" — it
+/// means "the I/O unit was reconfigured", and binding the input device during
+/// our own bring-up is itself a reconfiguration. On at least one Mac that made
+/// the recovery path self-sustaining: bind → notification → rebuild → bind →
+/// notification, measured at ~150ms per cycle, which never let the tap live
+/// long enough to deliver a single buffer (issue #147). The whole
+/// `maxMidSessionRestarts` budget — documented as "60 attempts at ≥1s apart" —
+/// was spent in about nine seconds, and the macOS microphone indicator
+/// flickered for the entire recording.
+///
+/// The watchdog path is naturally spaced by `captureStallTimeout`; this gate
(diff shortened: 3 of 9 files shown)


urisland:
@coderabbitai review

urisland:
@coderabbitai full review

PR #150, opened 2026-08-01
ci: require every PR to be linked to a GitHub issue

### LINKED ISSUE #149: Require every PR to be linked to a GitHub issue (enforce in CI)
## Problem

`.claude/rules/pull-requests.md` already requires that a PR fixing a bug or
changing behaviour be linked to a GitHub issue with a closing keyword, and
explains why: the issue is the durable, searchable record of *what was wrong and
why*, while the PR is only *how it was fixed*. Without the link, that context
scatters into commit messages and is effectively lost for release notes,
regression hunts and "why did we change this?" archaeology.

In practice the rule is being skipped. It is a convention with nothing enforcing
it, and the existing carve-out ("trivial chores don't need an issue") is
self-applied, so it stretches to cover almost anything.

## Intended fix

Make the rule enforced rather than advisory:

1. A GitHub Actions workflow that runs on every PR and fails when no issue is
   linked. It should accept both linking mechanisms — a closing keyword
   (`Closes`/`Fixes`/`Resolves`) in the PR body, and GitHub's own linkage
   (`closingIssuesReferences`, what the sidebar "Development" link sets) — and
   post a single, self-updating comment explaining how to fix it rather than a
   new comment per push.
2. Widen the rule from "every fix PR" to **every** PR, with two narrow, visible
   escape hatches so a blanket requirement cannot wedge the repo:
   - dependabot PRs exempt automatically (they can never have an issue, and
     otherwise every dependency bump is permanently red);
   - a maintainer-applied `no-issue-needed` label for genuine trivia.
3. Surface the requirement in `.github/pull_request_template.md` so filling in
   `Closes #` is the default path, not something CI tells you off about
   afterwards.
4. Update `.claude/rules/pull-requests.md` so the documented rule and the
   enforced rule agree, including both escape hatches.

## Notes

Making the check **blocking** is a separate branch-protection change and should
be done by hand by the repo owner after the workflow has proven itself on a few
PRs.

urisland:
Closes #149

## What & why

`.claude/rules/pull-requests.md` already said every fix PR must link an issue,
and explained why. Nothing enforced it, so it was routinely skipped. This makes
it a CI check.

New workflow `.github/workflows/require-linked-issue.yml` (check name
**"Linked issue"**) runs on `pull_request`
(`opened`, `edited`, `reopened`, `synchronize`, plus `labeled`/`unlabeled`) and
fails a PR with no linked issue. It accepts both mechanisms:

- **`closingIssuesReferences` via GraphQL** — the primary, authoritative signal.
  This is what the sidebar **Development → Link an issue** sets, and it also
  reflects closing keywords GitHub itself parsed out of the body.
- **A closing-keyword regex over the PR body** as fallback — covers
  `Closes`/`Fixes`/`Resolves` (any case, all tenses), the `owner/repo#123` and
  full-URL forms, and any eventual-consistency lag right after an edit.

HTML comments are stripped before matching, so template placeholder text — and
an unfilled `Closes #` line — cannot satisfy the check. On failure it posts one
comment and **updates it in place** on later runs, so pushes don't spam; once
the link is added the comment flips to a resolved note.

Also: `Closes #` moved to the top of the PR template (it was previously only
mentioned inside an HTML comment, which by design no longer counts), and the
rule doc rewritten to match what CI actually enforces.

## :warning: Two deliberate softenings of "every PR" — veto either if you disagree

The rule as literally stated would wedge the repo, so I added two exemptions.
Both are documented in the rule doc; both are easy to remove.

1. **Dependabot PRs are exempt automatically.** They can never have an issue and
   we get them regularly — without this, every dependency bump is permanently
   red. Matched on the PR author *or* triggering actor being `dependabot[bot]`
   (author, not just actor, so a human editing a dependabot PR doesn't
   accidentally un-exempt it).
2. **A new `no-issue-needed` label bypasses the check**, for genuine trivia
   (typo, comment tweak). Already created via `gh label create`, so it exists.
   It's a deliberate, attributable action visible in the PR timeline — the point
   being that skipping the rule leaves a trace instead of being the silent
   default.

If you'd rather have neither, deleting the two guard blocks in the script is a
~10-line change.

## `pull_request` vs `pull_request_target`

I used **`pull_request`**, deliberately:

- The job reads only PR metadata over the API. It performs **no checkout** and
  executes nothing from the PR, so there is no reason to hand it a privileged
  token — `pull_request_target` would be handing out write credentials on a PR
  event for no benefit, and leaves that footgun in the file for whoever edits it
  next.
- `pull_request_target` always runs the workflow definition from the **base**
  branch, so this workflow could never have been tested by the PR that adds it —
  the check below would not exist.
- **Fork PRs are still checked.** The job runs and passes/fails correctly. The
  only degradation is that `GITHUB_TOKEN` is read-only for fork PRs, so the
  explanatory comment can't be posted; the script treats commenting as
  best-effort and always writes the same guidance to the failure annotation and
  the run summary, which are visible on every run.

## Permissions

Minimum needed, set at both workflow and job level:
`contents: read`, `issues: read`, `pull-requests: write` (the comment only).
No blanket write.

Exemptions are evaluated **inside** the script rather than as a job-level `if:`,
so exempt PRs report a real `success` conclusion instead of `skipped` — a
cleaner signal if you later make this a required check.

## How it was tested

CI-and-docs only; no Swift touched, no app build involved.

- `actionlint` clean.
- `actions/github-script` pinned to a full commit SHA with a version comment
  (`3a2844b…` # v9.0.0), matching the house style; SHA verified against the
  upstream repo.
- The embedded script was extracted and run against a mocked
  `github`/`context`/`core` in the same `AsyncFunction` wrapper github-script
  uses — **26 cases, all passing**. Including: a body closing keyword → pass; no reference
  → fail + comment; dependabot → pass; `no-issue-needed` → pass; sidebar linkage
  with empty body → pass; unfilled template → **fail**; filled template → pass;
  `Related to #123` / `See #123` / `prefixes #12` / `closest #12` → correctly
  fail; GraphQL erroring → falls back to the regex; read-only fork token → still
  fails with a warning instead of erroring. Comment upsert verified: 3 failing
  runs produce exactly 1 comment, then it's updated (not duplicated) on fix.
- And this PR is its own live test — it links #149 with a closing keyword at the
  top, so the check should be green below.

  I also ran a **live negative test** on this PR: temporarily removing the
  closing line and re-saving the description fired the `edited` trigger, the
  check went red and posted the explanatory comment; restoring the line fired it
  again and the check went green with the comment updated in place (not
  duplicated). Both transitions are in this PR's checks history.

  > Note, since it nearly bit me while writing this: a *prose example* pairing a
  > closing keyword with an issue number is parsed by GitHub as a **real** link —
  > it would have quietly added a second closing reference to this PR. Every
  > example in this description is therefore worded to avoid that adjacency. It
  > is a pre-existing GitHub behaviour, not something this check introduces.

## :point_right: One manual step for you

I deliberately did **not** make this a required status check — that's a
branch-protection change and yours to make. When you're happy with it:

**Settings → Rules → Rulesets** (or **Branches → branch protection rule** for
`main`) **→ Require status checks to pass → add `Linked issue`.**

Worth letting a few PRs go through first to confirm it behaves before making it
blocking.

### DIFF
--- a/.claude/rules/pull-requests.md
+++ b/.claude/rules/pull-requests.md
@@ -1,21 +1,56 @@
 # Pull Requests
 
-## Every fix PR must be linked to a GitHub issue
+## Every PR must be linked to a GitHub issue
 
-When opening a PR that fixes a bug or changes reported behavior, it MUST be
-connected to a GitHub issue:
+**This is enforced by CI** — the `Require linked issue` workflow
+(`.github/workflows/require-linked-issue.yml`, check name **"Linked issue"**)
+fails any PR that has no linked issue.
 
 1. **Create the issue first** if one doesn't already exist. Describe the
    problem, the root cause, and the intended fix — a repro helps.
-2. **Link the PR to the issue** in the PR description with a closing keyword —
-   `Closes #<n>` (or `Fixes #<n>`) — so merging the PR auto-closes the issue and
-   the two stay permanently cross-referenced.
+2. **Link the PR to the issue.** Either mechanism satisfies the check:
+   - a closing keyword in the PR description — `Closes #<n>` (`Fixes` and
+     `Resolves` work too, any case; so do the `owner/repo#<n>` and full-URL
+     forms). `.github/pull_request_template.md` puts a `Closes #` line at the
+     top so this is the default path;
+   - GitHub's own linkage, via the sidebar **Development → Link an issue**.
+
+A closing keyword is preferred — it also auto-closes the issue on merge, so the
+two stay permanently cross-referenced.
 
 **Why:** the issue is the durable, searchable record of *what was wrong and
 why*; the PR is *how it was fixed*. Linking them keeps that history navigable
 (release notes, regression hunts, "why did we change this?" archaeology)
 instead of scattering the context across commit messages that are easy to lose.
 
-This applies to bug fixes and behavior changes. Trivial chores (typo fix,
-comment tweak, dependency bump) don't need an issue, but anything a user could
-file a bug about does.
+### Scope and escape hatches
+
+The requirement applies to **every** PR, not just bug fixes. This replaces the
+older "trivial chores don't need an issue" carve-out, which in practice was
+self-applied to almost everything. Two exemptions exist instead, both narrow
+and both visible:
+
+- **Dependabot PRs are exempt automatically.** They can never have an issue,
+  and the repo receives them regularly; without this every dependency bump
+  would be permanently red. Detected by the PR author (or triggering actor)
+  being `dependabot[bot]`.
+- **The `no-issue-needed` label bypasses the check.** For genuine trivia — a
+  typo, a comment tweak. It is a deliberate, attributable maintainer action
+  that shows up in the PR timeline, which is the point: skipping the rule
+  should leave a trace rather than being the silent default. Adding or removing
+  the label re-runs the check immediately.
+
+Anything a user could file a bug about does not qualify for the label.
+
+### Notes for maintainers
+
+- The check runs on `pull_request`, not `pull_request_target`, so it never runs
+  privileged against untrusted fork code and it checks nothing out. Fork PRs
+  are still checked; the only degradation is that the read-only fork token
+  cannot post the explanatory comment, so the guidance goes to the failure
+  annotation and the run summary instead.
+- The workflow posts a single comment and updates it in place — pushes do not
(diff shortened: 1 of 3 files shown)


urisland:
@coderabbitai review

urisland:
@coderabbitai review

urisland:
@coderabbitai review

urisland:
@coderabbitai review

PR #163, opened 2026-08-02
fix(ci): stop the review gate from invoking CodeRabbit on every failure

urisland:
Closes #162

The gate merged in #160 posts a failure comment whose "how to clear this check" section contains a **live** `@coderabbitai review` command inside a code block. CodeRabbit parses commands out of comment bodies regardless of fencing, so **the gate invoked a review every time it failed**.

Confirmed by ordering on #129 — the 17:31:32Z reply is addressed to `@github-actions[bot]`, nine seconds after the gate posted:

```
17:31:23Z github-actions   <!-- require-coderabbit-review --> failure comment
17:31:32Z coderabbitai     "@github-actions[bot] Requesting a review of the
                            current PR head, including commit 08a384b."
```

**Why this is worth a same-day fix:** CodeRabbit is on a per-user quota that has been the binding constraint on this repo all day. The gate fails on exactly the PRs that are waiting for a review slot, and every failure spends one. It is a feedback loop aimed at the scarcest resource in the pipeline.

**The fix** renders the command with a zero-width space (`@coderabbit<U+200B>ai review`) so it still reads correctly to a human, plus an explicit note that it must be retyped by hand. No logic change — the detection signals, the bypass label and the triggers are untouched.

## Still open, deliberately not fixed here

#129 currently **cannot** clear this gate. CodeRabbit answered `"Action performed. Review finished."` for head `08a384b` but left no artifact — no review object, no inline comments, no `Review complete` verdict — and its only substantive review is of `080931f`, five commits back. Since it "does not re-review already reviewed commits", the gate is asking for proof it will not produce.

I did not paper over that by trusting `"Review finished"` as a fourth signal: that string is an acknowledgement of the command, not evidence a review happened, and accepting it would reopen precisely the false-positive #160 exists to close. It needs a deliberate decision — documented in #162.

## Summary by CodeRabbit

* **Documentation**
  * Updated review failure guidance with clearer instructions for requesting an automated review.
  * Clarified that the command should be retyped manually to avoid accidental invocation from explanatory text.

PR #167, opened 2026-08-02
ci: bump the actions group across 1 directory with 6 updates (relay of #141)

### LINKED ISSUE #166: Keep GitHub Actions pinned to current SHAs (August 2026 group bump)
Dependabot opened #141 to bump six pinned actions. It cannot merge: the `CodeRabbit reviewed head` gate applies to every PR with no dependabot exemption (deliberately — the owner asked that nothing skip review), and CodeRabbit bills reviews to the **PR author's** per-user quota, which a bot does not meaningfully have. So the bump needs re-opening under a human account.

## The bumps

| Action | From | To |
|---|---|---|
| `docker/login-action` | 4.4.0 | 4.5.1 |
| `github/codeql-action/{init,analyze,upload-sarif}` | 4.37.1 | 4.37.3 |
| `actions/setup-python` | 6.3.0 | 7.0.0 |
| `ossf/scorecard-action` | 2.4.3 | 2.4.4 |

`actions/setup-python` is a **major** bump (v6 → v7) and deserves a look rather than a rubber stamp.

## Why it matters beyond hygiene

All 16 third-party `uses:` in this repo are pinned to full 40-char SHAs with version comments — that pinning is what makes a supply-chain compromise of an upstream tag a non-event here. Keeping the SHAs current is how that protection stays real rather than becoming a frozen snapshot of July.

urisland:
Closes #166
Supersedes #141 — dependabot's commit is cherry-picked unchanged, so the diff is byte-identical.

## Why this exists instead of #141

The `CodeRabbit reviewed head` gate has **no dependabot exemption** — a deliberate choice, since the ask was that nothing skips review. But CodeRabbit bills each review to the **PR author's** per-user quota, and a bot has none, so #141 could never satisfy the gate it is subject to. Re-opening under a human account is the way out that keeps the gate intact rather than punching a hole in it.

## The diff

Six SHA-pinned action bumps across 4 workflow files, +8/-8:

| Action | From | To |
|---|---|---|
| `docker/login-action` | 4.4.0 | 4.5.1 |
| `github/codeql-action/{init,analyze,upload-sarif}` | 4.37.1 | 4.37.3 |
| `actions/setup-python` | 6.3.0 | **7.0.0** (major) |
| `ossf/scorecard-action` | 2.4.3 | 2.4.4 |

Every entry stays a full 40-char SHA with a `# vX.Y.Z` comment, matching the house convention across all 16 third-party `uses:` in this repo.

## What I checked

- The diff is exactly dependabot's, cherry-picked — no hand edits.
- `setup-python` v6 → v7 is the only major bump. It is used by the Live AI E2E mock jobs and the speaches image build; those jobs passed on #141 before it went stale, and they run again here.
- Whether the org wants dependabot exempted from the CodeRabbit gate going forward is a separate decision worth making deliberately — otherwise every future bump needs this same manual relay. Flagging rather than deciding it here.

### DIFF
--- a/.github/workflows/build-speaches-image.yml
+++ b/.github/workflows/build-speaches-image.yml
@@ -62,7 +62,7 @@ jobs:
       # don't push — PRs from forks can't get a write token anyway.
       - name: Log in to GHCR
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # v4.4.0
+        uses: docker/login-action@abd2ef45e78c5afb21d64d4ca52ee8550d9572c7 # v4.5.1
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -45,7 +45,7 @@ jobs:
       # Initialize BEFORE the build so CodeQL's tracer observes Swift
       # compilation (manual build mode — autobuild can't drive XcodeGen).
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@7188fc363630916deb702c7fdcf4e481b751f97a # v4.37.1
+        uses: github/codeql-action/init@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
         with:
           languages: swift
           build-mode: manual
@@ -74,6 +74,6 @@ jobs:
             | cat
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@7188fc363630916deb702c7fdcf4e481b751f97a # v4.37.1
+        uses: github/codeql-action/analyze@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
         with:
           category: "/language:swift"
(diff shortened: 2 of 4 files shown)


urisland:
Applying \`coderabbit-not-required\` to merge — recording why, because this is the label's first use and it must not read as skipping review.

**CodeRabbit did review this PR at the current head.** Its walkthrough comment names the exact SHA:

- #163: `between c2d19fc… and 016d9177e385d617c2a25ab6bb01ba1544a855e7`
- #167: `between c2d19fc… and 6b94629ca306a8bc1ec068c2b3d812f182a781c6`

with **no actionable comments generated**. The `CodeRabbit reviewed head` check is nonetheless red, because it only recognises three signals — a review object with a body, inline comments matching `original_commit_id`, and a `Review complete … <sha>` comment — and **a clean review produces none of those**. It posts a walkthrough issue comment and creates no review object at all.

So the bypass is being used against a **false negative in the checker**, not to dodge a review that did not happen. That is a different thing from bypassing because of a quota, and I would not have used the label for the latter without asking.

The detection gap is being fixed in a separate PR (adding the walkthrough signal, verified against #163, #124, #119 and #115 as live test cases). This PR is part of that same repair: it stops the gate posting a live `@coderabbitai review` in its own failure comment, which made every failure spend one of the rate-limited review slots the gate is waiting for.

Issue #131, opened 2026-07-30
Settings → Updates tab looks disabled and shows a stray overflow chevron; fold it into General


urisland:
## Problem

Settings → **Updates** looks broken: the tab item renders greyed-out/inactive and there is an "odd expand button" (a `>>` chevron) in the Settings tab strip.

## Cause

Not a bug in the tab's own content — it's tab-strip **overflow**. `SettingsView` declares its `TabView` with a fixed `.frame(width: 560, height: 560)` while carrying 10 `tabItem`s (Hotkeys, Audio, Models, LLM, Speakers, Meetings, Live AI, Voice Memos, Storage, Updates).

Measuring the labels at the system tab font (13pt) plus icon + item padding gives roughly **845 pt** of tab items in a **560 pt** wide strip. AppKit's `NSTabView` handles that by collapsing the trailing items into an overflow (`>>`) menu, which is both the "expand button" and why the last tab reads as disabled.

This is a **general** property of the tab list, not something specific to Updates: at 9 tabs the strip is still ~758 pt, so the chevron does not go away just by removing one tab. Any tab added at the end will land in the overflow.

## Fix in this pass

The Updates tab held exactly one toggle and a version string — not enough to justify a tenth tab. Fold it into a small **Updates** group inside a **General** tab (the former Hotkeys tab, renamed), with:

- "Automatically check for updates" bound to Sparkle's own `SPUUpdater.automaticallyChecksForUpdates` (no second updater, no parallel user default), and
- the beta opt-in checkbox nested directly underneath it, keeping the explanatory caption and the `updates.beta.toggle` accessibility identifier, and
- the running version string.

## Not fixed here (follow-up)

The `560 pt` frame width vs. the tab strip is left alone — widening the Settings window changes the layout of every tab and wants visual verification. Tracking that separately.

PR #183, opened 2026-08-09
feat(mcp): mila-mcp — an opt-in MCP stdio server over Mila's transcriptions

### LINKED ISSUE #182: Expose Mila's transcriptions to local MCP clients (opt-in)
## Problem

Mila's transcriptions are only reachable through Mila. To ask an assistant
"what did we decide in Tuesday's standup?", you have to open the app, find
the recording, copy the transcript out, and paste it somewhere — for data
that is already sitting in a JSON store on the same machine as the
assistant.

## Proposal

Ship an MCP stdio server, `mila-mcp`, embedded in the app bundle, exposing
four read-only tools to an MCP client on the same Mac (Claude Code, Claude
Desktop):

- `list_recordings` — filter by speaker, folder, source, date range
- `get_transcript` — one recording, speaker names resolved, plus summary
  and action items
- `search_transcripts` — full-text across titles and transcripts
- `get_live_transcript` — poll a meeting **while it is still happening**

## Conditions

1. **Off by default.** No transcript is reachable until the user
   explicitly opts in.
2. **Local only.** No network listener, no port.

## Architecture

Two shapes were considered:

- **A** — stdio helper reading Mila's files directly.
- **B** — Mila serves a local API over a unix domain socket; `mila-mcp`
  becomes a thin client, decoupling it from the store's on-disk format.

**A is chosen**, for two reasons:

- *Availability.* Mila is an ordinary dock app (`LSUIElement: false`, no
  login item). The question this feature exists to answer arrives days
  after the meeting, with Mila closed. Files answer it either way; a
  socket only answers while the app runs.
- *The socket buys no isolation.* A socket at `0600` is reachable by
  exactly the same principals as a `0600` file — the user's own uid.

The decision must stay reversible: a `MilaDataSource` seam so that a
socket-backed implementation is a swap, not a rewrite, if a third-party
consumer ever appears or the store format starts churning.

## Consent gate

The opt-in should be honest about what it is. `mila-mcp` runs as the user
and could read `~/Library/Application Support/Mila/` regardless, so the
gate is **not** a privilege boundary and must not be described as one. It
prevents the failure that actually happens: an MCP client that is
configured and running quietly reading meeting transcripts and sending
them to a cloud model because the tools were simply there.

Requirements: fails closed on a missing/unreadable/malformed flag;
re-read per tool call so revoking access affects an already-running
server; the refusal names the setting that unblocks it.

## Prior art

Originally proposed in #128 by @ValeroK, building on @DanielSolomon's
work. That PR was parked pending this architecture decision and has since
gone stale against `main`.

urisland:
Closes #182

Relays #128 (@ValeroK, building on @DanielSolomon's work) onto current
`main` and finishes it under the architecture settled in review. The five
original commits are cherry-picked with their authorship intact; the sixth
is the new work.

## The architecture decision

#128 was parked on one question: should `mila-mcp` read Mila's files, or
should Mila serve a local API with the helper as a thin client? **Files**,
for two reasons.

*Availability.* Mila is an ordinary dock app — `LSUIElement: false`, no
login item. The question this feature exists to answer ("what did we decide
on Tuesday?") arrives days later, with Mila closed. Files answer it either
way; a socket only answers while the app is running.

*The socket buys no isolation.* A unix socket at `0600` is reachable by
exactly the same principals as a `0600` file: the user's own uid. The
security advantage it appears to have over reading files is not real.

`MilaDataSource` keeps the decision reversible. `FileBackedDataSource` is
the only implementation today; a socket-backed one would be the whole
change on this side, with the tool handler bodies untouched.

## Consent gate — off by default

`mcp-access.json` carries the opt-in. `MCPAccessSettings` writes it when
the user flips **Settings ▸ Storage ▸ "Allow MCP access to
transcriptions"** and re-mirrors it every launch, so a deleted file or an
upgrade from a build without one converges rather than drifting.
`MCPAccessGate` fails closed on missing, unreadable, or malformed input.

It is re-read on **every** tool call rather than cached at startup: an MCP
server an editor launched hours ago outlives many Mila sessions, and
revoking access should bite immediately. The refusal names the toggle,
because whoever reads it is inside an MCP client with no idea which macOS
app is refusing them.

The toggle sits beside the Obsidian card in Storage rather than in a tab of
its own — the tab strip's width is load-bearing and a tenth tab risks the
`>>` overflow chevron (#131/#137).

**On what the gate is.** It is a consent control, not a privilege boundary,
and it is described that way in the UI, `docs/mcp.md`, and `CLAUDE.md`.
`mila-mcp` runs as the user; so does everything else the user runs, and any
of it can read `~/Library/Application Support/Mila/` directly. What the
toggle prevents is the accident that actually happens — a configured MCP
client quietly reading meeting transcripts and shipping them to a cloud
model because the tools were simply there. Claiming it contains a hostile
process would be false, so nothing here claims it.

## Review findings from #128

**Fixed — sort comparators violated strict weak ordering.**
`order == .asc ? ascending : !ascending` reports both `a < b` and `b < a`
for equal elements. `sort(by:)` documents a strict weak ordering as a
requirement and gives unspecified results otherwise; in practice flipping
the direction silently reversed tied results. Replaced with a three-way
comparison in both `listRecordings` and `searchTranscripts`. The two new
regression tests were verified against the old comparator — they fail with
it (`["first","second","third"]` vs `["third","second","first"]`) and pass
with the fix.

**Not reproduced — the reported out-of-bounds slice in
`segments(sinceIndex:)`.** `start = max(0, min(index - 1, segments.count))`
clamps to `endIndex`, and `segments[endIndex...]` is a valid empty slice in
Swift, so a too-large `since_segment_index` returns `[]` rather than
trapping. Confirmed for `index` of 0, 1, 6 and 100 against a 5-element
array. Left unchanged.

**Corrected — the `mila-mcp` target comment.** It credited
ServiceLifecycle, which isn't linked at all (zero symbols). The size cost
is swift-nio: ~33k symbols in an 18 MB Debug binary, per `nm`. Note this is
the Debug number; Release dead-strips.

## Verification

- `make build` green.
- MilaKit package tests: **42 passing**, up from 33. New: six gate tests
  (default-denied, both directions, malformed-fails-closed, file location,
  every tool refused while off, revocation without restart) and two
  tie-ordering regression tests.
- The existing handler tests now open the gate through the real file in
  `setUp` rather than a stub, so the granted path exercises
  `MCPAccessGate.set` / `isEnabled` end to end.
- App-target tests (`MilaTests`) and the UI tests were left to CI — they
  include XCUITests, which are unreliable when run on a machine someone is
  using.

## Not in scope

The store model is still mirrored: MilaKit's `StoredRecording` duplicates
the app's `Recording`, with `StoredRecordingDriftTests` as the tripwire.
Collapsing them into one source of truth would delete that tripwire, but
`Recording`'s initialiser is referenced across dozens of files and the
refactor doesn't belong in a PR this size. The decoder is deliberately
lenient, so the mirror degrades gracefully meanwhile.

Co-authored-by: Daniel Solomon <DanielSolomon94.ds@gmail.com>
Co-authored-by: KobiValero-Pagaya <kobi.valero@pagaya.com>

## Summary by CodeRabbit

* **New Features**
  * Added optional, consent-controlled MCP access for desktop assistants.
  * Browse recordings, retrieve transcripts and summaries, search and filter conversations, and monitor live transcripts with speaker and session updates.
  * Added storage settings for enabling access and copying setup instructions.
  * Added the embedded `mila-mcp` helper and meeting-assistance workflow.
  * Added reliable live transcript availability during and after recordings.

* **Documentation**
  * Added setup guidance, available tools, live polling behavior, and meeting workflows.

* **Bug Fixes**
  * Improved store discovery, legacy data compatibility, and live transcript completion reliability.
  * Added fail-closed safeguards when access or storage cannot be verified.


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

PR #187, opened 2026-08-20
fix(llm): share one Claude Code project across all LLM invocations

### LINKED ISSUE #181: Every LLM invocation creates a new ~/.claude/projects directory - 175 orphans, 46 MB; all Mila sessions should share one project
**What happened**

Every LLM invocation Mila makes creates a **brand-new Claude Code project directory** under `~/.claude/projects/`, and none of them are ever cleaned up. After a few weeks of normal use my machine has **175 orphaned directories totalling 46 MB**:

~/.claude/projects filled with orphaned island-mila-llm directories

```
$ ls -d ~/.claude/projects/*island-mila-llm-* | wc -l
     175
$ du -shc ~/.claude/projects/*island-mila-llm-*/ | tail -1
  46M	total
```

Every one of them is a single-session directory named after a dead temp path:

```
~/.claude/projects/-private-var-folders-7g-2rnd36v54zz-…-T-island-mila-llm-<UUID>/<session-uuid>.jsonl
```

**Root cause**

`LLMRunner.executeProcess` chdirs the child into a fresh sandbox per invocation (`Mila/Actions/LLMRunner.swift:374-383`):

- one-shot calls (title generation, post-recording action, Settings → LLM test) → `makeSandboxDirectory()`, i.e. `$TMPDIR/island-mila-llm-<UUID>/` (`LLMRunner.swift:518`)
- Live AI ticks → `stableSandboxDirectory(key:)`, i.e. `$TMPDIR/island-mila-llm-session-<UUID>/` (`LLMRunner.swift:531`)

Claude Code keys its session store on a slug of the CWD, so **a new CWD per run means a new project directory per run**. The sandbox itself is torn down in the `defer` at `LLMRunner.swift:399` - but the `~/.claude/projects/…` entry it caused is not, and Mila has no knowledge of it. The result is write-only garbage that grows forever.

#112 fixed the Live-AI-side `$TMPDIR` leak and noted in passing that "`~/.claude/projects/` likewise accumulates one project directory per recording" - that half was never addressed, and it applies to *every* code path, not just Live AI.

**Steps to reproduce**

1. Configure Settings → LLM with `claude` and enable auto-naming (or just hit the Settings → LLM test button).
2. Record and transcribe a few recordings so title generation / the post-recording action fire.
3. `ls -d ~/.claude/projects/*island-mila-llm-*` - one directory per invocation, none removed.

**Expected behavior**

**All Mila-originated sessions should live under a single, stable Claude Code project** - one `~/.claude/projects/…` entry named recognisably for Mila, holding one `<session-uuid>.jsonl` per run.

That is how Claude Code is designed to work: a project directory already holds many sessions, keyed by session UUID inside it. The per-invocation split buys nothing and costs a lot.

**Why this matters beyond disk usage**

- **It pollutes the user's own `claude` install.** These directories show up in Claude Code's project/resume surfaces (screenshot above - 175 Mila entries burying the user's real projects, e.g. `-Users-leetwito-PycharmProjects-claims-qa` at the bottom).
- **The sessions are effectively unresumable.** Because the sandbox is deleted, `claude --resume <uuid>` can't find the conversation unless the user manually recreates the exact `/private/var/folders/…/T/island-mila-llm-<UUID>/` path first. A stable directory makes "go look at what the LLM actually did for that meeting" a one-liner - which matters for debugging bad titles and bad summaries.
- **Unbounded growth.** 46 MB after a few weeks, with no cleanup path anywhere in the app or in `claude` itself.

**Suggested fix**

Replace both sandbox factories with one stable directory that Mila owns, e.g. `~/Library/Application Support/Mila/llm-sandbox/` (or a fixed `$TMPDIR/island-mila-llm/`), and pass it as the CWD for every invocation.

This preserves the two properties the current design exists for:

- **TCC isolation** (the reason for the sandbox per the comment at `LLMRunner.swift:356-363`) - it is still an empty directory Mila owns, so `claude`/`cursor-agent` scanning the CWD finds nothing and no "Mila would like to access Desktop" prompts fire.
- **Live AI `--resume` continuity** (the reason `stableSandboxDirectory` exists) - `--resume <uuid>` resolves the session *within* the project directory, so a single shared CWD keeps resume working. This should let `stableSandboxDirectory(key:)` / `cleanupStableSandbox(key:)` be deleted outright, along with the cleanup-on-cancel obligation from #112.

An Application Support location is preferable to `$TMPDIR` because the slug stays stable across reboots (macOS rotates the per-user temp dir), so the project directory doesn't fragment again over time.

Worth verifying before shipping:

- **Concurrent runs sharing a CWD.** Live AI ticks can overlap a title-generation run. Sessions are separate `jsonl` files so they shouldn't collide, but any per-project state `claude` keeps (local settings, lock files) should be checked under concurrency.
- **Migration.** A one-off cleanup of existing `~/.claude/projects/*island-mila-llm-*` directories would be a nice touch, though users can do it themselves - worth at least a release-note line so people know they can delete them.

**Environment**
- macOS version: 26.5.1
- Mila version: 1.9.2 (58)
- Mac model: Apple Silicon (arm64)
- Feature involved: LLM integration - title generation, post-recording action, Live AI, Settings → LLM test (all paths)

urisland:
Closes #181

## What & why

Every LLM invocation minted a new `~/.claude/projects/` directory. The reporter had 175 of them / 46 MB; this machine had 91 / 11 MB.

### Confirmed root cause

`claude` derives its session store from the **process working directory** — it writes `~/.claude/projects/<slug-of-cwd>/<session-uuid>.jsonl`, where the slug is the cwd path with `/` and `.` replaced by `-`. `LLMRunner.executeProcess` gave every invocation its own cwd:

- one-shot calls (title generation, post-recording action, Settings → AI Provider test) → `makeSandboxDirectory()` → `$TMPDIR/island-mila-llm-<UUID>`
- Live AI ticks → `stableSandboxDirectory(key:)` → `$TMPDIR/island-mila-llm-session-<UUID>`

So: new cwd per run ⇒ new project directory per run. Evidence on this machine (91 orphans, of which 28 are Live-AI session dirs):

```
$ ls -d ~/.claude/projects/*island-mila-llm-* | wc -l
      91
$ du -shc ~/.claude/projects/*island-mila-llm-*/ | tail -1
  11M	total
$ getconf DARWIN_USER_TEMP_DIR
/var/folders/xh/lmcmd9sd57bf7zzyjb3pp9w40000gn/T/
$ ls ~/.claude/projects/ | grep island-mila-llm | head -1
-private-var-folders-xh-lmcmd9sd57bf7zzyjb3pp9w40000gn-T-island-mila-llm-03E62ADD-…
```

The directory name is exactly the slug of `$TMPDIR/island-mila-llm-<UUID>` — the `$TMPDIR` above, character for character. The same slug rule holds for the user's real projects (`/Users/uri/ai/work/Mila/.claude/worktrees/x` → `-Users-uri-ai-work-Mila--claude-worktrees-x`), so the mechanism is not in doubt. Each orphan contains a single `<session-uuid>.jsonl` — one run's conversation — and the temp path it is named after was deleted by the `defer` in `executeProcess`, which is why `claude --resume` could never find those conversations again.

This matches the issue's stated root cause, including the file/line references.

### The fix

One stable directory Mila owns — `~/Library/Application Support/Mila/llm-sandbox/` — used as the cwd for **every** invocation (`run`, session and non-session, plus `diagnose`). Both properties the per-run sandbox existed for survive:

- **TCC isolation.** It is still an empty directory Mila owns, so a CLI that scans its cwd for project context finds nothing and no "Mila would like to access Desktop" prompts fire.
- **`--resume` continuity.** Session identity comes from the UUID in `--session-id` / `--resume`, never from the cwd, so Live AI's resume still resolves — inside the shared project now. Side benefit the issue asked for: the conversations are finally resumable by hand (`cd "~/Library/Application Support/Mila/llm-sandbox" && claude --resume <uuid>`), which is the practical way to debug a bad title or summary.

Application Support rather than a fixed `$TMPDIR/island-mila-llm/` because the slug must stay stable across reboots: macOS both rotates the per-user temp directory and periodically purges it, either of which would fragment the project directory again.

`stableSandboxDirectory(key:)` / `cleanupStableSandbox(key:)` are deleted, and with them the cleanup-on-cancel obligation #112 put on `LiveAISession` (two call sites) — there is nothing per-run left to tear down. The now-stale "second reason" in `QuickActionsController.stopRecording`'s comment is updated; the first reason (a `pendingKickTask` waking up after the meeting) still stands, so the `cancel()` call stays.

On the issue's "worth verifying" list: **concurrent runs sharing a cwd** — a Live AI tick can overlap a title-generation run. They write different `<uuid>.jsonl` files, `createDirectory(withIntermediateDirectories: true)` is idempotent, and nothing in Mila writes into the cwd itself, so there is no new contention on Mila's side; any locking inside `claude`'s own per-project state is unchanged from the single-project case it is designed for (a human running two `claude` sessions in one repo).

### Why the 175 existing orphans are *not* cleaned up here

Deliberate, and I'd argue this is the right split:

1. **It is another tool's data store, and it holds conversations.** `~/.claude/projects/` belongs to Claude Code, not Mila. Each orphan is a real transcript of what the LLM did for one recording. The issue itself argues that being able to go read that back matters for debugging bad titles and summaries — so blanket-deleting the backlog works against one of the issue's own stated goals.
2. **The leak stops dead with this PR.** What remains is a bounded, static, one-time ~46 MB. It is not growing any more, so there is no urgency that justifies bundling data deletion into a leak fix.
3. **Doing it *well* needs its own design + UI review.** Silent deletion at launch is the worst option. The defensible shape is opt-in and visible (Settings → AI Provider: "N old LLM session folders, X MB — Remove"), with a strict name filter so nothing outside `*island-mila-llm-*` can ever match. That is a UI change needing human verification, and it does not belong in the review surface of a three-file leak fix.

Users who want the space back today can have it with the exact, narrowly-scoped one-liner (only Mila-generated names can match; no other project can):

```sh
rm -rf ~/.claude/projects/*island-mila-llm-*
```

Worth a line in the release notes for whichever version ships this — I have not added a `RELEASE_NOTES/` file because the version is not bumped here and post-1.9.2 PRs on `main` (e.g. #171) don't carry one either.

I have not opened a follow-up issue for the opt-in cleanup UI — that was not asked for and I would rather a maintainer decide whether it is wanted at all than leave an unrequested tracking issue behind. Happy to file it on request.

## How it was tested

- `make build` → **BUILD SUCCEEDED** (fresh worktree, with the local `PythonRuntime` symlink; removed again before committing).
- `xcodebuild test -only-testing:MilaTests/LLMSandboxDirectoryTests` → **Executed 6 tests, 0 failures**. All six new tests below ran and passed locally.

**Tests written but NOT executed locally**, and why: partway through this work I was told to stop running app-hosted tests on this machine — `MilaTests` is hosted in the app, every run launches a freshly ad-hoc-signed `Mila.app`, and test files touching `KeychainHelper` make macOS pop a keychain-authorisation dialog the user has to dismiss by hand. So the following did **not** run here and are relying on CI's `unit-and-ui-tests`:

- `MilaTests/LLMRunnerTests/test_runner_spawns_child_in_isolated_empty_directory` — the existing TCC-popup regression guard, renamed (it no longer asserts "under $TMPDIR") and now asserting the cwd *is* `LLMRunner.sandboxDirectory()`. Its empty-directory assertion was also switched from `ls -A` to `ls`: the sandbox is shared and persistent now, so an LLM CLI is free to leave its own dot-files there, while "the child sees no *user* content" — the actual TCC-relevant invariant — is what a visible-entry count proves, and it still fails loudly if the cwd ever became `$HOME`.
- `MilaTests/LiveAISessionTests`, `MilaTests/LiveAIThrottleTests` — not modified, but they cover the `LiveAISession.cancel()` / context-restart paths this PR edits. I started this run and it was killed before producing results.

It compiles: the `LLMSandboxDirectoryTests` run above builds the whole `MilaTests` target, including the modified `LLMRunnerTests.swift`.

New tests (`MilaTests/LLMSandboxDirectoryTests.swift`, kept out of `LLMRunnerTests` so it can be run alone — that class also holds smoke tests that invoke the real `claude` / `cursor-agent` / `gemini` binaries):

| test | pins |
| --- | --- |
| `test_sandbox_path_has_no_per_run_component` | path composition is pure and carries no UUID/PID/timestamp |
| `test_sandboxDirectory_is_stable_and_exists` | stable across calls, created on demand, under `Mila/llm-sandbox`, **not** under `$TMPDIR` |
| `test_two_oneshot_runs_share_one_working_directory` | the leak itself: two one-shot runs land in the same cwd |
| `test_session_runs_share_the_same_directory_as_oneshot_runs` | `.new` / `.resume` / a second session / one-shot all share that cwd, and the session UUID never reaches the cwd |
| `test_runner_leaves_the_shared_directory_in_place` | the cwd is not torn down after a run (a later `Process.run()` would fail, and deletion is what broke hand-resume) |
| `test_diagnose_runs_in_the_shared_directory` | the Settings → Test path, the third route into `executeProcess` |

No `~/.claude/` content was created, moved, or deleted during any of this; the tests use `/bin/sh` stand-ins that print `$PWD` and never invoke a real LLM CLI.

## Checklist

- [x] Builds locally (`make build`); tests written, partially executed locally — see above for exactly which and why
- [ ] User-facing change is noted in the README changelog (if applicable) — release-note line suggested above, not added (no version bump here)

## Summary by CodeRabbit

- **New Features**
  - LLM operations now use a persistent shared workspace across one-shot commands, ongoing sessions, resumed conversations, and diagnostics.
  - The workspace is created automatically when needed and remains available for future sessions.
  - Live AI conversations can be resumed without removing stored session data.

- **Bug Fixes**
  - Added a fallback workspace when the standard application location is unavailable.
  - Improved consistency across LLM operations and prevented unnecessary workspace removal when cancelling sessions or changing context.

### DIFF
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -774,9 +774,25 @@ enum LLMRunner {
     /// fails outright if `currentDirectoryURL` doesn't exist.
     static func sandboxDirectory() -> URL {
         let url = sandboxDirectory(appSupportRoot: applicationSupportRoot())
-        try? FileManager.default.createDirectory(at: url,
-                                                 withIntermediateDirectories: true)
-        return url
+        do {
+            try FileManager.default.createDirectory(at: url,
+                                                    withIntermediateDirectories: true)
+            return url
+        } catch {
+            // `applicationSupportRoot()` only falls back when the *lookup*
+            // fails. If the directory resolves but can't be created -- a
+            // read-only or permission-broken Application Support -- swallowing
+            // the error and returning the path anyway breaks every subsequent
+            // CLI launch, because `Process.run()` fails outright when
+            // `currentDirectoryURL` does not exist. Degrade to a temp directory
+            // instead: the Claude project slug stops being stable, which costs
+            // resumability, but LLM calls keep working.
+            os.Logger(subsystem: "io.island.whisper.IslandWhisper", category: "LLMRunner").error("llm sandbox unavailable at \(url.path, privacy: .public): \(error.localizedDescription, privacy: .public) -- falling back to a temp directory")
+            let fallback = sandboxDirectory(appSupportRoot: FileManager.default.temporaryDirectory)
+            try? FileManager.default.createDirectory(at: fallback,
+                                                     withIntermediateDirectories: true)
+            return fallback
+        }
     }
 
     /// Pure path composition, split out so tests can assert the layout
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -210,23 +210,12 @@ enum LLMRunner {
         let handle = ProcessHandle()
         return try await withTaskCancellationHandler {
             try await withCheckedThrowingContinuation { continuation in
-                // For Live AI's session continuity to work, every tick
-                // must share a CWD so claude's per-CWD session
-                // storage stays addressable. We key the stable
-                // sandbox off the session UUID.
-                let sandboxKey: String? = {
-                    switch session {
-                    case .none: return nil
-                    case .new(let id), .resume(let id): return id.uuidString
-                    }
-                }()
                 DispatchQueue.global(qos: .userInitiated).async {
                     do {
                         let outcome = try executeProcess(executable: executable,
                                                          arguments: tool.arguments(prompt: fullPrompt, model: model, session: session) + extraArgs,
                                                          timeout: timeout,
-                                                         handle: handle,
-                                                         sandboxKey: sandboxKey)
+                                                         handle: handle)
                         // The Swift Task that drove us was cancelled mid-flight
                         // — `handle` SIGTERM'd the process, so the user-visible
                         // truth is "we cancelled it", not "the CLI crashed".
@@ -400,8 +389,7 @@ enum LLMRunner {
                         let outcome = try executeProcess(executable: executable,
                                                          arguments: args,
                                                          timeout: timeout,
-                                                         handle: handle,
-                                                         sandboxKey: nil)
+                                                         handle: handle)
                         continuation.resume(returning: LLMTestResult(
                             command: command,
                             succeeded: !outcome.timedOut && outcome.exitCode == 0,
@@ -602,8 +590,7 @@ enum LLMRunner {
     private static func executeProcess(executable: URL,
                                        arguments: [String],
                                        timeout: TimeInterval,
-                                       handle: ProcessHandle,
-                                       sandboxKey: String? = nil) throws -> ProcessOutcome {
+                                       handle: ProcessHandle) throws -> ProcessOutcome {
         let process = Process()
         process.executableURL = executable
         process.arguments = arguments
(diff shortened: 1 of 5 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

Issue #177, opened 2026-08-09
Settings is a fixed 700x560 separate window with no room for a tenth tab; make it an in-app page instead


leetwito:
**Problem / use case**

Settings ships as a macOS `Settings { SettingsView() }` scene (`Mila/App/MilaApp.swift`), opened via `SettingsLink` from the sidebar, and pinned to `.frame(width: 700, height: 560)` - a fixed-size, non-resizable utility window separate from the main one.

That container is now full. Nine tabs today:

`General │ Audio │ Models │ AI Provider │ AI Features │ Speakers │ Meetings │ Voice Memos │ Storage`

and the tab strip has been a recurring structural problem rather than a one-off:

- #137 - the strip overflowed the then-560 pt frame.
- #138 - widened the window to 700 pt to make room.
- #131 - folded Updates into General to free a slot.
- #144 - concluded the strip is nine tabs at roughly 595 pt against ~740 pt available: *"It fits, with no room for a tenth."*

So the next setting that doesn't fit an existing tab has nowhere to go, and the two moves available so far - widen the window, or merge tabs - are both spent. Widening again fights the fixed frame; merging again costs clarity, which is what #144 was fixing in the first place.

Separate from the ceiling, being a detached fixed window has day-to-day costs:

1. **You can't see a setting and the thing it affects at the same time.** Tuning a name prompt while looking at the badly-named recording that prompted you to open Settings means switching windows back and forth, from memory.
2. **Dense tabs can't be given more room.** The frame is fixed, so a tab that needs vertical space (AI Features with several expanded sections, Models with the remote block) can't simply be resized - it has to be redesigned to fit 560 pt.
3. **Cross-references aren't navigable.** Inline warnings that point at another tab ("configure the provider on the AI Provider tab") are prose, not links, because there is no in-app navigation to target.

**Proposed solution**

Make Settings another destination in the main window - a sidebar entry alongside the recording list - with its sections as a scrollable list instead of a fixed-width `TabView` strip.

- **Removes the width ceiling permanently.** Sections in a sidebar list cost vertical space, which scrolls, instead of horizontal strip width, which doesn't. A tenth and eleventh section become free.
- **Sections get the whole window.** They inherit the main window's real, resizable geometry rather than a hardcoded 700×560.
- **Cross-references become real navigation** - "configure the provider" can select the AI Provider section directly.
- **Settings sit next to content**, so prompt-tuning against an actual recording stops being a two-window memory exercise.

[...]

PR #188, opened 2026-08-20
fix(settings): make prompt overwrites undoable

### LINKED ISSUE #176: No undo in Settings: "Reset to default" and Examples overwrite a custom prompt with no way back (Cmd+Z does nothing)
**What happened**

`Cmd+Z` does nothing in the Settings text editors. That matters more than usual here, because two **one-click** controls sitting directly above those editors replace a long custom prompt outright, with no confirmation and no way back:

- `AIPromptEditor`'s **"Reset to default"** button does `text = defaultText`.
- `ExamplesView`'s selection callback does `settings.namePrompt = $0` - picking an example **overwrites** the current prompt rather than appending or inserting.

So a carefully tuned multi-line prompt can be destroyed by a single misclick on a link-styled button, and the standard macOS recovery gesture doesn't restore it. The same applies to the other prompt editors on the tab (action prompt, summary prompt) and to the plain `TextField`s elsewhere in Settings.

Supporting evidence from the source: grepping every Swift file at `origin/main` @ `7ff1345` for `undoManager`, `NSUndoManager`, `registerUndo`, and `allowsUndo` returns **zero** matches, so there is no undo wiring anywhere in the app.

**Steps to reproduce**

1. Open Settings → **AI Features** → expand **Recording names**.
2. Type or paste a long custom prompt into the Name prompt editor.
3. Click **Reset to default** (or pick any item under **Examples**).
4. Press `Cmd+Z`.

**Expected behavior**

`Cmd+Z` restores the previous text, as in any standard macOS text view.

At minimum, the two destructive one-click overwrites should be recoverable. Options, roughly in increasing order of effort:

- Register an undo operation for "Reset to default" and for applying an Example, so `Cmd+Z` puts the previous prompt back.
- Give the Settings scene a working undo manager so ordinary typing/deletion in `TextEditor` and `TextField` is undoable too.
- As a cheap stopgap, make "Reset to default" a confirmation, or turn it into a toggle-style "Restore" that can be clicked again to put the custom text back.

**Environment**

- macOS version: 26.5.1 (25F80)
- Mila version: 1.9.2 (58)
- Mac model: Apple M3 Pro (Apple Silicon)
- Feature involved: Settings → AI Features prompt editors (also applies to Settings text fields generally)

**Logs / screenshots**

Not log-producing - it's a missing capability rather than a crash. Code references at `7ff1345`: `AIPromptEditor` (`TextEditor(text: $text)` plus the `Reset to default` button) and the `ExamplesView` callback that assigns to `settings.namePrompt`, both in `Mila/Views/SettingsView.swift`.

urisland:
Closes #176

"Reset to default" and the Examples list each replace a hand-written prompt in one click, and there was no way back — `⌘Z` did nothing anywhere in Settings.

## What changed

`UndoableTextEditor` hosts an `NSTextView` with `allowsUndo` and a dedicated per-editor `UndoManager`, held by `AIPromptEditor` as a `@StateObject` so the history survives the re-render every keystroke causes. It replaces the plain `TextEditor`, covering all four prompt editors — summary, name, action items, and the Live AI system prompt — through one type.

The part that makes the destructive actions recoverable: `updateNSView` detects when the **binding** changed behind the text view's back (Reset to default, an Examples pick, a `.milaconfig` import) and applies it through `shouldChangeText(in:replacementString:)` → `replaceCharacters` → `didChangeText()` rather than assigning `.string`. That puts the overwrite on the same undo stack as typing. Neither destructive call site in `SettingsView.swift` changes — they still just assign to the binding.

Undo is reachable three ways, all driving the one stack:

- `⌘Z` / `⇧⌘Z` via `performKeyEquivalent` on the text view subclass
- the `undo:` / `redo:` responder-chain actions a nil-targeted Edit menu item sends, with validation
- a visible **Undo** link in the editor header, shown while the stack is non-empty

The link exists because a recovery path you can see beats one that depends on where first responder happens to be.

## Verification — please read before merging

`make build` succeeds. Verified independently on a clean derived-data path: exit 0, no errors, and no warnings attributed to either new file.

**The seven tests in `MilaTests/PromptUndoTests.swift` have not been run.** They are app-hosted, and running them on the author's machine launches a freshly ad-hoc-signed `Mila.app`, which triggers a macOS keychain-authorisation prompt on every run — several `MilaTests` files talk to the real login keychain, and an ad-hoc build has no stable signing identity for "Always Allow" to attach to. CI is the right place for them, and CI failing them is a real signal rather than noise.

Two known risks if they do fail:

1. They assume `shouldChangeText` registers the undo action. If AppKit doesn't oblige, `Coordinator.applyExternalChange` needs an explicit `registerUndo(withTarget:)`.
2. They lean on AppKit's automatic undo-group closing via a `settle()` helper that turns the run loop for 50 ms — a plausible flake source, with `test_typing_is_undoable…` the most exposed.

**`⌘Z` itself is unverified.** The code was never exercised in a running app. The reasoning is that SwiftUI's Edit ▸ Undo item is inert in the Settings scene (no undo manager in that environment) and a *disabled* menu item does not consume its key equivalent, so the keystroke should fall through to `performKeyEquivalent` — but that is AppKit inference, not observation. One manual check settles it: type in a prompt, click **Reset to default**, press `⌘Z`. The **Undo** link does not depend on that path.

The reentrancy guard (`isApplyingExternalChange`) and the focus-reclaim branch after an external replacement are likewise unexercised at runtime.

## Scope

Deliberately excluded: the Settings window structure (#177 covers making Settings an in-app page), the AI Provider test-transcript scratch field, and the editors with no one-click destructive overwrite (`LiveAIRecordingView`'s context box, `SendToLLMSheet`).

## Summary by CodeRabbit

- **New Features**
  - Added undo and redo support to prompt editors, including keyboard shortcuts and an enabled Undo control when history is available.
  - Typing, resets, examples, and imports can now be undone and redone.
  - Undo history is maintained separately for each prompt editor.
  - Added accessibility metadata and improved focus and disabled-state handling.

- **Bug Fixes**
  - Undo and redo now reliably synchronize restored text with the editor content.

- **Tests**
  - Added coverage for typing, resets, example selection, redo, empty histories, unchanged edits, and independent editor histories.

### DIFF
--- a/Mila/Views/SettingsView.swift
+++ b/Mila/Views/SettingsView.swift
@@ -963,6 +963,14 @@ private struct AIFeatureRow<Content: View>: View {
 
 /// A prompt editor with a "Reset to default" affordance. Used by three of the
 /// four feature rows and by Live AI.
+///
+/// "Reset to default" and the Examples list next to it each replace a
+/// hand-written prompt in one click, and used to do it irrecoverably — `⌘Z` did
+/// nothing anywhere in Settings (#176). The editor is now `UndoableTextEditor`,
+/// which gives this field a real per-editor undo stack that covers typing *and*
+/// those overwrites, however they reach the binding. `⌘Z` is what users try
+/// first; the "Undo" link is here because a recovery path you can see beats one
+/// that depends on where first responder happens to be.
 private struct AIPromptEditor: View {
     let title: String
     @Binding var text: String
@@ -972,6 +980,10 @@ private struct AIPromptEditor: View {
     var caption: String?
     var isEnabled = true
 
+    /// Survives the re-renders that recreate the editor struct, so the undo
+    /// history outlives a keystroke.
+    @StateObject private var undo = PromptUndoBridge()
+
     var body: some View {
         VStack(alignment: .leading, spacing: 6) {
             HStack {
@@ -979,19 +991,24 @@ private struct AIPromptEditor: View {
                     .font(.caption.weight(.semibold))
                     .foregroundStyle(.secondary)
                 Spacer()
+                if undo.canUndo {
+                    Button("Undo") { undo.undo() }
+                        .buttonStyle(.link)
+                        .font(.caption)
+                        .disabled(!isEnabled)
+                        .help("Put back the previous prompt (\u{2318}Z).")
+                        .accessibilityIdentifier("ai.prompt.undo")
+                }
                 Button("Reset to default") { text = defaultText }
                     .buttonStyle(.link)
                     .font(.caption)
                     .disabled(!isEnabled)
             }
-            TextEditor(text: $text)
-                .font(.system(.callout, design: .monospaced))
-                .scrollContentBackground(.hidden)
+            UndoableTextEditor(text: $text, undo: undo, isEnabled: isEnabled)
                 .frame(minHeight: minHeight, maxHeight: minHeight + 60)
                 .padding(4)
                 .overlay(RoundedRectangle(cornerRadius: 6)
                     .strokeBorder(Color.primary.opacity(0.15), lineWidth: 1))
-                .disabled(!isEnabled)
             if let caption {
                 Text(LocalizedStringKey(caption))
                     .font(.caption)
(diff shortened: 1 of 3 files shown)


urisland:
@coderabbitai full review

Issue #180, opened 2026-08-09
RemoteWhisperEngine hardcodes response_format=verbose_json, locking remote transcription to whisper-1; select the format per model to unlock gpt-transcribe and server-side diarization


leetwito:
**Problem / use case**

`RemoteWhisperEngine.multipartBody` hardcodes the response format for every remote transcription:

https://github.com/island-io/mila/blob/7ff1345/Mila/Transcription/RemoteWhisperEngine.swift#L220

```swift
appendField("response_format", "verbose_json")
```

#178 documents the user-facing consequence (only `whisper-1` works against OpenAI, and the free-text model field invites IDs that 400) and proposes a known-good picker. It explicitly leaves the engineering half open — *"either not offering the `gpt-*` models, or fixing `response_format` per model first so they can be offered."* This issue is that second option.

**The constraint is narrower than "gpt-\* models are unsupported"**

Sending Mila's exact multipart shape (an `.m4a` as `audio/mp4`, plus `language=en`) against `api.openai.com/v1/audio/transcriptions`, sweeping every documented `response_format` — verified 2026-08-09:

| model | `json` | `text` | `srt` | `vtt` | `verbose_json` | `diarized_json` |
|---|---|---|---|---|---|---|
| `whisper-1` | ok | ok | ok | ok | **ok** | 400 |
| `gpt-4o-transcribe` | ok | ok | 400 | 400 | 400 | 400 |
| `gpt-4o-mini-transcribe` | ok | ok | 400 | 400 | 400 | 400 |
| `gpt-transcribe` | ok | ok | 400 | 400 | 400 | 400 |
| `gpt-4o-transcribe-diarize` | ok | ok | 400 | 400 | 400 | **ok** |

The `gpt-*` models reject `srt` and `vtt` too, not just `verbose_json`. Every rejected format is a timestamped one; the only two they accept carry no timing at all. The error is explicit:

```
response_format 'verbose_json' is not compatible with model 'gpt-transcribe-api-ev3'. Use 'json' or 'text' instead.
```

[...]

Issue #144, opened 2026-08-01
Settings: reorganise AI configuration into "AI Provider" + "AI Features" tabs


urisland:
## Problem

Mila's AI configuration lives in **two** Settings tabs today, split by feature
rather than by what the setting *is*:

- **LLM** — the tool picker (Off / Claude CLI / Cursor CLI / OpenAI Compatible),
  executable path, extra CLI args, the OpenAI-compatible endpoint fields
  (provider / base URL / model / API key), the timeout, the automatic-summary
  toggle, the name and action prompts, and the test panel.
- **Live AI** — the Live AI enable toggle, the output language, the live system
  prompt, and an Advanced block (live model override, VAD, background mode,
  cadence/cost dials).

That split is confusing in a few concrete ways:

1. **The provider looks like it needs configuring twice.** Live AI is powered by
   whatever tool is set on the LLM tab, but nothing on the Live AI tab says so —
   it just shows a warning pointing at the other tab once you're already there.
2. **A genuinely shared setting is hidden in the wrong place.** "Output
   language" sits on the Live AI tab, but it fills the `{{LANGUAGE}}` slot for
   the *post-recording automatic summary* too (`RecordingSummarizer` and
   `LiveAISession` both read `LiveAISettings.outputLanguage`). Users who never
   enable Live AI never find the control that decides what language their
   summaries come back in.
3. **The automatic-summary prompt isn't editable at all.** `liveAI.summaryPrompt`
   is persisted and used on every recording, but no tab surfaces it — the LLM
   tab only has the on/off toggle.
4. **Neither tab is skimmable.** Both are dense walls of same-size caption text,
   so you can't tell at a glance which AI features are actually on.

## Design

**Superseded:** the first attempt merged both tabs into a single **AI** tab with
the provider on top and five collapsible feature sections below. Built, run and
rejected — see the "what went wrong" note at the bottom. The accepted design is
below.

Two tabs, split by **how often you touch them** rather than by which feature
owns the setting.

### Tab A — "AI Provider"

Technical, set once, frequently supplied by a `.milaconfig` import rather than
typed by hand.

- Tool picker as a pop-up menu sized to its content — *not* a full-width
  segmented control.
- Only the fields the selected tool needs: executable path + extra CLI args for
  the CLIs; provider preset / base URL / model / Keychain API key for
  OpenAI-compatible.
- Timeout as an editable text field **with** a stepper, plus a secondary reset.
- The connection test panel.
[...]

PR #191, opened 2026-08-21
feat(settings): pick remote transcription models from a known-good list

### LINKED ISSUE #178: Remote transcription invites free-text model IDs while only whisper-1 works against OpenAI; add a known-good picker and demote free text to Advanced
**Problem / use case**

The Models tab configures the remote backend with three plain free-text fields - endpoint, model, and the optional English model - plus the Keychain API key. Their only guidance is placeholder text (`https://api.openai.com/v1`, `whisper-1`, `Same as above`). Nothing tells you which model IDs Mila can actually use, so picking one is trial and error.

That isn't hypothetical - **most of OpenAI's current transcription models cannot work with Mila at all**, and the UI happily invites you to type them in.

`RemoteWhisperEngine.multipartBody` hardcodes `response_format=verbose_json`, and every `gpt-*` transcription model rejects that format outright. Verified against `api.openai.com/v1/audio/transcriptions` on 2026-08-09, sending Mila's exact multipart shape (an `.m4a` as `audio/mp4`, plus `language=en`):

| model | `verbose_json` |
|---|---|
| `whisper-1` | accepted - returns `segments` |
| `gpt-4o-transcribe` | **HTTP 400** |
| `gpt-4o-mini-transcribe` | **HTTP 400** |
| `gpt-transcribe` | **HTTP 400** |
| `gpt-4o-transcribe-diarize` | **HTTP 400** |

The 400 reads:

```
response_format 'verbose_json' is not compatible with model 'gpt-transcribe-api-ev3'. Use 'json' or ...
```

So on OpenAI's hosted API, `whisper-1` - a 2023 model - is the only one that works, and a user who reasonably types the newest model name gets an opaque failure with no hint that the model name was never the problem.

Two things make this worse:

1. **"Test connection" goes green anyway.** It probes `GET /models` and treats any 2xx as reachable (and 404 as "reachable, no /models endpoint"). It never attempts a transcription, so it cannot detect that the configured *model* will 400 on every real recording.
2. **Mila already encodes model knowledge - just not visibly.** `RemoteTranscriptionSettings.isHebrewOnlyModel()` substring-matches `"ivrit"` to decide whether a separate English model is needed, and `defaultEnglishModel` is a specific Hugging Face repo id (`deepdml/faster-whisper-large-v3-turbo-ct2`). There is effectively an implicit catalogue in the code; the UI just doesn't expose it.

**Proposed solution**

Lead with a **picker of known-good presets**, and demote the raw fields to Advanced.

- A `Picker` listing configurations Mila is known to work with, each setting endpoint + model + the English-model prefill **together** as one coherent choice rather than three fields the user has to get consistent:
  - OpenAI hosted → `https://api.openai.com/v1` + `whisper-1`
  - self-hosted `speaches` / faster-whisper (Hebrew) → the `ivrit-ai/...` finetune, with the English model prefilled - the case `isHebrewOnlyModel` already special-cases
  - self-hosted multilingual → `deepdml/faster-whisper-large-v3-turbo-ct2`
  - **Custom…** as the final item
- The existing endpoint / model / English-model `TextField`s move behind an **Advanced** disclosure, revealed when *Custom…* is selected (and available but collapsed otherwise). Free-text stays fully supported - it just stops being the default surface, which matches how #144 treated "technical, set once" configuration.
- **Presets must only list what Mila can actually parse.** Given the table above, that means either not offering the `gpt-*` models, or fixing `response_format` per model first so they can be offered. Either is fine; silently offering a model that 400s is not.

**Alternatives considered**

- *Populate the list from `GET /models`.* Doesn't solve it: the endpoint lists many models that aren't transcription models, and Mila's own `verbose_json` constraint isn't discoverable from that response at all. It would produce a longer list of mostly-broken choices.
- *Improve the placeholder text / add a help tooltip.* Documents the trap instead of removing it, and still 400s at transcription time.
- *Detect the 400 and retry with `json`.* Worth doing regardless as a robustness fix, but it changes behaviour silently (dropping `segments` costs per-segment timestamps and therefore speaker assignment), so it's not a substitute for telling the user what's supported.

**Additional context**

A companion improvement, if it's wanted: make "Test connection" do a one-shot transcription of a tiny bundled clip with the configured model and format, instead of `GET /models`. That would turn the green pill into a real guarantee and would have caught this class of failure directly.

Inspected at `origin/main` @ `7ff1345`; API behaviour verified live on 2026-08-09.

urisland:
Closes #178

## What was wrong

Settings → Models configured the remote backend with a free-text **Model** field whose only guidance was a `whisper-1` placeholder. Nothing told you which model ids Mila could actually talk to, so picking one was trial and error — and before #180 / #189 most of OpenAI's transcription line-up answered Mila's request with an opaque HTTP 400, because `response_format=verbose_json` was hardcoded.

#189 fixed the engineering half: `RemoteWhisperEngine.ResponseFormat.forModel(_:)` now selects the format per model, so `gpt-transcribe`, `gpt-4o-transcribe`, `gpt-4o-mini-transcribe` and `gpt-4o-transcribe-diarize` are genuinely reachable. This PR is the UI half the issue deferred.

## What changed

**Model is a picker, not a text field.** `RemoteModelPreset.all` (new, `Mila/Models/RemoteModelPreset.swift`) is the catalogue, grouped into two picker sections plus a `Custom…` row:

| Group | Entries |
|---|---|
| OpenAI hosted | `whisper-1`, `gpt-transcribe`, `gpt-4o-transcribe`, `gpt-4o-mini-transcribe`, `gpt-4o-transcribe-diarize` |
| Self-hosted (speaches / faster-whisper) | `ivrit-ai/whisper-large-v3-turbo-ct2`, `deepdml/faster-whisper-large-v3-turbo-ct2` |

Selecting an entry sets the model id and, where the configured endpoint is another preset's canonical endpoint (or isn't a usable URL), the endpoint too — so a choice lands coherent instead of as three fields the user has to reconcile. A URL the user typed themselves is **never** overwritten: a private gateway or reverse proxy can serve any of these ids, and repointing it on an unrelated model change would break a working setup.

**The timestamp tradeoff is stated at the point of choice.** The `gpt-*` models reject every timestamped format, so a transcript from one arrives as a single untimed segment — no SRT timings, and local diarization has nothing to align its speaker turns to. The picker shows a caption per selection and an orange notice for the untimed models, rather than letting that surface later as a mysteriously flat transcript. `gpt-4o-transcribe-diarize` is deliberately *not* warned about: it returns per-speaker-turn timings.

**Free text is demoted, not removed.** The raw **Model id** and **English model** fields moved into an `Advanced` disclosure, matching how #144 treated "technical, set once" configuration. It opens itself when the configured model isn't a preset, so a hand-typed id — or one arriving via `.milaconfig` — is visible immediately rather than hidden behind a collapsed row.

### Two structural choices worth reviewing

1. **No second model→format table.** `RemoteModelPreset.timestamps` is read off `RemoteWhisperEngine.ResponseFormat.forModel(id).timestamps` — the same code that decides what to put on the wire. `#189`'s `ResponseFormat` gained a `TimestampSupport` property (`perSegment` / `perSpeakerTurn` / `none`) so the capability lives with the format that determines it. Nothing in the UI re-derives anything from a model id.
2. **`selectedPreset` is derived from `model`, not persisted beside it.** A stored "selected preset" would be a second copy of a fact `model` already carries, and `MilaConfigImporter` writes `model` directly — so it could disagree. Deriving it means an imported configuration shows up correctly identified in the picker with **no importer change at all**. Likewise `apply(_:)` sets `model` and lets its existing `didSet` reconcile `englishModel`, so the ivrit.ai prefill/withdraw rule stays in the one place that owns it.

## Deviations from the issue, and why

- **Endpoint and API key stay at the top level** rather than moving into Advanced. They are per-deployment necessities — a self-hosted user (the `mila-asr` case) must type a URL every time — and neither is the trap the issue is about. Burying the endpoint would make Mila's primary real-world configuration harder, not easier. The model field is what invited a silent 400.
- **No "endpoint doesn't match this model's host" warning.** It would have been the natural companion, but **Test connection** was since upgraded to perform a real one-shot transcription of the bundled clip (it no longer probes `GET /models` — the issue's complaint #1 is already stale on `main`), so a mispaired endpoint/model already fails loudly and specifically. A second heuristic warning would just be noise.
- **`matching(_:)` is an exact match, not the tolerant substring style of `isHebrewOnlyModel(_:)`.** It decides whether to *describe* a model to the user with a preset's guarantees; doing that for a near-miss id (`whisper-1-preview`, `openai/gpt-4o-transcribe`) would be a claim about a model Mila has never seen. `forModel(_:)` stays tolerant, so a near-miss still gets the right wire format — it just reads as `Custom…`.

Model ids and their accepted `response_format` values were re-checked against the OpenAI docs, not taken from the issue text; they agree with #189's table, which `docs/REMOTE_TRANSCRIPTION_SERVER.md` keeps (now cross-referenced to the picker).

## Verification

`make build` — **BUILD SUCCEEDED**, no new warnings.
`xcodebuild … build-for-testing` — **TEST BUILD SUCCEEDED**, so the new test sources compile.

**Tests written but not executed locally.** `MilaTests/RemoteModelPresetTests.swift` (17 tests, in two classes) is committed and compile-verified, but **I did not run it.** App-hosted `MilaTests` launch a freshly ad-hoc-signed `Mila.app`, and several files in that target read the real login keychain, which pops a macOS password dialog on the maintainer's machine; they asked for that to stop. CI's `unit-and-ui-tests` job is the first execution of these:

- `RemoteModelPresetTests` — every preset has a usable id/endpoint/caption; **every preset resolves to a `response_format` its model family actually accepts** (the load-bearing one: `forModel` falls back to `verbose_json` for unknown ids, which is precisely what the `gpt-*` models reject, so "it returned something" is not evidence of reachability); no duplicate ids; every group is non-empty and every preset belongs to a listed group; the default model is a preset (otherwise a fresh install opens on `Custom…`); timestamp classification per family; `matching(_:)` is case-insensitive, whitespace-tolerant, and rejects near-misses; endpoint normalisation.
- `RemoteModelPresetSettingsTests` — `selectedPreset` derives from the model id; `apply(_:)` sets model + endpoint together, inherits the English-model prefill and its withdrawal, moves back to OpenAI's endpoint, **leaves a user-typed endpoint alone**, fills in an unusable one, treats a case/slash-equivalent preset endpoint as already correct, and persists across instances.

Each uses an isolated `UserDefaults` suite and an isolated keychain item key, per the repo convention.

## Not done

- No `TranscriptionBackend`-level change and no new `@EnvironmentObject`: this adds no app-wide settings object, only a pure value type plus two members on the existing `RemoteTranscriptionSettings`, so the Environment Objects convention doesn't come into play.
- The picker is not populated from `GET /models` — for the reason the issue gives: that list is mostly non-transcription models and says nothing about format compatibility.
- UI not visually confirmed on a running app. Verified by build only, per the maintainer's standing request not to launch app-hosted tests or steal focus; the screenshot check is worth doing before merge.

## Summary by CodeRabbit

* **New Features**
  * Added selectable presets for hosted and self-hosted remote transcription models.
  * Automatically configure compatible model, endpoint, and English-model settings.
  * Preserved support for custom models and endpoints.
  * Added warnings for models without timestamp support and related SRT or speaker-label limitations.
  * Improved grouped model selection and endpoint guidance in Settings.

* **Documentation**
  * Updated remote transcription setup instructions with available models and timestamp behavior.

* **Tests**
  * Added coverage for presets, matching, endpoint handling, and settings persistence.

### DIFF
--- a/Mila/Models/RemoteModelPreset.swift
+++ b/Mila/Models/RemoteModelPreset.swift
@@ -1,0 +1,151 @@
+import Foundation
+
+/// One entry in the remote-transcription model picker: a model id Mila is
+/// known to be able to *talk to*, together with the endpoint it belongs to.
+///
+/// The picker exists because the Model field used to be plain free text whose
+/// only guidance was a placeholder, and most of OpenAI's transcription models
+/// answer a request Mila can't form with an opaque HTTP 400. Naming the ones
+/// that work turns "type a model id and find out" into a choice (issue #178).
+///
+/// **Every preset must be reachable by `RemoteWhisperEngine`.** That is not a
+/// convention to remember: `timestamps` is read off
+/// `RemoteWhisperEngine.ResponseFormat.forModel(id)`, so a preset is described
+/// by the same code that decides what to send for it. There is no second
+/// model→format table to drift out of step with the first.
+struct RemoteModelPreset: Identifiable, Hashable, Sendable {
+
+    /// Where the model runs — the picker's section headers.
+    enum Group: String, CaseIterable, Sendable {
+        /// OpenAI's hosted API. Fixed endpoint, key required.
+        case openAI
+        /// A server the user runs (`speaches`, faster-whisper, vLLM, …).
+        case selfHosted
+
+        var displayName: String {
+            switch self {
+            case .openAI:     return "OpenAI hosted"
+            case .selfHosted: return "Self-hosted (speaches / faster-whisper)"
+            }
+        }
+    }
+
+    /// The model id sent as the multipart `model` field. Doubles as the
+    /// identity, since a picker never needs two rows for one id.
+    let id: String
+    /// Short label for the picker row.
+    let displayName: String
+    let group: Group
+    /// The endpoint this model lives behind, when there is a canonical one.
+    ///
+    /// Applied by `RemoteTranscriptionSettings.apply(_:)` only when the
+    /// endpoint currently configured is another preset's canonical endpoint (or
+    /// unusable) — a URL the user typed themselves is never overwritten, since
+    /// a private gateway can proxy any of these models.
+    let endpoint: String
+    /// One line under the picker explaining what this choice is for.
+    let detail: String
+
+    /// What timings a transcript from this model will carry. Derived, never
+    /// declared: see the type's note.
+    var timestamps: RemoteWhisperEngine.ResponseFormat.TimestampSupport {
+        RemoteWhisperEngine.ResponseFormat.forModel(id).timestamps
+    }
+
+    /// The endpoints presets own, and may therefore replace when the user
+    /// switches groups. Anything else is the user's and is left alone.
+    static var canonicalEndpoints: Set<String> {
+        Set(all.map { Self.normalizeEndpoint($0.endpoint) })
+    }
+
+    /// Compare endpoints ignoring case and a trailing slash, so
+    /// `https://API.openai.com/v1/` counts as the OpenAI endpoint.
+    static func normalizeEndpoint(_ raw: String) -> String {
+        var s = raw.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+        while s.hasSuffix("/") { s.removeLast() }
+        return s
+    }
+
+    /// The endpoint the docs' self-hosted quickstart uses
+    /// (`docs/REMOTE_TRANSCRIPTION_SERVER.md`, Option A).
+    static let selfHostedEndpoint = "http://localhost:8000/v1"
+
+    /// Every configuration Mila is known to work with.
+    ///
+    /// The OpenAI half is the full transcription line-up as of 2026-08, all of
+    /// which `RemoteWhisperEngine` can now reach — `whisper-1` via
+    /// `verbose_json`, the `gpt-*` family via `json`, and the diarizing variant
+    /// via `diarized_json` (issue #180 / PR #189). Before that the engine
+    /// hardcoded `verbose_json` and only `whisper-1` worked, which is why this
+    /// list could not have been offered earlier.
+    static let all: [RemoteModelPreset] = [
+        RemoteModelPreset(
+            id: RemoteTranscriptionSettings.defaultModel, // "whisper-1"
+            displayName: "whisper-1",
+            group: .openAI,
+            endpoint: RemoteTranscriptionSettings.defaultEndpoint,
+            detail: "Multilingual, with per-segment timings. The safe default — SRT export and local speaker labels both work."),
+        RemoteModelPreset(
+            id: "gpt-transcribe",
+            displayName: "gpt-transcribe",
+            group: .openAI,
+            endpoint: RemoteTranscriptionSettings.defaultEndpoint,
+            detail: "OpenAI's current general-purpose transcriber. Usually more accurate text than whisper-1."),
+        RemoteModelPreset(
+            id: "gpt-4o-transcribe",
+            displayName: "gpt-4o-transcribe",
+            group: .openAI,
+            endpoint: RemoteTranscriptionSettings.defaultEndpoint,
+            detail: "GPT-4o audio transcription."),
+        RemoteModelPreset(
+            id: "gpt-4o-mini-transcribe",
+            displayName: "gpt-4o-mini-transcribe",
+            group: .openAI,
+            endpoint: RemoteTranscriptionSettings.defaultEndpoint,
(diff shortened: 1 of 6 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

Issue #179, opened 2026-08-09
Reference the transcript by file path instead of inlining it in argv (the .srt already has timestamps the prompt drops)


leetwito:
**Problem / use case**

`LLMRunner.composedPrompt` inlines the entire transcript into the prompt argument, and the doc comment explains why: `cursor-agent -p` only reads argv, so argv is the portable shape. That reasoning is sound for a one-shot "summarize this" call. It works against you as soon as the prompt drives an **agentic** action, which is what the Send-to-LLM feature enables.

Concretely, inlining loses information that Mila already has on disk:

1. **Timestamps are dropped.** `TranscriptFormatter.plainText` keeps speaker labels (collapsing consecutive same-speaker segments into `SPEAKER_00: …` paragraphs) but carries no times. The auto-written `.srt` sidecar has both - speaker *and* `00:00:07,000 --> 00:00:14,000`. So the prompt gets a strictly poorer rendering of a file sitting right next to the audio.

2. **The agent can't re-read, seek, or grep.** An inlined blob is read once, in order, at whatever position it landed in the context. A path lets the model read the part it needs, come back for more on a second pass, and quote exact timestamps.

3. **The richer artifacts already exist and already have computed paths.** `.txt`, `.srt` and `.summary.txt` sidecars are auto-written alongside the audio, and `Recording.srtFileName` already derives the `.srt` path from `audioFileName`. Nothing new has to be produced - the proposal is to *reference* what's written today.

4. **Every run re-sends the whole transcript** even when the action needs one section of it.

**To be explicit about what this is *not* about:** argv size limits. Measured across the 74 transcribed recordings in my library, the largest transcript body is ~65 KB against an `ARG_MAX` of 1048576 - about 6%. Even a very long meeting isn't close, so this is a fidelity and capability argument, not a "we'll hit the limit" argument.

**Proposed solution**

Add a transcript-delivery mode. When *reference by path* is selected, `composedPrompt` emits the sidecar paths instead of the body:

```
<user prompt>

---
Transcript (diarized, with timestamps): /…/Recordings/<name>.srt
Transcript (plain text):                /…/Recordings/<name>.txt
Audio:                                  /…/Recordings/<name>.m4a
Read the transcript file(s) above before answering.
```

Constraints that should hold:

- **Inlining stays the default.** `cursor-agent`'s argv-only behaviour and any provider without filesystem access must keep working exactly as they do now.
[...]

Issue #158, opened 2026-08-02
LLM executable lookup: honour a configured npm prefix; Executable-override tooltip is stale


urisland:
Follow-ups from the CodeRabbit review on #119, deferred so they don't hold up that PR. Neither is a defect in #119 — #119 is a strict improvement in both areas — but both are worth closing out.

## 1. Discover the *configured* npm prefix, not just the conventional one

`LLMRunner.searchDirectories()` adds `~/.npm-global/bin`. That is one convention, not the actual npm prefix. A user who set `prefix=/somewhere/else` in `~/.npmrc` still won't have `claude` / `cursor-agent` / `gemini` auto-discovered from a GUI-launched Mila, and has to paste an absolute path into Settings -> AI Provider -> Executable.

Suggested shape (kept subprocess-free on purpose — asking `npm config get prefix` needs `npm` on `PATH`, which is exactly what we don't have in a launchd-stripped environment, and it costs a process spawn on every lookup):

- read `~/.npmrc`, find a `prefix=` line, expand a leading `~` / `$HOME` / `${HOME}`, append `/bin`
- ignore blank, relative, or commented values so a malformed file degrades to "no extra dir"
- `searchDirectories(home:pathEnv:fileManager:)` already takes injectable `home` + `fileManager`, so this is straightforwardly unit-testable against a temp home

Worth a test with a prefix that is *not* `~/.npm-global`, plus one for tilde expansion and one for an unusable value.

## 2. The Executable override tooltip oversells when you need it

`Mila/Views/SettingsView.swift`, the `help:` on the "Executable" row:

> Set this when `claude` / `cursor-agent` / `gemini` lives somewhere a GUI app won't see by default — ~/.local/bin, an asdf shim, a Homebrew prefix that isn't on the system PATH.

Every example it lists (`~/.local/bin`, asdf shims, Homebrew) is now searched automatically by `searchDirectories()`, so the text invites manual configuration that is no longer needed. It also enumerates three tool names, which the AI-tab rework in #145 was deliberately moving away from.

Suggested reword, shorter and tool-agnostic:

> Only needed when Mila can't find the CLI itself — it already searches $PATH, Homebrew, ~/.local/bin and the node version managers (nvm, Volta, asdf).

Note the visible caption below the field (`Leave blank to look the binary up on $PATH.`) is fine and should stay as-is — this is only the hover tooltip.

PR #192, opened 2026-08-21
feat(llm): log every LLM CLI invocation to the unified log

### LINKED ISSUE #175: LLM CLI runs are unobservable: LLMRunner logs nothing, and the exit-code/stderr diagnostic is wired only to the Settings test panel
**Problem / use case**

When Mila shells out to the `claude` CLI for a recording name, an auto-summary, or a Send action, there is no way to see what actually ran. I can't tell whether a title is being generated right now, whether it succeeded, or why it failed - which makes both debugging and simple progress tracking guesswork.

The gap is specifically in `LLMRunner`, not in its callers:

1. **`LLMRunner` logs nothing.** It `import os` but declares zero `Logger`, while 17 other files in the app use `Logger(subsystem: "io.island.whisper.IslandWhisper", ...)`. Its only output is 4 `print()` calls, and those go to stdout - which is not captured by Console.app for a GUI-launched `.app`, so in practice they are invisible to a user.

2. **The rich diagnostic already exists, but only for the test panel.** `LLMRunner.diagnose()` captures exactly what's needed - the copy-pasteable command line, exit code, stdout and stderr - and its own doc comment calls it "the 'why isn't my LLM working?' debugging path". It is called from exactly one place: `Mila/Actions/LLMSettings.swift`. Real runs go through `run()`, which maps the outcome onto a thrown `LLMRunnerError` and discards the argv, the exit code, and stderr.

3. **So a real failure yields only a localized string.** The callers do log lifecycle events - `PostRecordingCoordinator` logs `send succeeded ... -> N chars` and `send failed ...: <localizedDescription>`, and `RecordingSummarizer` logs started/skipped/backfill states. But none of them can log what they don't receive: the resolved executable, the arguments, the exit code, or the stderr tail.

4. **Nothing persists.** `activityStatus` is a transient `@Published String?`. Once it clears there is no record that a run happened at all, so there's no way to answer "did the name generation fire for this recording, and what did it return?"

**Proposed solution**

1. Give `LLMRunner` a `Logger` (`subsystem: "io.island.whisper.IslandWhisper"`, `category: "LLMRunner"`), matching the rest of the app, and convert the 4 `print()` calls to it.

2. Log one line when a run starts and one when it ends:
   - start: feature (name / summary / action), tool, resolved executable path, model, session mode, `extraArgs` count, and prompt/transcript **character counts**
   - end: exit code, wall-clock duration, stdout/stderr byte counts, and on failure a bounded stderr tail

   Metadata as `privacy: .public`; prompt and transcript bodies must stay `.private` - they are meeting content. This mirrors how `RecordingSummarizer` already logs (`transcript=\(transcript.count)c`) rather than logging the text.

3. Route real runs through the same capture `diagnose()` uses, so a non-zero exit or a timeout records the exit code and stderr instead of collapsing to `errorDescription`.

4. Optionally surface a short **Recent LLM runs** list (last ~20: timestamp, feature, tool, exit code, duration, and the copy-pasteable command). `diagnose()` already produces the copy-pasteable `command` string, so most of this exists - it just isn't reachable from real runs.

**Alternatives considered**

- *Rely on Console.app as-is.* Doesn't work today: the only in-runner output is `print()`, which never reaches the unified log.
- *Reuse the Settings test panel.* It runs a synthetic test transcript against the current config; it can't tell you what happened during an actual recording's run, which is the case that fails in the wild (stale executable path, PATH not inherited, non-zero exit, timeout).
- *Log the full prompt.* Rejected - it would put meeting transcripts in the system log.

**Additional context**

The motivating case is an automated post-recording hook: Mila fires `claude -p` with a prompt that runs a skill, and when that silently does nothing there is currently no artifact to inspect - no command, no exit code, no stderr. Even just the start/end log lines in (2) would make it diagnosable.

urisland:
Closes #175

## What & why

`LLMRunner` was the one link in the LLM chain that logged nothing. It `import os` but declared no `Logger`, and its only output was four `print()` calls — which never reach the unified log for a GUI-launched `.app`, so in practice they were invisible. The rich diagnostic that already existed (`diagnose()`: argv, exit code, stdout, stderr) was wired only to the Settings → AI Provider test panel, so a real failure during a recording — stale executable path, non-zero exit, timeout — surfaced as one localized sentence with the exit code and stderr thrown away.

Now every invocation leaves an artifact behind.

**Start line** (`notice`) — feature, tool, resolved executable path, model, session mode, `extraArgs` count, and prompt/summary/transcript/composed character counts:

```
llm run start feature=name tool=claude exe=/opt/homebrew/bin/claude model=(default)
  session=none extraArgs=0 prompt=142c summary=0c transcript=8210c composed=8375c timeout=300s
```

**End line** — exit code, wall-clock duration, stdout/stderr byte counts. `notice` on success; `error` on timeout or non-zero exit, and then it also carries the redacted argv and a bounded stderr tail:

```
llm run end feature=summary tool=claude exit=0 duration=6.41s stdout=1284B stderr=0B

llm run failed feature=action tool=claude exit=1 duration=0.94s stdout=0B stderr=118B
  cmd=/opt/homebrew/bin/claude -p <prompt:8375c> --model haiku --api-key <value:41c>
  stderr-tail=<private>
```

That failure line is issue #175's item 3 — "route real runs through the same capture `diagnose()` uses". `run()`'s throwing contract is unchanged, but the full outcome is logged *before* the throw collapses it into `errorDescription`, so the exit code and stderr no longer vanish with it.

Also in scope, because each was a hole of the same kind:

- **Setup and launch failures now log.** `.toolDisabled`, `.executableNotFound`, and `Process.run()` refusing all bypass `executeProcess`, so the single most common real-world failure — a stale or missing executable path — previously produced no line at all.
- **All four `print()` calls converted**, at levels picked by meaning rather than uniformly. The post-normal-exit pipe-drain timeout is `error` because that output *is* returned to the caller and may be truncated mid-stream (a silently short summary); the post-kill one is `notice` because the output is already being discarded. A child surviving SIGKILL is `error`.
- **The OpenAI-compatible HTTP path** gets the same start/end/cancel/failure events — it was equally invisible, and "wrong endpoint" vs "endpoint rejected us" is exactly the distinction people need.
- **#187's `sandboxDirectory()` fallback** now logs a `notice` naming where it landed when it engages, not just the `error` that caused it. The fallback costs Claude Code project-slug stability, so "why did my resumable conversations scatter?" should be answerable.
- **`LLMFeature`** so a line says *why* a CLI ran: `name` / `summary` / `action` / `live-ai` / `settings-test`. Threaded through `PostRecordingCoordinator.RunLLM` because the auto-title and Send paths share that one seam and would otherwise be indistinguishable; fixed at the call site everywhere else. The parameter is last in `run()`'s signature and defaulted, so no existing call site had to change.

### Level choices

The rule was: *a user diagnosing a failed summary should not need `--info --debug`.* Most of Mila's output sits at debug level, so anything below `notice` is effectively invisible to someone running a plain `/usr/bin/log show`. So start, success, and every failure are `notice`/`error`. Deliberately `debug`: the redacted argv on a *healthy* run (it repeats what the start line already says, and would double log volume for nothing — the failure lines carry it at their own level, so it is never missing when it matters), and cancellations (the user discarded or re-ran; not a fault, but worth having when someone asks "why did my summary never appear?").

### Privacy choices

**No prompt, transcript, or credential text reaches the log by default.** Bodies appear as `.count` and nothing else — mirroring how `RecordingSummarizer` already logs `transcript=\(transcript.count)c`. The guarantee is structural rather than a matter of annotation discipline: the only strings interpolated into a `Logger` call are ones built by the pure helpers in the new `// MARK: - Observability` block, and none of them can carry body text. That is what the tests pin.

The logged command line is a **redacted twin** of the panel's copy-pasteable one. Two positions get replaced by a `<kind:Nc>` placeholder:

1. **The composed prompt** — it is the meeting transcript.
2. **The values of the user's "Extra args"** — free text from Settings, and people put credentials in them (`--api-key sk-…`, a bearer token, an `--append-system-prompt` blob). Mila can't know which of a third-party CLI's flags take a secret, so no extra-argument *value* is logged. Flag *names* survive, including the glued `--flag=value` form (name kept, value redacted, split on the first `=` only). Mila's *own* generated arguments — `--model`, `--session-id`, `--resume`, `-f`, `--skip-trust` — stay visible, because those values are ones Mila chose rather than user input, and a wrong model name is a common cause of a non-zero exit. The user slice is identified by position (argv is always `tool.arguments(…) + extraArgs`), not by value-matching, so an extra arg that happens to equal one of Mila's own values can't be mistaken for trusted.

Placeholders are left unquoted, so the result can't be mistaken for something runnable.

Fields are `privacy: .public` **except `stderr-tail`, which is `.private`.** Public is the deliberate default for metadata because `.private` renders as `<private>` for the very person reading their own log, who is the entire audience. But stderr is not metadata: a CLI run verbosely can echo the prompt (the transcript), and the non-zero-exit path falls back to stdout, which is the model's answer *about* the meeting. For an app whose premise is that recordings stay on the machine, a structured always-on excerpt of that content is not something to enable by default. What stays public is the triage signal — exit code, stderr/stdout byte counts, duration, redacted argv — so "exit 1, 4 KB of stderr" is legible without private-data logging; only the payload is held back. Full stderr remains available in the Settings test panel, which is content the user explicitly asked to see.

Two more worth naming:

- **`exe=` (the resolved executable path)** is public. "Mila is running a different binary than my shell is" is a failure this exists to catch, and the path is useless redacted. It can contain the account's short name — the same exposure the pre-existing sandbox-path line already accepts.
- **The OpenAI path logs `host=`, never the full URL** — a user's endpoint can carry a token in its query string. Its error messages stay public: they are server-generated diagnostics parallel to an HTTP status (`Authentication failed (401)`), taken from the endpoint's error body, not from a completion.

## How it was tested

`make build` — **BUILD SUCCEEDED**. Also `xcodebuild … build-for-testing` — **TEST BUILD SUCCEEDED** — to confirm the `MilaTests` target (the new file plus the two updated `runLLM` stubs) compiles. That is a *build*, not a test run: nothing was executed and no app was launched. Both were re-run after the review fixes.

### Tests written but NOT executed locally

`MilaTests/LLMRunnerObservabilityTests.swift` (20 tests) is committed and compiles, but **I did not run it**, and I have no local pass/fail evidence for it. CI's `unit-and-ui-tests` job is the first thing to actually execute it.

Why not: every `MilaTests` run is app-hosted — it launches a freshly ad-hoc-signed `Mila.app`, and several files in the target read the real login keychain, which pops a macOS password dialog at the machine's owner. `LLMRunnerTests` in the same target also holds smoke tests that invoke the real `claude` / `cursor-agent` / `gemini` binaries and can wedge. Running them on the developer's active machine was explicitly off-limits for this change.

The new tests are all pure and process-free — nothing spawns a subprocess or touches the filesystem, which is also why they live in their own class rather than in `LLMRunnerTests`:

- `redactedCommand`: replaces the prompt with `<prompt:Nc>`; **never contains transcript or prompt text** (the privacy invariant, asserted directly, so a refactor that stops redacting fails here rather than in someone's log); redacts a credential-shaped extra-argument value; splits the glued `--flag=value` form; keeps Mila's own generated arguments readable; redacts nothing but the prompt when `extraArgsCount` is 0; doesn't wildcard-match empty argv entries when the prompt is empty; quotes paths/args containing spaces.
- `redactedExtraArg`: bare flags survive, bare values don't, and a value containing `=` splits on the first one only.
- `stderrTail`: folds newlines and drops blank lines; keeps the **tail** on truncation (`…` + exactly `limit` chars); empty for whitespace-only stderr; untouched under the limit.
- `modelTag` / `sessionTag` / `durationTag`: `(default)` vs a pinned model, all three session modes, fixed 2-decimal seconds.
- `LLMFeature` raw values pinned as stable grep tokens, and asserted whitespace-free so a token can't split a `key=value` pair.

One thing the tests **cannot** cover: OSLog exposes no way to assert a field's privacy level from a unit test, so `stderr-tail=` being `.private` is enforced by code review, not by a test.

## Deliberately not included

**Issue #175 item 4, the "Recent LLM runs" list.** The issue marks it optional, and it is the only part that needs new state and a new UI surface — a ring buffer with nothing rendering it would be dead code. Worth its own issue if wanted.

**Tightening `LLMRunnerError.nonZeroExit`.** Its `errorDescription` embeds stderr verbatim, and `PostRecordingCoordinator` logs that `.public` — so full stderr already reaches the log on the failure path today, independently of this PR. Narrowing that is a behaviour change to a user-visible error string and belongs in its own change.

Also untouched on purpose, to avoid colliding with work in flight in this same file: the transcript-passing mechanism (#179) and executable discovery (#158).

## Checklist

- [x] Builds locally (`make build`); tests written and committed but **not** run locally — see above
- [ ] User-facing change is noted in the README changelog — n/a, no user-visible behaviour change (logging only)

## Summary by CodeRabbit

* **Improvements**
  * Added unified, privacy-safe logging for Live AI, summaries, title suggestions, and actions.
  * Logs now capture execution outcomes, durations, model and session details, and diagnostic information without exposing prompts, transcripts, credentials, or user-entered command values.
  * Improved diagnostics for process execution, HTTP requests, timeouts, and sandbox fallback behavior.
* **Bug Fixes**
  * Preserved existing cancellation, timeout, error-handling, and diagnostic result behavior while improving execution monitoring.

### DIFF
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -1,6 +1,45 @@
 import Foundation
 import os
 
+/// Unified-log destination for every LLM invocation. Same subsystem as the
+/// rest of the app so `log stream --predicate 'subsystem == "…"'` catches
+/// runner lines alongside the callers' own lifecycle lines
+/// (`PostRecordingCoordinator`, `RecordingSummarizer`).
+///
+/// File-level rather than a member of `LLMRunner` so `ProcessHandle` and the
+/// error types can reach it too, matching `PostRecordingCoordinator`'s
+/// `postRecordingLog`.
+private let llmRunnerLog = os.Logger(subsystem: "io.island.whisper.IslandWhisper",
+                                     category: "LLMRunner")
+
+/// Which product surface an LLM invocation belongs to.
+///
+/// Exists purely for observability: without it every line in the log reads
+/// "some CLI ran", and the question users actually ask — "did the *title*
+/// generation fire for this recording?" — stays unanswerable. `LLMRunner`
+/// itself behaves identically for every case.
+///
+/// `rawValue`s are the log tokens, so they are deliberately short, lowercase,
+/// and stable — grep-ability is the whole point, and renaming one silently
+/// breaks anyone's saved log predicate.
+enum LLMFeature: String {
+    /// Recording title suggestion (rename sheet's "Suggest", and the
+    /// hands-off auto-title in `PostRecordingCoordinator`).
+    case name
+    /// The automatic post-transcription summary (`RecordingSummarizer`).
+    case summary
+    /// The user-triggered "Send to <tool>" post-recording action.
+    case action
+    /// Live AI's periodic in-meeting tick.
+    case liveAI = "live-ai"
+    /// Settings → AI Provider's "Test" button (the `diagnose` path).
+    case settingsTest = "settings-test"
+    /// Caller didn't say. Present so adding the parameter couldn't break any
+    /// existing call site — a line tagged this way means a caller wasn't
+    /// updated, which is itself worth seeing in the log.
+    case unspecified
+}
+
 /// Errors surfaced from the CLI invocation. The Settings UI / rename sheet
 /// renders `errorDescription` directly so users can self-diagnose path /
 /// permission issues without reading logs.
@@ -142,6 +181,11 @@ enum LLMRunner {
     ///
     /// `timeout` defaults to 5 minutes. Pass a smaller value for foreground
     /// callers that block UI (e.g. the Suggest button).
+    ///
+    /// `feature` only tags the unified-log lines this call emits — it changes
+    /// no behaviour. It is the LAST parameter (rather than the first, where it
+    /// would read better) because Swift requires arguments in declaration
+    /// order: appending it keeps every existing call site compiling untouched.
     static func run(tool: LLMTool,
                     prompt: String,
                     transcript: String,
@@ -163,8 +207,13 @@ enum LLMRunner {
                     openAIAPIKey: String? = nil,
                     jsonMode: Bool = false,
                     temperature: Double? = nil,
-                    transport: OpenAITransport? = nil) async throws -> String {
-        guard tool != .none else { throw LLMRunnerError.toolDisabled }
+                    transport: OpenAITransport? = nil,
+                    feature: LLMFeature = .unspecified) async throws -> String {
+        guard tool != .none else {
+            logSetupFailure(feature: feature, tool: tool,
+                            error: LLMRunnerError.toolDisabled)
+            throw LLMRunnerError.toolDisabled
+        }
 
         // OpenAI-compatible HTTP path — handled before any CLI resolution, so
         // `executablePathOverride` / `session` / `extraArgs` (CLI-only) are
@@ -174,7 +223,11 @@ enum LLMRunner {
         if tool == .openaiCompatible {
             guard let baseURL = openAIBaseURL,
                   !baseURL.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
-            else { throw LLMRunnerError.toolDisabled }
+            else {
+                logSetupFailure(feature: feature, tool: tool,
+                                error: LLMRunnerError.toolDisabled)
+                throw LLMRunnerError.toolDisabled
+            }
             let t = transport ?? URLSessionTransport(URLSession.shared)
             return try await runOpenAICompatible(prompt: prompt,
                                                  transcript: transcript,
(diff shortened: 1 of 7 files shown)


urisland:
@coderabbitai full review

Issue #137, opened 2026-07-30
Settings tab strip still overflows: the 560 pt TabView frame is too narrow for the tab bar


urisland:
## Problem

Settings' tab strip still overflows into a `>>` chevron menu, which makes the collapsed trailing tabs look disabled. #131 diagnosed this and folded the Updates tab into General (#133), but deliberately left the window width alone — widening it relayouts every tab and wanted visual verification. This is that follow-up.

`SettingsView` pins its `TabView` to `.frame(width: 560, height: 560)`, and the tab strip needs more than that.

## Measurement

Measured on macOS 26 at the default system font, by building the app with a deliberately over-wide frame and reading each tab item's frame out of the accessibility tree:

| tab | width (pt) |
|---|---|
| Hotkeys | 55 |
| Audio | 55 |
| Models | 55 |
| LLM | 55 |
| Speakers | 60 |
| Meetings | 60 |
| Live AI | 55 |
| Voice Memos | 82 |
| Storage | 55 |
| Updates | 56 |

The ten items occupy a **599 pt** run (they lay out contiguously, so the run is the sum). The strip is given the frame width plus the 40 pt contributed by the `.padding(20)` on the same view — so at `width: 560` it has ~600 pt for a 599 pt run and collapses the last item.

Note this corrects the estimate in #131: that issue put the strip at ~845 pt with 10 tabs and ~758 pt with 9, derived offline from AppKit text metrics. The real figure is 599 pt, so the fix is a much smaller widening than #131 implied — and, contrary to that issue, dropping a single tab *would* have been enough to clear the overflow on its own.

## Fix

Widen to `width: 700`. That fits the 599 pt run with ~70 pt of margin on each side — headroom for roughly one more average-width tab — without adding the dead whitespace a 780–800 pt window would (tab content is left-aligned with inner max widths around 360 pt).

## Related

- #117 proposes `640` for the same bug. By the measurement above 640 does also clear the overflow (680 pt available vs. a 599 pt run), just with ~40 pt less margin. Either value fixes it; the PR for this issue takes the more generous one and documents the constraint in a comment.
- The pending tab-set changes both *shrink* the strip, so they stay comfortably inside 700: #133 renames Hotkeys → General and drops Updates, and `feat/unified-ai-settings-tab` replaces LLM + Live AI with a single AI tab.

PR #194, opened 2026-08-21
feat(settings): replace the tab strip with a sidebar so sections scale past nine

### LINKED ISSUE #177: Settings is a fixed 700x560 separate window with no room for a tenth tab; make it an in-app page instead

urisland:
Closes #177

## What & why

Settings was a `TabView` (AppKit `NSTabView`) pinned to `700x560`, and that
size was load-bearing: AppKit collapses whatever does not fit the tab strip
into a `>>` overflow menu, which reads as a stray "expand button" and makes
the collapsed tabs look disabled — reported as #131, fixed in #137 by sizing
the window to hold the whole strip. #144 then measured the result: nine tabs
in a 599 pt run against ~740 pt of strip room, *"it fits, with no room for a
tenth."* Both remaining moves (widen again, merge tabs) were spent, so new
settings — the Obsidian vault section, and the MCP section on its way — were
being parked inside the Storage tab for want of a slot.

**This PR replaces the tab strip with a `NavigationSplitView` sidebar list.**
Destinations now cost vertical space, which scrolls, instead of horizontal
strip width, which does not. The window loses its hardcoded frame and becomes
resizable.

### I did not implement the change the issue asked for — here is why

#177 proposes making Settings a page inside the main window. I judged the
narrower fix better and implemented that instead. The reasoning, so it can be
argued with:

1. **The overflow ceiling is the actual defect, and a sidebar removes it
   completely.** There is no `NSTabView` in the view tree any more, so the
   `>>` chevron of #131 cannot come back at *any* section count. A tenth and
   eleventh section are now one enum case each. This is also what macOS System
   Settings itself does.

2. **`Cmd+,` and `SettingsLink` keep working for free.** Keeping the `Settings`
   scene means zero changes to `MilaApp.swift`: no hand-rolled
   `CommandGroup(replacing: .appSettings)`, no "the main window is closed, so
   reopen it before you can reach Settings" path, no divergence from the
   platform idiom users have muscle memory for.

3. **The issue's strongest argument actually favours the window.** #177 wants
   to "see a setting and the thing it affects at the same time" — but a
   sidebar destination in the main window *replaces* the recording list and
   detail, so you still cannot see both. Two windows can be tiled side by
   side; one window showing a settings page cannot. Keeping the window is the
   better answer to that goal, not the worse one.

4. **The other two costs #177 lists are fixed here anyway.** "Dense tabs can't
   be given more room" — the window is resizable now. "Cross-references aren't
   navigable" — they are: see below.

What the issue asked for that this does *not* deliver is Settings and a
recording visible simultaneously in one window. If that is genuinely wanted,
it is a separate (and much larger) piece of work; happy to be overruled.

### Changes

- `SettingsView` is a `NavigationSplitView`: sidebar `List` over
  `SettingsTab.allCases`, detail switches on the selection.
- Same nine destinations, **same order** (old left-to-right becomes
  top-to-bottom), same titles, same SF Symbols. Nothing is renamed or moved.
- `.frame(width: 700, height: 560)` → `minWidth: 700, idealWidth: 920,
  minHeight: 440, idealHeight: 620`. The ideal reproduces the old content
  width plus the sidebar; the floor is an explicit judgement call (documented
  as such in the code — the widest *rigid* element in any destination is a
  160 pt label; everything else is `maxWidth:`), not a measurement.
- `.padding(20)` moves from around the old `TabView` to around the detail
  content, so every per-destination layout keeps the exact inset it was
  written against.
- `.toolbar(removing: .sidebarToggle)` — every destination lives in that list,
  so a collapse button would hide all navigation. (macOS 14.0+, matches the
  deployment target.)
- `.navigationTitle(section.title)` on the detail preserves the old behaviour
  where the window title tracks the selected section.
- The `List` selection binding swallows `nil` sets, so Escape or a click on
  empty sidebar space cannot blank the detail pane.
- `SettingsTab` gains `CaseIterable`/`Identifiable` plus `title`,
  `systemImage`, `accessibilityID`. The **type name is deliberately kept** —
  it is the key of the `SettingsNavigation.pendingTab` deep-link contract, its
  raw values are pinned by `AISettingsKeyCompatibilityTests`, and the
  per-destination views are all still `…SettingsTab` (two of them in their own
  files). Renaming would churn all three for no behavioural gain.
- **One cross-reference becomes real navigation**, because it can now: AI
  Features' "No AI provider yet — set one up in Settings → AI Provider" prose
  is now a link that selects AI Provider. This is the first producer for
  `SettingsNavigation.pendingTab`, which until now had none.
- Comments that recorded the width constraint (`SettingsView`,
  `GeneralSettingsTab`'s "didn't justify a tenth tab", `StorageSettingsTab`'s
  note on why Obsidian is parked there) are updated in place rather than
  deleted — the #131/#137/#144 history is why the strip existed and is worth
  keeping.

### Explicitly out of scope

**The Obsidian section stays inside Storage.** So does the MCP section when it
lands. Both were put there *because a tenth tab wouldn't fit*; that reason is
now gone, so promoting them to their own sidebar destinations is possible —
but it is a UX call that deserves its own review, and folding it in here would
bloat a container-only diff. `StorageSettingsTab` carries a comment saying
exactly this.

### Notes for maintainers

- **The `.environmentObject`-on-both-scenes convention in `CLAUDE.md` is
  unchanged.** Settings is still a `Settings` scene, so new app-wide settings
  objects still have to be injected into *both* the main window and the
  Settings scene. `MilaApp.swift` is untouched by this PR. (Had this gone the
  in-app-page route, that rule would have become a trap — one more reason it
  didn't.)
- Every persisted `UserDefaults` key and all Keychain items are untouched;
  this is a container change only, so `AISettingsKeyCompatibilityTests` holds
  as-is.
- **Conflicts with #191** (`feat/remote-model-picker`, still open at time of
  writing) should be small and mechanical: that PR edits `RemoteBackendSection`
  in the middle of the file, this one rewrites the first ~150 lines plus three
  comment blocks. Will rebase once it lands.

## How it was tested

- `make build` — **BUILD SUCCEEDED**, no new warnings.
- `xcodebuild … build-for-testing` — **TEST BUILD SUCCEEDED**; both test
  bundles compile.
- Tests **written but deliberately not run**: a new
  `test_every_settings_destination_is_reachable_from_the_sidebar` in
  `MilaTests/AISettingsKeyCompatibilityTests.swift` pins the destination
  inventory (count, unique titles/identifiers/ids, non-empty title + icon,
  order == raw-value order) and asserts the `settings.section.speakers`
  identifier that `SpeakerSelfHealUITests` hardcodes. The app-hosted test
  targets read the real login keychain and pop a password prompt on every run
  on the maintainer's machine, so they were compiled, not executed. CI should
  run them.
- `SpeakerSelfHealUITests` is updated: its stale "SettingsView is a TabView,
  macOS exposes each tab as a toolbar button" lookup now prefers the
  `settings.section.speakers` identifier and falls back to the `Speakers`
  label. Not run (it is skipped on CI and is an 8-minute local network test).

### ⚠️ Not visually verified — please eyeball

**I did not launch the app**, so no claim here about how it looks is based on
observation. A human should check, on a real build:

1. `Cmd+,` from the main window opens Settings; so does the sidebar footer
   **Settings** row; the App menu ▸ Settings… item works. Only one Settings
   window opens (no duplicate on repeat `Cmd+,`).
2. The sidebar shows all nine rows — General, Audio, Models, AI Provider, AI
   Features, Speakers, Meetings, Voice Memos, Storage — in that order, with
   icons, **and no `>>` chevron and no sidebar-collapse button** in the
   titlebar.
3. Clicking each of the nine renders the right pane, and the window title
   changes to the section name.
4. The window resizes, and shrinking it to the ~700 pt floor does not clip
   anything — worth a specific look at **Models** (remote backend block) and
   **AI Provider** (label/field rows), which are the widest.
5. The first-open size looks sane, i.e. roughly the old window plus a sidebar,
   not a hairline or a full screen.
6. **Audio**: the input-level VU meter still starts when you select Audio and
   stops when you navigate away (it hangs off `.task`/`.onDisappear`, which
   now fire on section switch rather than tab switch).
7. **AI Features** with no provider configured: the orange banner's "Set one
   up in AI Provider" link jumps to AI Provider.
8. Pressing Escape with the sidebar focused does not blank the right pane.

## Checklist

- [x] Builds locally (`make build`); tests compile (`build-for-testing`) but
      were **not run** — app-hosted targets prompt for the login keychain
- [ ] User-facing change is noted in the README changelog — deferred to the
      release that ships it

## Summary by CodeRabbit

* **New Features**
  * Replaced the fixed settings tab bar with a resizable, scrollable sidebar for easier navigation.
  * Added labeled icons and accessibility identifiers to settings destinations.
  * Added direct navigation from the AI provider banner to AI Provider settings.

* **Bug Fixes**
  * Improved deep-link and destination selection behavior when opening settings.

* **Tests**
  * Added coverage to verify settings destinations are reachable, uniquely identified, and consistently ordered.

### DIFF
--- a/Mila/Views/SettingsView.swift
+++ b/Mila/Views/SettingsView.swift
@@ -2,85 +2,183 @@ import SwiftUI
 import AppKit
 import Carbon.HIToolbox
 
-enum SettingsTab: Int, Hashable {
+/// One destination in Settings.
+///
+/// Still called `SettingsTab` even though Settings is no longer a `TabView`
+/// (#177 replaced the strip with a sidebar). The name is deliberately kept:
+/// it is the key of the deep-link contract (`SettingsNavigation.pendingTab`),
+/// its raw values are pinned by `AISettingsKeyCompatibilityTests`, and the
+/// per-destination views are all still named `…SettingsTab`. Renaming the
+/// type would churn all three for no behavioural gain.
+///
+/// `allCases` order IS the sidebar order, top to bottom. Adding a case adds a
+/// row; it costs vertical space in a scrollable list, so — unlike the old tab
+/// strip — there is no width budget to re-measure. See `SettingsView.body`.
+enum SettingsTab: Int, Hashable, CaseIterable, Identifiable {
     /// `aiProvider` + `aiFeatures` replace the former separate `llm` and
     /// `liveAI` tabs. The split is by *how often you touch it*: the provider
     /// is technical and set once (often via a `.milaconfig` import), while the
     /// feature switches are what people come back for. Keeping them on one
     /// screen forced every feature below the fold behind a disclosure arrow.
-    ///
-    /// Net tab count is unchanged from before the AI work (two AI tabs
-    /// replacing two AI tabs), which matters: the strip has to stay inside
-    /// the run measured in `SettingsView.body` below.
     case general, audio, models, aiProvider, aiFeatures, speakers, meetings, voiceMemos, storage
+
+    var id: Int { rawValue }
+
+    var title: String {
+        switch self {
+        case .general: return "General"
+        case .audio: return "Audio"
+        case .models: return "Models"
+        case .aiProvider: return "AI Provider"
+        case .aiFeatures: return "AI Features"
+        case .speakers: return "Speakers"
+        case .meetings: return "Meetings"
+        case .voiceMemos: return "Voice Memos"
+        case .storage: return "Storage"
+        }
+    }
+
+    var systemImage: String {
+        switch self {
+        case .general: return "gearshape"
+        case .audio: return "mic"
+        case .models: return "cube.box"
+        case .aiProvider: return "sparkles"
+        case .aiFeatures: return "wand.and.stars"
+        case .speakers: return "person.2"
+        case .meetings: return "video.fill"
+        case .voiceMemos: return "waveform"
+        case .storage: return "externaldrive"
+        }
+    }
+
+    /// Stable per-row accessibility identifier, so GUI tests can select a
+    /// destination without matching on the (localisable) visible label.
+    var accessibilityID: String {
+        switch self {
+        case .general: return "settings.section.general"
+        case .audio: return "settings.section.audio"
+        case .models: return "settings.section.models"
+        case .aiProvider: return "settings.section.aiProvider"
+        case .aiFeatures: return "settings.section.aiFeatures"
+        case .speakers: return "settings.section.speakers"
+        case .meetings: return "settings.section.meetings"
+        case .voiceMemos: return "settings.section.voiceMemos"
+        case .storage: return "settings.section.storage"
+        }
+    }
 }
 
 @Observable
 final class SettingsNavigation {
     @MainActor static let shared = SettingsNavigation()
+
+    /// Set this to jump Settings to a destination. Consumed (reset to nil) by
+    /// `SettingsView` as soon as it applies. Cross-references inside Settings
+    /// use it to become real navigation instead of prose — see
+    /// `AIFeaturesSettingsTab.notConfiguredBanner`.
     var pendingTab: SettingsTab?
 }
 
-/// Standard `Settings` scene. Opened via `Cmd+,` from the menu bar.
+/// Standard `Settings` scene. Opened via `Cmd+,` from the menu bar, and via
+/// `SettingsLink` in the sidebar footer.
+///
+/// ## Why a sidebar and not a tab strip (#177)
+///
+/// This used to be a `TabView`, and the window width was load-bearing because
+/// of it: AppKit's `NSTabView` collapses whatever does not fit into a `>>`
+/// overflow menu, which reads as a stray "expand button" and makes the
+/// collapsed tabs look disabled. That was reported as #131 and fixed in #137
+/// by pinning the window to 700×560 (740 pt of strip room) to hold the
+/// measured 599 pt run of nine tab items. #144 then concluded the obvious:
+/// nine fits, "with no room for a tenth". Every remaining move — widen again,
+/// or merge tabs — was spent, so new settings were being parked inside the
+/// Storage tab for want of a slot.
+///
+/// A sidebar list removes the ceiling instead of raising it. Destinations cost
+/// vertical space, which scrolls, rather than horizontal strip width, which
+/// does not. There is no `NSTabView`, so the `>>` chevron of #131 cannot come
+/// back at any count, and the window no longer needs a hardcoded size — it is
+/// resizable, which also settles the "dense tab can't be given more room"
+/// complaint in #177.
+///
+/// Settings stays its own `Settings` scene rather than becoming a page inside
+/// the main window: that keeps `Cmd+,`, `SettingsLink` and the App-menu item
+/// working for free (no hand-rolled `.appSettings` command, no "reopen the
+/// main window first" path when it is closed), keeps the macOS-native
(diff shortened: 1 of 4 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

Issue #190, opened 2026-08-20
LLM sandbox is now shared, so one invocation's leftover files are visible to the next


urisland:
Follow-up from the CodeRabbit review of #187 (Major, Security & Privacy), deferred there because a good fix has a real design tension and #187's own change is a strict improvement without it.

## The problem

#187 fixed a leak by pointing every LLM invocation at **one stable working directory** — `~/Library/Application Support/Mila/llm-sandbox/` — because `claude` derives its project identity from a slug of the cwd, so a per-run cwd minted a new `~/.claude/projects/` directory every time (and made those conversations unresumable).

The cost of that stability: the directory is no longer empty at the start of each run. Any CLI that writes a regular file into its cwd leaves it there, visible to every later invocation that inspects its working directory. `LLMRunner`'s own documentation still describes the cwd as providing empty isolation, and that is no longer true.

This is not an external exposure — the directory is under the user's own Application Support and readable only by them, and Mila does not write transcripts there. The concern is narrower: content derived from one recording's LLM call could be discovered by an unrelated later call, including one for a different recording.

## Why it wasn't fixed in #187

The obvious fix — sweep the directory before each run — conflicts with two things #187 deliberately preserves:

1. **The stable project path.** Deleting and recreating the directory is fine, but the path must not change, or the leak returns.
2. **Concurrent invocations.** Live AI ticks run concurrently with one-shot calls (there is now a test pinning that they share the cwd). A sweep at the start of a run could delete a file a concurrent invocation is actively using.

So a correct fix needs a concurrency story, not just a `removeItem`. Options worth weighing:

- **Sweep at app launch only.** Clears anything left by previous sessions; accepts retention between invocations within one session. Simple, no concurrency hazard, partial fix.
- **Per-invocation subdirectory with a stable parent.** Only helps if the CLI's project identity keys on the parent, which it does not — the slug is the full cwd path, so this reintroduces the leak.
- **Reference-count active invocations** and sweep when the count drops to zero. Correct, more machinery.
- **Accept and document.** Update `LLMRunner`'s doc comment to stop claiming empty isolation, and treat the directory as Mila-owned scratch.

## Tests to add

In `MilaTests/LLMRunnerTests.swift`: an artifact created during one invocation must not be discoverable by a later one — written so it does not assume the shared directory starts empty (it won't).

## Note

`LLMRunner`'s doc comment currently overstates the isolation. Whichever option is chosen, that comment needs correcting, since it is what a future reader will trust.

PR #197, opened 2026-08-23
feat(llm): reference the transcript by file path for the Send action

### LINKED ISSUE #179: Reference the transcript by file path instead of inlining it in argv (the .srt already has timestamps the prompt drops)

urisland:
Closes #179

## What & why

The Send-to-LLM action can now hand the CLI the recording's own sidecar **paths** instead of pasting the transcript into the prompt argument. Off by default; inlining is unchanged for everything else.

**The premise checks out, with one correction.** The transcript really is in argv today — `composedPrompt` puts it in the `-p` argument, and `LLMRunnerTests.test_runner_passes_prompt_in_argv_and_closes_stdin` already proves it by echoing the last argv token. And the fidelity loss is real: `TranscriptFormatter.plainText` keeps speaker labels but carries no times, while the `.srt` written next to the same audio has both.

Two things in the issue don't hold as written, and the implementation reflects the corrected version:

- The accessor is `Recording.subtitleFileName` / `RecordingStore.subtitleURL(for:)`, not `srtFileName`.
- **The `.txt` sidecar is *poorer* than what we inline today, not richer.** It holds `Recording.fullText` — the plain join, with no speaker labels — whereas the inlined body goes through `TranscriptFormatter.plainText`. Only the `.srt` is a strict superset. So `.srt` is listed first, `.txt` is offered as the whole-file plain option (and is the only reference that exists for a recording with no segments), and the labels describe the *format* rather than claiming diarization the URL can't attest to.

The issue is also explicit that this is **not** about argv size limits, and that holds: ~65 KB largest transcript against a 1 MB `ARG_MAX`, and #192 already keeps the body out of the log. Nothing here is justified by a size or quoting problem — it's fidelity and agent capability.

### What changed

- **`TranscriptFiles` / `TranscriptDelivery`** in `LLMRunner.swift`. `TranscriptFiles.existing(...)` filters to paths that are on disk *with bytes in them* — a 0-byte sidecar reads to a model as "nothing was said", which is a confidently wrong answer rather than a missing one.
- **`composedPrompt(_:transcript:summary:delivery:)`** stays the single place either wire format is built, so the existing format tests cover both modes. Reference mode emits the issue's shape, keeping the same `---` envelope so a prompt that says "below" still reads correctly. The read instruction covers only the transcripts; the audio line sits below it, unmentioned, so an agent doesn't burn a tool call on an `.m4a` it can't decode.
- **`LLMRunner.effectiveDelivery(_:tool:)`** — the gate, applied once in `run` so the prompt and the log line can't disagree. Two downgrades back to `inline`, both "the reference would be a lie": a tool with no filesystem (`.openaiCompatible`, `.none` — the issue's hard constraint, expressed once on `LLMTool.readsLocalFiles` so the UI and the runner can't drift), and a reference set that came back empty. Both degrade rather than throw: **the worst case of the new mode is exactly today's behaviour.**
- **`LLMSettings.actionTranscriptByPath`** (`llm.action.transcriptByPath`, default off) plus `actionDeliversTranscriptByPath`, the readiness gate per `.claude/rules/feature-gates.md`. The stored preference survives switching to a provider that can't read files, so switching back restores it.
- **Settings → AI Features → Send-to-LLM action** gains one switch, greyed with an explanatory caption when the provider can't read files (same treatment the Live AI row gives its gates). No other UI change.
- **`PostRecordingCoordinator`** derives the paths from the store, so neither sheet that calls `sendToLLM` needed touching. The `byPath` decision is a click-time snapshot (a Settings edit mid-send must not repaint it); the *paths* are resolved after the transcript wait, because the `.srt` is written at the end of the transcription pass and asking earlier would find nothing for every send fired mid-recording.

### Where the transcript file lives: nowhere new

**No temp file is written.** The `.srt`, `.txt` and audio are files `TranscriptionService` and `RecordingStore` already maintain next to each other in the recordings directory — the proposal is to *reference* what is written today, so the feature adds no copy of the user's content anywhere.

That side-steps #190 entirely rather than mitigating it: nothing of the user's is placed in the shared `llm-sandbox` cwd from #187, which every later invocation is also chdir'd into and could therefore read. There is also nothing to clean up and nothing to accumulate.

### Timestamps: reference the `.srt`, don't inline it

Referencing the `.srt` is where the value is — it's the only artifact with times. But inline mode is deliberately **not** changed to carry timestamps:

- SRT roughly triples the payload (sequence number + paired timecodes per cue) for content the model mostly doesn't need.
- Whisper's cues are ~5–10s fragments. That fragmentation reads *worse* to a summarizer than the speaker-collapsed paragraphs `TranscriptFormatter.plainText` already produces.

So the timestamped rendering is available to any agent that wants it, at the cost of a file read, and the default one-shot path keeps the shape that summarizes well. The issue never actually asked for inlined timestamps, and I don't think it should.

### #192's redaction rationale: updated, not relaxed

The tempting conclusion is that once the body leaves argv the prompt argument no longer needs redacting. It's wrong, and the doc comment and a new test both say why: the argument still carries the Live-AI summary, and it carries the sidecar paths — whose filenames Mila derives from the recording title, which Mila generated from the meeting with an LLM. A run log full of `…/Q3 layoffs — legal review.srt` leaks the same class of thing the transcript would, one line at a time, into a store that persists for days.

So `<prompt:Nc>` holds in both modes. What replaces the detail reference mode *did* make un-diagnosable ("did it send paths or the body, and which files?") is a new `delivery=` / `refs=` pair on the start line, naming extensions and never filenames. A silent fallback now reads as `delivery=inline` with a non-zero `transcript=`; a genuine reference run shows `transcript=` bytes that never left the machine.

`LLMRunnerObservabilityTests`' invariant test is unchanged and still passes (it pins inline mode); it gained a cross-reference plus a sibling, `test_redacted_command_never_contains_referenced_sidecar_paths`, asserting the title, the `.srt` path and the summary are all absent from the logged command in reference mode.

## How it was tested

**Built, not run.** `make build` → `** BUILD SUCCEEDED **`, and `xcodebuild … build-for-testing` → `** TEST BUILD SUCCEEDED **` after `make project` picked up the new test file (so every test below compiles).

**I did not execute any test.** App-hosted tests launch a freshly ad-hoc-signed `Mila.app` and several `MilaTests` files read the real login keychain, which pops a password dialog at the machine's owner; `LLMRunnerTests` additionally holds smoke tests that invoke the real `claude` / `cursor-agent` / `gemini` binaries. CI runs them.

Tests written and committed but **not executed**:

- `MilaTests/TranscriptDeliveryTests.swift` (new, 18 tests, pure + temp files only, no process/network): `TranscriptFiles.existing` dropping missing and 0-byte paths; `logToken` / `refsToken` renderings; `readsLocalFiles` per tool; all four `effectiveDelivery` branches; the reference wire format for srt+txt+audio, srt-only, txt-only, audio-only (no read instruction) and summary-above-paths; that reference mode omits the body; that an empty reference set falls back to the inline format; that the default delivery is byte-identical to pre-#179 inline output; and the `LLMSettings` default-off, key round-trip, and provider gate.
- `MilaTests/LLMRunnerObservabilityTests.swift`: `test_redacted_command_never_contains_referenced_sidecar_paths`, `test_delivery_log_tokens_are_stable`.
- `MilaTests/PostRecordingCoordinatorSendTests.swift`: the seam receives `.reference` with all three store-derived paths when enabled; `.inline` by default; `.inline` when the recording is gone; and the auto-title path always inlines. Two existing stub closures gained a `_` for the new seam parameter.

## Left out, deliberately

- **Executable discovery** — #158 owns it and touches this same file.
- **Title generation, the automatic summary, Live AI and the Settings test panel all stay inlined.** The first two run unattended on every recording, so a CLI that declined the read would degrade them silently and permanently; Live AI has no sidecar yet (the recording is still running); the test panel has no recording at all, only its editable sample.
- **The OpenAI-compatible HTTP path** gets no `delivery` forwarded, by construction.
- No `RELEASE_NOTES/` entry — that belongs to whichever version ships this.

## Checklist

- [x] Builds locally (`make build`); tests written and compiled but **not run** — see above
- [ ] User-facing change is noted in the README changelog — not touched; flag if wanted

## Summary by CodeRabbit

* **New Features**
  * Added an option to send transcripts as local file references instead of inline text.
  * Supports subtitle, plain-text transcript, and audio files when available.
  * Added settings guidance and automatic fallback to inline delivery when file access isn’t supported.
  * Transcript references are available for compatible local tools.

* **Bug Fixes**
  * Auto-generated titles continue to use inline transcript delivery.

* **Tests**
  * Added coverage for delivery modes, settings persistence, file availability, fallback behavior, and logging.

### DIFF
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -114,6 +114,111 @@ struct LLMTestResult: Equatable {
     var didLaunch: Bool { exitCode != nil || httpStatus != nil }
 }
 
+/// The on-disk artifacts a recording already has, for the *reference*
+/// transcript delivery (issue #179). Nothing here is produced for the LLM's
+/// benefit: these are the `.srt` / `.txt` / audio files `TranscriptionService`
+/// and `RecordingStore` write next to every completed recording anyway, so
+/// referencing them puts no new copy of the user's content on disk.
+///
+/// Every field is optional because none of them is guaranteed to exist at the
+/// moment a run starts: the `.srt` is written after the transcription pass and
+/// is skipped entirely for a recording with no segments, and the `.txt` is
+/// removed when the transcript comes up empty. Use `existing(...)` rather than
+/// the memberwise init so a path that is not actually on disk can never reach a
+/// prompt — a reference the CLI can't open is worse than no reference at all,
+/// because the model then answers from the prompt's framing instead of from
+/// anything that was said.
+struct TranscriptFiles: Equatable {
+    /// The `.srt` sidecar: SubRip cues carrying `00:00:07,000 --> 00:00:14,000`
+    /// timings and, when diarization ran, a `SPEAKER_00: ` prefix per cue. The
+    /// only artifact Mila writes that has timestamps at all.
+    var subtitles: URL?
+    /// The `.txt` sidecar. This is `Recording.fullText` — the plain join, with
+    /// **no** speaker labels (unlike the inlined body, which goes through
+    /// `TranscriptFormatter.plainText`). Offered as the easy-to-read whole-file
+    /// option, and it is the only reference available for a recording that has
+    /// no segments; `subtitles` is the richer one whenever it exists.
+    var plainText: URL?
+    /// The recording's audio. Referenced, never suggested for reading — an LLM
+    /// CLI can't decode it, but an agentic action ("file this recording in
+    /// $SYSTEM") needs to know where it is.
+    var audio: URL?
+
+    /// True when there is nothing to point the model at, which is the signal
+    /// for `LLMRunner.effectiveDelivery` to fall back to inlining.
+    var isEmpty: Bool { subtitles == nil && plainText == nil && audio == nil }
+
+    /// Which references survived, as a stable `+`-joined log token
+    /// (`srt+txt+audio`). Extensions only — a *filename* is derived from the
+    /// recording's title, which Mila generates from the meeting, so it is
+    /// content and does not belong in the unified log. See `redactedCommand`.
+    var logToken: String {
+        var parts: [String] = []
+        if subtitles != nil { parts.append("srt") }
+        if plainText != nil { parts.append("txt") }
+        if audio != nil { parts.append("audio") }
+        return parts.isEmpty ? "none" : parts.joined(separator: "+")
+    }
+
+    /// Keep only the paths that exist on disk *and* have bytes in them. The
+    /// zero-length check matters: `RecordingStore` deletes an emptied sidecar,
+    /// but an interrupted atomic write can still leave a 0-byte file behind,
+    /// and "read this empty file" reads to a model as "nothing was said".
+    static func existing(subtitles: URL? = nil,
+                         plainText: URL? = nil,
+                         audio: URL? = nil,
+                         fileManager: FileManager = .default) -> TranscriptFiles {
+        func usable(_ url: URL?) -> URL? {
+            guard let url else { return nil }
+            guard let size = (try? fileManager.attributesOfItem(atPath: url.path))?[.size] as? NSNumber,
+                  size.int64Value > 0 else { return nil }
+            return url
+        }
+        return TranscriptFiles(subtitles: usable(subtitles),
+                               plainText: usable(plainText),
+                               audio: usable(audio))
+    }
+}
+
+/// How the transcript reaches the model (issue #179).
+///
+/// `inline` is the historical behaviour and stays the default everywhere: the
+/// transcript body is pasted into the prompt argument. It is the only shape
+/// that works for a provider with no filesystem — notably the
+/// OpenAI-compatible HTTP endpoint.
+///
+/// `reference` swaps the body for the paths of files Mila already wrote. It
+/// exists because inlining is lossy in a way that is invisible from inside the
+/// prompt: `TranscriptFormatter.plainText` keeps speaker labels but drops every
+/// timestamp, while the `.srt` sitting next to the audio has both. It also
+/// turns a one-shot paste into something an agent can re-read, seek within and
+/// quote exact times out of — which is the point of the Send-to-LLM action.
+///
+/// A caller only ever *requests* a delivery; what actually happens is decided
+/// by `LLMRunner.effectiveDelivery(_:tool:)`, which downgrades to `inline` for
+/// a tool that cannot read files and for an empty reference set.
+enum TranscriptDelivery: Equatable {
+    case inline
+    case reference(TranscriptFiles)
+
+    /// `delivery=` log token.
+    var logToken: String {
+        switch self {
+        case .inline:    return "inline"
+        case .reference: return "reference"
+        }
+    }
+
+    /// `refs=` log token. `none` for inline, so the pair always reads
+    /// consistently.
+    var refsToken: String {
+        switch self {
+        case .inline:               return "none"
+        case .reference(let files): return files.logToken
+        }
+    }
+}
+
 /// Spawns the configured `claude` or `cursor-agent` binary with the user's
 /// prompt + transcript and returns whatever the CLI prints to stdout.
 ///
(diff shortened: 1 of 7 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

PR #200, opened 2026-08-24
fix(llm): restore action items on the one-shot summary

### LINKED ISSUE #199: One-shot summary lost action items and can dump raw JSON as the summary
## Problem

After the live one-shot split (#87), the post-recording summarizer only stored a plain-text summary. Two user-visible failures:

1. **No action items from the full transcript.** Live AI used to land a summary + items pair at stop; the one-shot call no longer asked for items, so regenerate/backfill could not refresh them. An empty/glitchy response could also leave the UI without a useful list.
2. **Malformed JSON shown as the summary.** When the model returned a `{"summary":...,"items":[...]}` envelope (out of habit) with unescaped quotes or newlines inside the summary — common in long Hebrew text — `JSONDecoder` failed and the raw `{...}` blob was stored as the summary.

## Intended fix

- Ask the one-shot call for a plain-text summary plus a `###ACTION_ITEMS###` JSON array (summary stays outside JSON).
- Parse sentinel / valid envelope / salvage a broken envelope; never surface a raw blob.
- Restore a richer default summary prompt, migrating users still on the old concise default without clobbering custom prompts.

ValeroK:
Closes #199

## What

The post-recording one-shot summarizer again produces **summary + action items** from the full transcript, and no longer stores a raw `{...}` blob when the model emits broken JSON.

## Why

#87 moved the final summary to a one-shot CLI call that only stored plain text. That dropped the action-items half of the old Live-AI stop tick, and when a model still returned a JSON envelope with unescaped quotes/newlines (typical in long Hebrew summaries) the UI showed the malformed object as the summary.

## How

- `RecordingSummarizer` wraps the user's summary prompt with an output contract: plain-text summary, then `###ACTION_ITEMS###`, then a JSON array. Custom prompts keep working because the extra contract is not persisted.
- Parse order: sentinel → valid envelope → salvage a broken envelope (extract `"summary"` without decoding the object) → prose fallback.
- Fresh items replace the recording's list; an empty list **keeps** existing Live-AI items so a glitchy response cannot wipe them.
- Default summary prompt is richer again (pre-#87 depth). Exact match of the old concise default is migrated; any other stored value is treated as a customisation.

## Tests

`RecordingSummarizerTests` + `LiveAISettingsMigrationTests`: **36/36 passed** (sentinel, envelope, prose, regenerate replace/preserve items, Hebrew salvage, prompt migration).

### DIFF
--- a/Mila/Actions/RecordingSummarizer.swift
+++ b/Mila/Actions/RecordingSummarizer.swift
@@ -373,8 +373,15 @@ final class RecordingSummarizer: ObservableObject {
                 return "Hebrew"
             }
         }()
-        let prompt = liveAISettings.summaryPrompt
+        let basePrompt = liveAISettings.summaryPrompt
             .replacingOccurrences(of: "{{LANGUAGE}}", with: promptLanguageName)
+        // Wrap the user's plain-text summary prompt so the one-shot call
+        // also returns action items. Restores the summary + action-items
+        // pair the pre-#87 inline final Live-AI tick used to produce. See
+        // `promptWithActionItems`: the summary stays PLAIN TEXT and only a
+        // compact item array is JSON, so a long multi-line Hebrew summary
+        // full of quotes can't corrupt the parse.
+        let prompt = Self.promptWithActionItems(base: basePrompt)
         let transcript = recording.fullText
         let timeout = timeoutSeconds
         // OpenAI-compatible config captured up front so the detached call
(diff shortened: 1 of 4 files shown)


ValeroK:
@urisland Rebased onto current `main`. There were no inline review comments to address — CodeRabbit was rate-limited on the original head.

@coderabbitai review

`RecordingSummarizerTests` + `LiveAISettingsMigrationTests` are green. Ready to merge when you are.

urisland:
Thanks for this, @ValeroK — the diagnosis and the fix are both yours, and the
shape of it is right: keeping the summary *outside* the JSON so a long
multi-line Hebrew summary can't break the parse is the detail that makes this
hold up, and the salvage path means the raw blob can't come back.

I've relayed it to an in-repo branch as #227, which closes #199. The relay is
purely mechanical: your commit is cherry-picked unmodified, so you are the
`Author` on it, and my own commit on top carries a `Co-authored-by:` trailer
for you. Nothing of your design was rewritten. I did add three follow-ups —
dictated (`voice_command`) action items are now preserved instead of being
replaced by the model's inferred set, the salvage no longer leaks the items
array when the envelope is pretty-printed, and an unreadable JSON object now
yields no summary rather than the blob, with the parse branch logged — all
described in #227.

The reason for relaying rather than merging here is the usual one in this repo
(#198 → #204, #129 → #164): a fork PR can't get an automated review under the
contributor's own quota — this PR never got one, the review was refused twice
for "review limit reached" on your account — and fork CI needs per-run approval
on every single run. Nothing about your work.

Closing this as superseded by #227. Credit stays with you there.


---

PR #202, opened 2026-08-27
feat(settings): replace the tab strip with a sidebar, and prove it renders in CI

### LINKED ISSUE #177: Settings is a fixed 700x560 separate window with no room for a tenth tab; make it an in-app page instead

urisland:
Closes #177

Second attempt at #177. The first (#194, `60218f1`) was merged on green CI and reverted a day later in `6b64e8b` because Settings rendered completely empty.

## Why the first attempt was empty

Diagnosed, not guessed. I re-applied `60218f1` on a throwaway branch with a capture-only UI test that dumps the Settings window's accessibility tree from CI. Inside the 900×472 Settings window:

```
Window, {{62,134},{900,472}}, identifier: 'com_apple_SwiftUI_Settings_window', title: 'General'
  Group,      {{62,134},{900,472}}
    SplitGroup, {{62,-2671},{900,6115}}, identifier: '…SidebarNavigationSplitView'
      …
        StaticText, {{92,-2621},{72,16}}, identifier: 'settings.section.general'
        StaticText, {{94,-2589},{57,16}}, identifier: 'settings.section.audio'
        …
        CheckBox,   {{230,3314},{216,17}}, identifier: 'updates.autoCheck.toggle'
```

The split view laid out **6115 pt tall inside a 472 pt window**, vertically centred, so it began 2671 pt *above* the window. Every sidebar row landed at y ≈ −2600 and General's controls at y ≈ +3314 — thousands of points outside the visible content area. Nothing was missing; everything was off-screen. The window title still tracked the section, which is precisely why the window read as "chrome, but no content", and why the post-mortem saw only the traffic lights and the title.

**The mechanism.** The old `.frame(width: 700, height: 560)` was a *rigid* frame, and it had been the only thing pinning the content to the window. #194 replaced it with `.frame(minWidth:idealWidth:minHeight:idealHeight:)` — no `max` in either axis. A flexible frame reports `clamp(childSize, min, max)`, so with no upper bound it adopted the split view's content-driven 6115 pt instead of the 472 pt the window offered. This is the same failure mode `DetailLayoutUITests` was originally written for, one window over.

## What & why

The design reasoning from #194 is unchanged and still holds. What changes is the layout mechanism.

A plain `HStack` of two **bounded** columns — a fixed-width `List` sidebar and a detail column that can never report more height than it is proposed. `SettingsTab.providesOwnScrollView` records which destinations already root a `ScrollView`; the four that don't (General, Audio, Speakers, Voice Memos) get wrapped in one, so the invariant "every child of the window-sized frame is bounded" holds for all nine. Double-wrapping the other five would nest two vertical scrollers, which fight.

The window frame declares min/ideal/max in **both** axes, and `windowResizability(.contentSize)` on the `Settings` scene maps those onto the window's own resize limits:

| | width | height |
|---|---|---|
| floor | 788 (188 sidebar + 560 content + 2×20 inset) | 600 (the old window's height) |
| default | 928 (the old window's 700 pt of content) | 700 |
| ceiling | none | none |

Everything else carries over from #194:

* **The ceiling is removed, not raised.** Destinations cost scrollable vertical space instead of horizontal strip width, so #144's "nine fits, with no room for a tenth" stops being a constraint. Adding a case plus a `switch` arm is the whole job.
* **#131 cannot regress.** There is no `NSTabView` in the tree, so AppKit's `>>` overflow chevron has nowhere to come from at any destination count.
* **`Cmd+,` and `SettingsLink` keep working with zero code**, because Settings stays its own `Settings` scene rather than becoming a page in the main window. That also keeps the `.environmentObject`-on-both-scenes convention in `CLAUDE.md` true as written, and — being a real window — it can sit side by side with the main window, which an in-app page taking over the content area cannot.
* **Container change only.** Section order, titles, icons, every persisted `UserDefaults` key and the Keychain items are untouched (`AISettingsKeyCompatibilityTests` still pins them).

One behavioural addition, because a cross-reference inside one window can now actually navigate: AI Features' "no AI provider yet" banner becomes a link to AI Provider. It is the first producer of `SettingsNavigation.pendingTab`, which until now had consumers but no caller.

## How it was tested

`make build` and `xcodebuild build-for-testing` both succeed locally. **The app was not launched or screenshotted on the author's machine** — visual confirmation comes from CI, which is the point of the rest of this section.

#194 passed every check, `UI layout tests` included, because nothing in CI ever looked inside the Settings window. Three tests in `DetailLayoutUITests` — the class `.github/workflows/llm-live-ai-e2e.yml` runs — now do:

1. `test_settings_sidebar_lists_every_section` — all nine rows exist, are hittable, **and their frames are inside the window's frame**. That last clause is what fails on #194's build, where `exists` was true for all nine rows at y ≈ −2600.
2. `test_settings_each_section_renders_its_content_and_retitles_the_window` — selects each of the nine in turn and asserts a control unique to that destination appears inside the window bounds, that the window title becomes the section name, and that the *previous* destination's control has disappeared. That last assertion is also the VU-meter check the post-mortem asked for: it proves the old destination is torn down on selection change, which is what makes `AudioSettingsTab`'s `.onDisappear` stop the monitor and its `.task` restart it (the `TabView` used to give that for free; `.id(selectedTab)` on the detail pane guarantees it now).
3. `test_settings_content_is_not_clipped_at_the_window_floor` — drags the window to its size floor and asserts Models and AI Provider (the widest) are not clipped. It skips rather than fails if the runner doesn't land the corner drag, since a missed drag says nothing about clipping.

Each writes a PNG of the Settings window into `$MILA_UI_SCREENSHOTS_DIR`, so `scripts/llm-verify-screenshots.py` judges them too. That script gains a `settings` profile that fails an empty sidebar, a blank pane, or a reappearing tab strip; the floor shots are attached to the test result but kept out of the vision input, since a window at its minimum size is exactly what a "does this look right?" judge would flag.

`Mila/Views/SettingsView.swift` and `Mila/Views/ContentView.swift` join the workflow's paths filter — without them a future change to either container could not trigger the job that screenshots it.

Also added: an app-side inventory test asserting every destination has a non-empty title, icon and namespaced identifier, no duplicates, and `allCases` order equal to raw-value order; plus a tripwire on the two window-width constants so the 560 pt content floor can't be eaten by a wider sidebar.

## What the CI run on this head actually showed

`UI layout tests (+ optional vision)` green in 4m47s. The Settings window came up 900×632 on the runner (its ideal is 928×700; the runner's screen clamps it, and 900 is comfortably above the 788 floor). I downloaded the `ui-screenshots` artifact and looked at all ten Settings shots: sidebar populated with the nine rows in order, the selected row highlighted, each destination's real controls rendered inside the window, and the title bar naming the section. Storage — the tallest — shows all four of its cards with nothing cut off; Models shows the backend picker plus both model rows with their download bars out to x ≈ 880; AI Features shows the new "Set one up in AI Provider" link.

Two caveats, both visible in the artifact:

* **`test_settings_content_is_not_clipped_at_the_window_floor` skipped.** The corner drag did not move the window (`(62,54,900,632) -> (62,54,900,632)`), so the floor was never exercised. The bounds assertions at the default size still ran on all nine destinations. If someone wants the floor genuinely covered, the honest fix is an app-side test seam that sets the window to its declared minimum, rather than a synthetic drag.
* **The Claude-vision step skipped** — `ANTHROPIC_API_KEY` is not available to this run, so the new `settings` profile in `llm-verify-screenshots.py` has not executed against a real image. Its classifier mapping is unit-checked against the actual artifact filenames, and the profile text is the only untested part.
* Several shots contain macOS's own "Mila would like to access the Microphone" sheet floating over the pane. That is the runner's TCC prompt, not a layout defect.

## Checklist

- [x] Builds locally (`make build`) and the test target builds (`xcodebuild build-for-testing`)
- [ ] User-facing change is noted in the README changelog — leaving to the release that ships it

## Summary by CodeRabbit

* **New Features**
  * Redesigned Settings with a sidebar for navigating sections and a dedicated detail pane.
  * Added clearer Settings window titles and improved content sizing and resizing behavior.
  * Added a “Set one up in AI Provider” action to the “No AI provider” message.
  * Improved support for navigating and displaying all Settings sections without tab overflow.

* **Bug Fixes**
  * Prevented Settings content from being clipped when the window is resized smaller.

### DIFF
--- a/.github/workflows/llm-live-ai-e2e.yml
+++ b/.github/workflows/llm-live-ai-e2e.yml
@@ -108,6 +108,14 @@ jobs:
               - 'scripts/fixtures/**'
               - 'Mila/Models/LiveAISettings.swift'
               - 'Mila/Actions/LiveAISession.swift'
+              # DetailLayoutUITests screenshots the main window AND the
+              # Settings window, so a change to either container has to be
+              # able to trigger this job. #194 shipped an empty Settings
+              # window past a green CI precisely because nothing here looked
+              # at Settings; leaving the view out of the filter would put the
+              # new coverage back out of reach.
+              - 'Mila/Views/SettingsView.swift'
+              - 'Mila/Views/ContentView.swift'
               - 'MilaUITests/DetailLayoutUITests.swift'
               - '.github/workflows/llm-live-ai-e2e.yml'
       - name: Install xcodegen
--- a/.github/workflows/llm-live-ai-e2e.yml
+++ b/.github/workflows/llm-live-ai-e2e.yml
@@ -109,13 +109,29 @@ jobs:
               - 'Mila/Models/LiveAISettings.swift'
               - 'Mila/Actions/LiveAISession.swift'
               # DetailLayoutUITests screenshots the main window AND the
-              # Settings window, so a change to either container has to be
-              # able to trigger this job. #194 shipped an empty Settings
-              # window past a green CI precisely because nothing here looked
-              # at Settings; leaving the view out of the filter would put the
-              # new coverage back out of reach.
-              - 'Mila/Views/SettingsView.swift'
-              - 'Mila/Views/ContentView.swift'
+              # Settings window, so a change to any view this job renders —
+              # or to any accessibility identifier or launch seam it queries
+              # — has to be able to trigger it.
+              #
+              # A glob rather than a hand-kept list of files, deliberately.
+              # #194 shipped an empty Settings window past a green CI
+              # precisely because nothing here looked at Settings, and an
+              # incomplete filter reproduces that failure exactly: the job
+              # never runs, so the coverage exists on paper only. Enumerating
+              # files re-opens that hole every time a view moves, gets split,
+              # or a new destination lands in a new file — three ways for
+              # this list to rot silently. Naming the whole directory cannot
+              # rot. The identifiers these tests match on are spread across
+              # SidebarView (sidebar.settings.link, sidebar.folder.default,
+              # sidebar.recording.*), RecordingDetailView
+              # (detail.title.label), SettingsView (settings.section.* plus
+              # most destination probes), StorageSettingsTab and
+              # VoiceMemosSettingsTab (their probes) — all under Mila/Views.
+              - 'Mila/Views/**'
+              # The --ui-test-clean-store / --ui-test-seed-recording launch
+              # seams and the Settings scene's windowResizability live here,
+              # and the tests' launch-state contract depends on both.
+              - 'Mila/App/MilaApp.swift'
               - 'MilaUITests/DetailLayoutUITests.swift'
               - '.github/workflows/llm-live-ai-e2e.yml'
       - name: Install xcodegen
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -783,6 +783,13 @@ struct MilaApp: App {
                 .environmentObject(obsidianVaultSettings)
                 .environmentObject(obsidianExporter)
         }
+        // Settings used to be pinned to 700×560 by a rigid `.frame` inside
+        // `SettingsView`; since #177 it is a resizable sidebar window whose
+        // content declares min/ideal/max instead. `.contentSize` is what
+        // turns those declarations into the window's own resize limits —
+        // without it the window would not honour the minimum and could be
+        // dragged narrower than the destinations can lay out in.
+        .windowResizability(.contentSize)
     }
 
     /// Build a diagnostic zip and let the user pick where to save it.
--- a/Mila/Views/SettingsView.swift
+++ b/Mila/Views/SettingsView.swift
@@ -2,85 +2,321 @@ import SwiftUI
 import AppKit
 import Carbon.HIToolbox
 
-enum SettingsTab: Int, Hashable {
+/// One destination in Settings.
+///
+/// Still called `SettingsTab` even though Settings is no longer a `TabView`
+/// (#177 replaced the strip with a sidebar). The name is deliberately kept:
+/// it is the key of the deep-link contract (`SettingsNavigation.pendingTab`),
+/// its raw values are pinned by `AISettingsKeyCompatibilityTests`, and the
+/// per-destination views are all still named `…SettingsTab`. Renaming the type
+/// would churn all three for no behavioural gain.
+///
+/// `allCases` order IS the sidebar order, top to bottom. Adding a case adds a
+/// row; it costs vertical space in a scrollable list, so — unlike the old tab
+/// strip — there is no width budget to re-measure.
+enum SettingsTab: Int, Hashable, CaseIterable, Identifiable {
     /// `aiProvider` + `aiFeatures` replace the former separate `llm` and
     /// `liveAI` tabs. The split is by *how often you touch it*: the provider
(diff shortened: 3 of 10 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

PR #204, opened 2026-08-27
feat(speakers): cross-recording voice recognition, opt-in and off by default (relay of #198)

### LINKED ISSUE #24: Speaker recognition via learned voice profiles for single-mic in-person meetings
**Problem / use case**
Diarization separates speakers but labels them generically (Speaker 1/2). For in-person meetings recorded from a single mic with multiple people, there's no way to attach real identities to speakers.

**Proposed solution**
Let Mila learn a background voice model for the user (and optionally other known people), then perform speaker recognition - matching diarized segments to known identities - for single-mic recordings with multiple speakers.

**Alternatives considered**
Manual relabeling of speakers after transcription.

**Additional context**
Builds on the existing pyannote.audio diarization pipeline; enrollment would store per-person voice embeddings locally.

urisland:
Closes #24

## Credit: this feature is @noamleibovitch's work

**Cross-recording voice recognition was designed and written by
@noamleibovitch. This PR relays it — it supersedes #198, which is where the
work actually happened.** Six of his commits are cherry-picked unmodified, so
he is the `Author` on every one of them; everything I added sits in a single
commit on top, and it is a feature gate plus tests, not the feature.

> **On the `Co-authored-by` trailer.** My commit carries
> `Co-authored-by: noamleibovitch <noam@example.com>`, and that address will
> not link to his GitHub account. It is not a transcription error on my part:
> `gh api repos/island-io/mila/pulls/198/commits --jq '.[].commit.author'`
> returns `noam@example.com` for all nine of his commits — a placeholder in
> his local git config — and it is the same address recorded as `Author` on
> the cherry-picked commits here. I used it verbatim rather than substituting
> a guess. The `@noamleibovitch` handle above is therefore the attribution
> that actually resolves; please read the credit from it, not from the
> trailer.

The two commits in #198 that added and then removed an unrelated `/review`
skill cancelled out exactly (105 lines added, 105 removed) and were dropped.
Nothing else was rewritten.

It is relayed rather than merged for the usual reason (the #129 → #164
precedent): a fork PR can't get a CodeRabbit review under the contributor's
own quota, and its CI needs per-run approval. Please close #198 as superseded
by this once it's green.

## What & why

Cross-recording voice recognition. When you name a speaker, their 256-dim
voice embedding is persisted to `speaker-profiles.json`; new recordings seed
the live diarizer's pool with known profiles, so a returning speaker is
identified without being renamed again. Same design as #198:

- **`SpeakerProfileStore`** — persistence, centroid merging, cosine-similarity
  matching, CRUD.
- **`LiveSpeakerDiarizer.seedPool(with:)` / `currentProfiles()`** — pre-populate
  the pool with known speakers; snapshot it for persistence.
- **`RecordingStore.onSpeakerNamed`** — hook that fires when the user names a
  speaker, so the profile is saved from the existing `SpeakerNamePicker` flow.
- **`MilaApp`** — seeds at recording start, auto-assigns recognised names on
  stop (only for speakers who actually spoke, verified against
  `diarizer.intervals`).

### The one behavioural change: it is opt-in and off by default

Maintainer's call. As it stood, the feature persisted embeddings and seeded
recognition on its own. It now does nothing at all until the user turns it on.

**New setting** — `VoiceRecognitionSettings`, key
`speakers.voiceRecognition.enabled`, default off. Follows the house shape
(`DiarizationSettings`, `VoiceMemosSettings`): `@Published` writing through to
a namespaced key with an injectable `defaults` suite. Instantiated in
`MilaApp.init()` as a `@StateObject` and injected via `.environmentObject()` on
both the main window and the Settings scene.

**The gate is `isConfigured`, not `isEnabled`** (per
`.claude/rules/feature-gates.md`): `isEnabled && diarizationReady()`, where
readiness is injected as a read of `DiarizationSettings.isConfigured`. Voice
recognition owns no embedding pipeline of its own — every centroid it stores or
matches comes out of the live diarizer's pool, which is only populated while
diarization is enabled *and* its local Python pipeline is verified. Acting on
the toggle alone would mean seeding an empty pool and writing profiles that
could never be recognised. An unwired `diarizationReady` returns false, so a
forgotten injection fails closed.

**While off, nothing is written and nothing is read** — the promise is "never
stored", not "stored but ignored". Every step of the path is closed:

| Path | While off |
|---|---|
| Naming a speaker (`RecordingStore.onSpeakerNamed` → `updateProfile`) | Gated at the call site in `MilaApp`, so the centroid is never even read out of the diarizer pool; `updateProfile` refuses again. |
| Seeding a new recording (`seedPool` ← `seedEntries`) | Gated at the call site; `seedEntries()` returns `[]`; and `profiles` is empty anyway. |
| Auto-assign on stop (`autoAssignRecognisedSpeakers`) | Returns immediately. |
| Reading `speaker-profiles.json` | Not parsed at all — `load()` only runs on opt-in, so no embedding reaches memory. |
| Writing `speaker-profiles.json` | `save()` refuses. |

Also fixed on the way through: a `try!` in the store's production init (a
missing Application Support directory would have crashed at launch — now the
non-throwing lookup + temp fallback `SpeakerDirectory` uses, including the
`--ui-test-clean-store` branch), the store's own `UserDefaults.standard`
access (moved to the injectable settings object so tests never touch
`.standard`), and an unused `[weak store, weak diarSettings]` capture that was
producing two build warnings.

### Judgement call 1: what happens to already-stored profiles on opt-out

**Decided: keep them on disk, drop them from memory, and keep an explicit
"Delete Voice Profiles" button visible — including while the feature is off.**

Deleting on toggle-off is destructive and surprising. A user switching off to
stop *further* learning, or fat-fingering the switch, would silently lose every
named voice with no undo, and no settings toggle should behave like that.
Leaving them silently is the other failure: voice biometrics staying at rest
after somebody opted out, with nothing in the UI admitting it.

The button resolves both. Opting out takes effect completely and immediately —
nothing loaded, nothing seeded, nothing written — and separately, whenever a
profiles file exists, Settings says so and offers to delete it. Deletion is
therefore an explicit, informed decision rather than a side effect, it works
*while off* (which is exactly when somebody wants their voice data gone), and
it removes the file rather than writing an empty array over it, so nothing is
left at rest. Re-enabling picks up where the user left off.

One trap this created and closed: while off, `profiles` is empty, so any
mutation reaching `save()` would have overwritten the stored file with `[]` and
destroyed the data we just promised to keep. `save()` refuses while off, and
there's a test for it.

### Judgement call 2: the caption

Verbatim, shown in both states — it is the disclosure you need *before*
deciding, not a teaser that expands once you've already switched it on:

> Naming a speaker also saves a voice fingerprint for them: 256 numbers
> describing how the voice sounds, written to speaker-profiles.json in Mila's
> Application Support folder. The audio is not part of it and a fingerprint
> can't be played back — but it is enough to pick the same person out of a
> later recording, which is the point of it. Nothing is uploaded: the
> fingerprints stay on this Mac and are left out of diagnostic reports. They
> describe everyone whose name you fill in, not only you, so turn this on only
> where that is yours to decide.

Plain about what is stored (numbers derived from the voice, not audio), where
(named file, named folder), that it never leaves the machine, and that it
covers other people. No "telemetry", no "improve your experience". The
diagnostic-report claim was checked, not assumed: `DiagnosticReporter` copies
only `recordings.json`, `folders.json`, a prefix-whitelisted `settings.json`,
logs and crash reports — `speaker-profiles.json` is never enumerated.

### UI

Settings → **Speakers**, below the known-speakers list, in that tab's flat
divider-separated style. Off-state footprint is one toggle and one caption.
Two things appear only when they apply: the delete row (whenever a profiles
file exists, on or off) and a line explaining that an enabled toggle is waiting
on diarization when that isn't ready yet — otherwise the feature would look
broken while silently doing nothing.

## How it was tested

- `make build` — **BUILD SUCCEEDED**; `build-for-testing` — **TEST BUILD
  SUCCEEDED**. No new warnings (two pre-existing ones removed).
- Tests were **not run locally** (they're app-hosted and pop a keychain
  prompt) — CI's `unit-and-ui-tests` ran them and passed. New
  `MilaTests/VoiceRecognitionGateTests.swift`, 16 tests aimed squarely at the
  off state, since a feature gate with no test for its off-state is how these
  regress:
  - off by default on a fresh install; the persisted key is
    `speakers.voiceRecognition.enabled` and is *unset* (not a stored `false`);
    opt-in survives relaunch
  - `isConfigured` needs both halves — enabled-but-not-ready and
    ready-but-not-enabled are both false; unwired readiness fails closed
  - while off: `updateProfile` writes nothing, and no file is created; naming a
    speaker through the real `RecordingStore.onSpeakerNamed` path persists no
    embedding (while the per-recording label still works); enabled-but-not-ready
    also writes nothing
  - while off: an existing profiles file is not parsed (`profiles` empty,
    `hasStoredProfilesOnDisk` true); `seedEntries()` and `match()` return
    nothing; seeding the diarizer with `[]` leaves its pool empty
  - enable → disable keeps the file but drops memory; re-enable restores;
    an off-state mutation cannot clobber the file
  - `deleteAllProfiles()` works while off and removes the file; after deleting,
    re-enabling restores nothing
- `MilaTests/RecognisedSpeakerMergeTests.swift`, 4 tests on the merge-count
  invariant (above).
- `MilaTests/SpeakerProfileStoreTests.swift` (9 tests from #198) updated for the
  new init and moved onto per-test `UserDefaults` suites — never `.standard`.
- Not verified: no manual end-to-end run with real audio (two people, two
  recordings, returning-speaker recognition). Worth a pass before release.

## Fixed: the double merge (CodeRabbit finding, Major)

`MilaApp.autoAssignRecognisedSpeakers` called `updateProfile` directly *and*
called `setSpeakerName`, which synchronously fires `onSpeakerNamed` — wired in
`MilaApp.init` to that same `updateProfile`. Every recognised speaker was
merged twice per recording.

The centroid *value* survives a double merge (a weighted average of x with x is
x), which is why it went unnoticed. What doubles is `sampleCount`, and since
the merge weights by sample count, each recording makes the profile more rigid
until it stops adapting — recognition decays with nothing visibly failing.
That is worse than a crash, so it is fixed here rather than deferred.

The fix is to delete the direct `updateProfile` call: `setSpeakerName` is the
single persistence trigger. That also makes the path idempotent — a duplicated
stop event finds the name already set, so `setSpeakerName` returns without
firing the hook, and nothing is re-merged. A direct call alongside it is
exactly what would destroy that, so the call site now carries a comment saying
so.

New `MilaTests/RecognisedSpeakerMergeTests.swift` (4 tests) names the
invariant — *one recognition merges exactly once* — against real
`RecordingStore` + real `SpeakerProfileStore` through the real hook: 4 stored +
2 observed samples must give 6, not 8, with the resulting centroid as a second
independent witness (2/3, 1/3 rather than 0.5, 0.5); re-running for the same
recording merges nothing further; a *later* recording does merge again (the
invariant is per recognition, not once ever); and one test reproduces the
removed shape and asserts it yields 8, so the `6` above is provably
discriminating rather than vacuous.

## Also fixed — seeded sample-count inflation

CodeRabbit's second Major finding. Seeded pool entries start at
`min(sampleCount, 3)`, and `currentProfiles()` handed that total back as if it
were all new samples, so a profile with three prior samples and one confident
match merged with the weight of four — `3 + 4 = 7` persisted where `4` is
correct.

This was initially left for a follow-up issue as a design question. That was
wrong, and the reason is worth recording: a correct delta needs a count **and**
a centroid to merge against. Persisting `observedCount` while handing back the
pool's matching centroid gives the right count with the wrong centroid, because
that centroid already contains the seeded profile's own contribution —
`updateProfile` would merge the stored centroid against something that already
contains it. So an observed-only centroid is not a separable redesign, it is
required for the count fix to mean anything. It came to ~15 lines.

`SpeakerProfile` now carries two pairs with distinct jobs:

- `centroid` / `sampleCount` — **matching**. Seeded entries still start at the
  persisted centroid with capped weight, unchanged, so seeding still anchors
  against one noisy utterance.
- `observedCentroid` / `observedCount` — a running mean over **only this
  recording's** embeddings, empty for a seeded entry until it first matches.
  This is what `currentProfiles()` reports and the only thing ever persisted.

For a speaker minted during the recording the two pairs are identical, so the
common path is untouched. A seeded speaker who never spoke reports count 0,
which `updateProfile` already refuses, so no spurious write is possible.

`RecognisedSpeakerMergeTests` is now 11 tests, driving the real `seedPool` →
real `assign` fold → real `currentProfiles()` → real `SpeakerProfileStore`:
3 stored + 1 confident match persists **4, not 7**, with a negative control
asserting the old seeded-inclusive path lands on 7; 40 stored + 1 match → 41;
three utterances → 43; a seeded speaker who never spoke persists nothing; and
seeded weight still anchors matching and forks no speaker, which guards against
"fixed the count, broke the seeding".

One sub-claim in that finding was **not** acted on: a dimension guard on the
fold, for a seeded profile from a different embedding model. It is unreachable —
`cosineSimilarity` returns 0 on a count mismatch, so such an entry can never be
the best match above threshold. The reasoning is documented at the fold instead
of adding dead code.

What genuinely remains is narrower and is filed separately: the seed anchor
weight (`min(sampleCount, 3)`) was picked without measurement, and now that
seeded weight never reaches persistence it can be tuned in isolation.

## Also fixed — three further Major findings from later review rounds

Two more review rounds after the above each found a caller still on the old
semantics, plus one unsound assumption. All are fixed.

### Cross-recording label bleed

Naming a speaker in an **older** recording could persist a **different**
recording's voice into that profile. Diarizer labels (`SPEAKER_00`, …) are
reused after `reset()`, and the `onSpeakerNamed` hook was originally
`{ _, rawID, name in }` — it ignored the `recordingID` that
`RecordingStore.setSpeakerName` already provides, and resolved `rawID` against
whatever was currently in the live pool.

New `ObservedVoiceSnapshots` keys observations by recording id, snapshotted
before anything can trigger a name. The hook resolves through the snapshot and
never touches the live pool. A recording with no snapshot resolves to `nil` and
persists nothing — honest rather than a guess, since its pool is gone. Bounded
to 8 recordings, and dropped entirely on opt-out.

### The auto-naming gate had no confidence check

`assign` deliberately attaches borderline matches and sub-second utterances to
the nearest existing speaker *without* folding them into the centroid. Those
attachments still produce intervals, and the gate tested only intervals — so a
stored profile's name could be stamped onto a speaker that never confidently
matched it: a misattributed transcript, with the wrong person's name against
their words. The gate now requires `observedCount > 0`, which is precisely the
"confidently observed this recording" signal.

Both of these were **pre-existing**, not introduced by the delta fix above:
`c377ede` (the fork's final commit) already had the intervals-only gate and the
`rawID`-only hook. The delta fix had, however, already halved the second one's
blast radius — `observedCount == 0` was failing `updateProfile`'s guard, so the
profile was no longer corrupted, leaving only the wrong label.

The interval check was then **removed** rather than kept alongside, because the
rationale for keeping it did not survive checking. `intervals` holds diarizer
*utterance* intervals, appended in `process` before anything is matched to
transcript segments (`applySpeakerLabels` does that later) — so it never
evidenced transcript presence. And it was redundant: `observedCount` is only
incremented by `assign`, whose sole production caller is `process`, which
appends an interval for exactly the id `assign` returned, while `reset()` clears
pool and intervals together so they cannot drift. A dead check defended by a
false rationale is worse than no check.

### The finished recording was identified by list order

The snapshot was keyed off `store.recordings.first`. That is unsound, though not
for the reason it first appears: `createdAt` is set in the *finalize* path, so it
is effectively stop time, and date ordering would mostly work. The actual problem
is that `add()` does `recordings.insert(_, at: 0)` and never re-sorts — ordering
is not consulted at all, so **any** recording added afterwards becomes `first`
regardless of its date, and `.onChange(of: isRecording)` is delivered a runloop
tick later, which leaves that window open (the Voice Memos importer adds from an
async flow). A test pins this directly: an *older* recording added later is still
`first`.

The id is now carried, not inferred, via
`QuickActionsController.onRecordingFinalized`. That site was chosen over
publishing a last-finalised id because it is also the only moment at which
reading the live diarizer is sound — after `awaitPending()` so intervals are
final, after `store.update(updated)` so assigned names are not clobbered, and
inside the `isFinalizingRecording` window so no new recording can have `reset()`
the pool. The deferred `.onChange` had none of those guarantees.

Fail-closed behaviour is pinned by `test_a_wrong_id_fails_closed`, so a future
"fall back to the newest recording" cannot quietly undo it.

### Two things fixed in passing

- **`onEnabledChange` became `addEnabledObserver`.** A single assignable slot
  meant a second registrant silently unhooked the first, so only the
  last-constructed object would ever hear an opt-out. With
  `ObservedVoiceSnapshots` now needing that signal too, this would have silently
  broken the store's unload. A test pins both registrants firing.
- **The assign loop moved into `RecognisedSpeakerAssigner`** with explicit
  dependencies. The gate had only ever been testable through a replica, which is
  a poor way to pin a security-relevant invariant; it is now driven directly and
  the replica is deleted (its negative control kept).

61 tests now cover this feature across six files, with negative controls on each
invariant: the live-pool lookup provably leaks the stranger, the intervals-only
gate provably names the borderline speaker, and the ordering test provably picks
the wrong recording.

### Known, not addressed here

- Naming a speaker on the **live** transcript still creates no profile —
  `LiveAIRecordingView` writes `transcriber.speakerNames[raw]` directly and never
  reaches `setSpeakerName`. Pre-existing and unchanged by this PR; filed as #209.
- The seed anchor cap (`min(sampleCount, 3)`) is unmeasured; filed as #206.
- No manual end-to-end run with real audio and two speakers has been done.

## Checklist

- [x] Builds locally (`make build`); tests written, left for CI to run

## Summary by CodeRabbit

- **New Features**
  - Added optional cross-recording voice recognition for previously named speakers.
  - Added speaker profile management, including renaming, merging, and deletion.
  - Added voice recognition settings with opt-in controls and readiness status.

- **Privacy**
  - Voice recognition is off by default and profiles are stored locally.
  - Deleting a profile immediately stops future matching.
  - Voice observations remain associated with their recordings.

- **Bug Fixes**
  - Improved speaker naming when recordings are finalized.
  - Prevented duplicate profile updates and inaccurate sample counts.
  - Improved handling of invalid or incompatible voice-profile data.

### DIFF
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -148,6 +148,20 @@ final class QuickActionsController: ObservableObject {
     /// `stopRecording`.
     @Published var isFinalizingRecording: Bool = false
 
+    /// Late-bound by MilaApp. Called once per finished recording, **with
+    /// that recording's id**, from inside the finalize drain: after
+    /// `liveDiarizer.awaitPending()` (so `intervals` are final), after the
+    /// recording's final state is written to the store, and before
+    /// `liveDiarizer.stop()` — i.e. inside the window where the live
+    /// diarizer's pool is still guaranteed to belong to this recording.
+    ///
+    /// Cross-recording voice recognition hangs off this. It has to be told
+    /// the id: observing `isRecording` flipping false carries none, and
+    /// recovering one from `store.recordings.first` is unreliable because
+    /// `add` inserts at index 0 with no re-sort, so any recording added in
+    /// between (a Voice Memos import, say) becomes `first`.
+    var onRecordingFinalized: ((UUID) -> Void)?
+
     /// Late-bound by MilaApp. When the live-transcript path saves a
     /// recording directly (skipping `transcription.enqueue` because the
     /// VAD path already produced segments), TranscriptionService's
@@ -837,6 +851,22 @@ final class QuickActionsController: ObservableObject {
         updated.actionItems = finalItems.isEmpty ? nil : finalItems
         updated.status = liveTranscriptIsAuthoritative ? .completed : .pending
         store.update(updated)
+
+        // Cross-recording voice recognition gets its one shot here, with the
+        // id of the recording that actually just finished. This is the only
+        // point where reading the live diarizer for *this* recording is
+        // sound: `awaitPending()` above means `intervals` are final, the
+        // `isFinalizingRecording` window below hasn't closed so no new
+        // recording can have `reset()` the pool, and `updated` is already
+        // written so the speaker names it assigns won't be clobbered by the
+        // snapshot above.
+        //
+        // Deliberately a callback with the id rather than something MilaApp
+        // observes: it used to run off `.onChange(of: actions.isRecording)`,
+        // which carries no id and had to guess with `store.recordings.first`
+        // — and `add` inserts at index 0, so a Voice Memos import landing in
+        // between made the memo "the recording that just stopped".
+        onRecordingFinalized?(recording.id)
 
         // ---- END OF THE LIVE-PIPELINE-OWNING PHASE.
         //
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -589,17 +589,23 @@ struct MilaApp: App {
         let profileStoreRef = SpeakerProfileStore(settings: voiceSettings)
         _speakerProfileStore = StateObject(wrappedValue: profileStoreRef)
         // Save a voice profile when a speaker is named — if the live
-        // diarizer has a centroid for that speaker, persist it. The store
-        // refuses writes while the feature is off; the guard here means an
-        // opted-out user's centroid isn't so much as read out of the pool.
+        // diarizer observed that speaker, persist it. The store refuses
+        // writes while the feature is off; the guard here means an opted-out
+        // user's voice isn't so much as read out of the pool.
+        //
+        // `currentProfiles()` reports what was observed in *this* recording,
+        // not the pool's matching centroid — for a seeded speaker the latter
+        // carries weight that is already stored on disk, and folding it back
+        // in counts it twice. This is the only place voice profiles are
+        // written, so a recognised speaker merges exactly once per recording.
         let diarizerRef = liveDiar
         store.onSpeakerNamed = { _, rawID, name in
             guard voiceSettings.isConfigured else { return }
             if let entry = diarizerRef.currentProfiles().first(where: { $0.id == rawID }) {
                 profileStoreRef.updateProfile(
                     name: name,
-                    embedding: entry.centroid,
-                    sampleCount: entry.sampleCount
+                    embedding: entry.observedCentroid,
+                    sampleCount: entry.observedCount
                 )
             }
         }
@@ -903,11 +909,18 @@ struct MilaApp: App {
         guard voiceRecognitionSettings.isConfigured else { return }
         guard let rec = store.recordings.first else { return }
         let poolEntries = liveSpeakerDiarizer.currentProfiles()
-        // Only assign names for pool entries that were seeded AND
-        // actually matched utterances (sampleCount > the seed cap of 3).
+        // Only assign names for pool entries that were seeded (they carry a
+        // profileName) AND actually spoke. Both halves matter: a seeded
+        // speaker who never said anything must not be labelled onto the
+        // recording just because their voice was in the pool.
         for entry in poolEntries {
             guard let profileName = entry.profileName else { continue }
-            // Check this speaker actually spoke (has intervals).
+            // Check this speaker actually spoke (has intervals). Now that
+            // seeded weight is kept out of the persisted delta,
+            // `entry.observedCount > 0` says much the same thing — but
+            // `intervals` is the stronger signal, since it means the speaker
+            // reached the transcript rather than merely matching an
+            // embedding.
             let spoke = liveSpeakerDiarizer.intervals.contains { $0.speaker == entry.id }
             guard spoke else { continue }
             store.setSpeakerName(profileName, forSpeaker: entry.id, recordingID: rec.id)
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -361,6 +361,7 @@ struct MilaApp: App {
     /// Applies double-clicked `.milaconfig` files (with a confirmation sheet).
     @StateObject private var configImporter: MilaConfigImporter
     @StateObject private var updater = UpdaterViewModel()
+    @StateObject private var speakerProfileStore: SpeakerProfileStore
 
     init() {
         // RecordingStore's no-arg init handles the legacy migration and
@@ -570,6 +571,21 @@ struct MilaApp: App {
                 summarizer?.summarizeIfNeeded(rec)
             }
         }
+        // Save voice profile when a speaker is named — if the live
+        // diarizer has a centroid for that speaker, persist it.
+        let profileStoreRef = SpeakerProfileStore()
+        _speakerProfileStore = StateObject(wrappedValue: profileStoreRef)
+        let diarizerRef = liveDiar
+        store.onSpeakerNamed = { _, rawID, name in
+            if let entry = diarizerRef.currentProfiles().first(where: { $0.id == rawID }) {
+                profileStoreRef.updateProfile(
+                    name: name,
+                    embedding: entry.centroid,
+                    sampleCount: entry.sampleCount
+                )
+            }
+        }
+
         // Auto-drop accidental short+empty captures (issue #61). A recording
         // that finishes transcription with no text AND under the user's
         // minimum-duration threshold (Settings ▸ Storage, default 5s, 0 = keep
@@ -691,11 +707,20 @@ struct MilaApp: App {
                 .task { await runFinalizeRegressionIfRequested() }
                 .task { recordingSummarizer.backfillIfNeeded() }
                 .task { voiceMemosImporter.start() }
+                .onChange(of: actions.isRecording) { wasRecording, isRecording in
+                    if wasRecording, !isRecording {
+                        // Recording just stopped — auto-assign names from
+                        // recognised voice profiles for seeded pool entries
+                        // that actually matched utterances.
+                        autoAssignRecognisedSpeakers()
+                    }
+                }
                 .environmentObject(recordingSummarizer)
                 .environmentObject(meetingDetectionSettings)
                 .environmentObject(voiceMemosSettings)
                 .environmentObject(voiceMemosImporter)
                 .environmentObject(speakerDirectory)
+                .environmentObject(speakerProfileStore)
                 .environmentObject(configImporter)
                 .sheet(item: Binding(
                     get: { configImporter.pending },
@@ -774,6 +799,7 @@ struct MilaApp: App {
                 .environmentObject(voiceMemosSettings)
                 .environmentObject(voiceMemosImporter)
                 .environmentObject(speakerDirectory)
+                .environmentObject(speakerProfileStore)
                 .environmentObject(configImporter)
                 // Settings ▸ General shows the Sparkle-backed
                 // "Automatically check for updates" toggle, so the Settings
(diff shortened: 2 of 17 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

PR #210, opened 2026-08-27
fix(ui): actually skip the Settings window in applyChrome

### LINKED ISSUE #207: applyChrome(to:) never skips the Settings window — case-sensitive identifier guard
`AppDelegate.applyChrome(to:)` (`Mila/App/MilaApp.swift:1894`) opens with a guard
that is meant to leave the Settings window alone:

```swift
if let id = window.identifier?.rawValue,
   id.contains("settings") || id.contains("alert") {
    return
}
```

`String.contains(_:)` is case-sensitive, and SwiftUI names its Settings scene
window **`com_apple_SwiftUI_Settings_window`** — capital `S`. So
`id.contains("settings")` never matches, the guard falls through, and
`applyChrome` runs on the one window it was written to skip.

What then gets applied to Settings:

- `titlebarSeparatorStyle = .none` and `toolbar?.showsBaselineSeparator = false`
- the walk that switches every `NSVisualEffectView` in the window to
  `.withinWindow` blending

The `"alert"` half of the guard is likely dead for the same reason — worth
checking what identifiers actually appear rather than guessing.

**Why it matters more now:** #202 replaces the Settings tab strip with a
sidebar, so the Settings window now *has* vibrant sidebar material for that
walk to find. Before that change the guard being dead was close to harmless;
after it, Settings silently inherits main-window chrome treatment.

**Fix:** compare case-insensitively (or match the identifier exactly). Something
like:

```swift
let id = window.identifier?.rawValue.lowercased() ?? ""
if id.contains("settings") || id.contains("alert") { return }
```

Worth logging the identifiers once while fixing, to confirm the `alert` case and
avoid replacing one guess with another.

Found while reviewing the Settings layout work; not urgent, and deliberately
kept out of #202 so that PR stays a layout change.

urisland:
Closes #207

`applyChrome(to:)`'s early return never fired for the Settings window, so the
method it was meant to skip ran on Settings on every
`NSWindow.didBecomeKeyNotification`.

```swift
if let id = window.identifier?.rawValue,
   id.contains("settings") || id.contains("alert") {
    return
}
```

`String.contains` is case-sensitive, and SwiftUI names that window
`com_apple_SwiftUI_Settings_window` — capital `S`. Confirmed from an XCUITest
accessibility dump of the running app:

```
identifier: 'com_apple_SwiftUI_Settings_window', title: 'General'
```

## The `"alert"` arm was dead too — but not because of case

Worth stating separately, because case-insensitivity does **not** fix it and
"lowercase both arms" would have left a second dead guard behind. I probed the
real objects with a standalone headless AppKit binary rather than assuming:

| window | identifier | class | `is NSPanel` |
|---|---|---|---|
| `NSAlert().window` | `_NS:87` | `_NSAlertPanel` | yes |
| `NSAlert(critical).window` | `_NS:87` | `_NSAlertPanel` | yes |
| bare `NSPanel` | `nil` | `NSPanel` | yes |
| `NSOpenPanel` | `nil` | `NSOpenPanel` | yes |
| bare `NSWindow` | `nil` | `NSWindow` | no |

Two things follow:

1. An alert's identifier contains no `alert` substring in **any** casing — it is
   an opaque AppKit nib id. Matching `_NS:87` would be worse than the bug, since
   it is an implementation detail with no stability guarantee across OS
   versions.
2. `NSOpenPanel`/`NSSavePanel` and the dictation overlay's `NSPanel` report **no
   identifier at all**, so `if let id = window.identifier?.rawValue` fell
   through for them as well — meaning system open/save panels were also being
   restyled. That was never intended either.

So the two exemptions need different shapes: Settings by identifier (lowercased),
panels by **class**, which covers alerts, open/save panels, and the dictation
overlay together. The main window is a plain `NSWindow`, so it keeps its chrome
and the half-divider this method exists to remove stays removed.

## The change

A pure `nonisolated static` helper, so the decision is testable without an
`NSWindow` and without a main-actor context:

```swift
nonisolated static func isChromeExempt(identifier: String?, isPanel: Bool) -> Bool {
    if isPanel { return true }
    guard let identifier else { return false }
    return identifier.lowercased().contains("settings")
}
```

`applyChrome` now calls it with `window.identifier?.rawValue` and
`window is NSPanel`.

## No user-visible change — measured, not assumed

I initially expected this to be user-visible: with the guard firing, Settings
stops receiving `titlebarSeparatorStyle = .none`, so I assumed its default
titlebar hairline would return. **That was wrong**, and the CI screenshots #202
added are what showed it.

Comparing the `ui-screenshots` artifact from merged `main` (`ead07ad`, before
this fix) against this PR (`9a09aea`, after), every *static* Settings pane is
**byte-identical**: general, audio, speakers, meetings, aiFeatures, aiProvider,
voiceMemos. The images that do differ are exactly the ones carrying dynamic
content — `settings-models` (model install state), `settings-storage` (disk
usage), and the four main-window shots (seeded recording timestamps) — none of
which this diff touches.

The mechanism explains it: **Settings has no toolbar at all** (no `.toolbar` in
`SettingsView`, its tabs, or the `Settings` scene). So
`window.toolbar?.showsBaselineSeparator = false` was always a nil no-op, and
`titlebarSeparatorStyle = .none` changed nothing visible either, because with no
toolbar the default `.automatic` style draws no separator to begin with.

So this is behaviour-preserving dead-code removal, not a visual change. What it
buys is that the guard now means what it says, and open/save panels stop being
restyled — which is a correctness improvement in the same place rather than a
cosmetic one.

`freezeSidebarMaterials` was already a no-op on Settings and remains one: #202's
Settings is a bounded `HStack`, not a `NavigationSplitView`, so there is no
`NSSplitView` for the walk to find. Verified against merged `main`, not the
pre-#202 state.

## Tests

`MilaTests/WindowChromeExemptionTests.swift` covers the helper, including two
negative controls that reproduce each dead-guard shape and assert it fails — so
the passing cases are provably discriminating rather than vacuous:

- the real `com_apple_SwiftUI_Settings_window` is exempt, and a case-sensitive
  `contains("settings")` provably is not
- a real `NSAlert` window is exempt via its class, and
  `id.lowercased().contains("alert")` provably never matches it
- open panels are exempt, and carry no identifier
- a plain `NSWindow` (identifier or not) is **not** exempt, so the main window
  keeps its chrome
- `_NS:87` on a non-panel does not exempt, pinning that the class check is what
  identifies panels

The alert and open-panel cases assert against real AppKit objects rather than
hard-coded identifier strings, since the identifier is exactly the detail this
change stops depending on.

## Verification

`xcodebuild build-for-testing` succeeds with no new warnings in either touched
file. The tests are compile-verified locally only — app-hosted tests are not run
on the dev machine (they trigger keychain prompts), so CI is their first
execution.

## Checklist

- [x] Builds locally

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed toolbar separator handling for the Settings window.
  - Ensured alerts, open/save dialogs, dictation overlays, and other panels retain their expected window chrome.
  - Improved handling of window identifiers regardless of capitalization.

- **Tests**
  - Added coverage for Settings, panel, alert, dialog, and main-window behavior.

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -1888,6 +1888,43 @@ final class MilaAppDelegate: NSObject, NSApplicationDelegate {
         applyChrome(to: window)
     }
 
+    /// Whether `applyChrome` must leave a window alone.
+    ///
+    /// Pure and `static` so the decision is unit-testable without an
+    /// `NSWindow` — see `WindowChromeExemptionTests`. The two exemptions
+    /// are deliberately different shapes, because identifiers turn out
+    /// to be a usable discriminator for only one of them:
+    ///
+    /// - **Settings** is matched on its identifier, lowercased. SwiftUI
+    ///   names that window `com_apple_SwiftUI_Settings_window` — capital
+    ///   `S`. The original guard tested `id.contains("settings")`, which
+    ///   is case-sensitive and therefore never matched, so the early
+    ///   return never fired and this whole method ran on the Settings
+    ///   window on every `didBecomeKeyNotification` (#207). Lowercasing
+    ///   once and comparing is what actually honours the intent.
+    ///
+    /// - **Panels** are matched on their *class*, not their identifier,
+    ///   and this replaces the old `id.contains("alert")` arm rather
+    ///   than merely case-correcting it. That arm could never have
+    ///   worked: an `NSAlert`'s window is an `_NSAlertPanel` whose
+    ///   identifier is an opaque AppKit nib id — `_NS:87` when probed on
+    ///   macOS 26 — containing no recognisable substring in any case,
+    ///   and not something to depend on across OS versions. A bare
+    ///   `NSPanel` and an `NSOpenPanel` report no identifier at all, so
+    ///   `if let id = window.identifier` fell through for them too and
+    ///   they were also being restyled. Alerts, open/save panels, and
+    ///   the dictation overlay are all `NSPanel`s, and none of them
+    ///   wants a toolbar-separator tweak; the main window is a plain
+    ///   `NSWindow`, so it is unaffected.
+    /// `nonisolated` because it is pure — it touches no delegate state and
+    /// needs no main-actor context, which is also what lets the unit tests
+    /// call it synchronously.
+    nonisolated static func isChromeExempt(identifier: String?, isPanel: Bool) -> Bool {
+        if isPanel { return true }
+        guard let identifier else { return false }
+        return identifier.lowercased().contains("settings")
+    }
+
     /// Apply Mila's window chrome tweaks to a single NSWindow. Strips
     /// the hairline separator that renders below the toolbar in the
     /// detail pane but stops at the sidebar splitter — looks like a
@@ -1899,8 +1936,8 @@ final class MilaAppDelegate: NSObject, NSApplicationDelegate {
     /// macOS version + SwiftUI version, either or both can produce the
     /// line. Setting both is harmless and covers everything.
     private func applyChrome(to window: NSWindow) {
-        if let id = window.identifier?.rawValue,
-           id.contains("settings") || id.contains("alert") {
+        if Self.isChromeExempt(identifier: window.identifier?.rawValue,
+                               isPanel: window is NSPanel) {
             return
         }
         // Belt and suspenders: queue the property change behind any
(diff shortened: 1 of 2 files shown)


urisland:
@coderabbitai full review

urisland:
@coderabbitai full review

PR #217, opened 2026-08-30
fix(speakers): stop matching stored voices when diarization is turned off mid-recording

### LINKED ISSUE #215: Turning diarization off mid-recording leaves seeded voice profiles matching
`VoiceRecognitionSettings.addEnabledObserver` fires on the **`isEnabled`** toggle
only. But `isConfigured` is `isEnabled && diarizationReady`, so there is a second
way for the feature to become unconfigured mid-recording: diarization itself
being turned off (or becoming unavailable) in Settings.

When that happens no observer fires, so `LiveSpeakerDiarizer`'s pool keeps the
seeded centroids it copied at record-start, and `assign` keeps matching against
them for the rest of the recording.

**Impact is limited but not nil.** The write gates hold — `RecognisedSpeakerAssigner.finish`
guards on `isConfigured`, so nothing is persisted and no profile is created or
updated. What survives is the *read* path: seeded names can still auto-label the
in-flight transcript from stored voices after the feature stopped being
configured.

**Why it is worth closing anyway.** #204 established the rule that revoking
access takes effect immediately and completely, and wired two observers to
enforce it — `SpeakerProfileStore.addDeletionObserver` (profiles deleted) and
`addEnabledObserver` (opt-out), both calling
`LiveSpeakerDiarizer.forgetSeededProfiles`. This is the same rule with a third
trigger that was simply not wired, so the inconsistency is the bug rather than
the impact.

**Likely shape of the fix:** observe `isConfigured` rather than `isEnabled`, or
add a second observer on `diarizationReady`, calling the same
`forgetSeededProfiles()`. The mechanism already exists — `addEnabledObserver`
holds a list of observers, and `forgetSeededProfiles` is idempotent (entries
whose `profileName` is already `nil` are skipped), so an extra trigger is safe.

A regression test in the style of `VoiceRecognitionGateTests` would pin it:
seed the pool, flip `diarizationReady` false mid-recording, and assert a stored
voice is no longer auto-labelled.

Found while fixing the opt-out case in #204; deliberately left out of that PR as
a different toggle and a wider change than the finding covered.

urisland:
Closes #215

#204 established that revoking access to learned voices takes effect
**immediately and completely**, and wired two observers to enforce it — profiles
deleted, and the user opting out. Both call
`LiveSpeakerDiarizer.forgetSeededProfiles(...)`.

There is a third way for the feature to stop being available mid-recording, and
it was not wired: `isConfigured` is `isEnabled && diarizationReady`, so turning
**diarization** off (or it becoming unavailable) leaves the pool holding the
centroids it copied at record-start, and `assign` keeps matching against them
until the recording stops.

Confirmed before changing anything, with a standalone harness compiling the real
`LiveSpeakerDiarizer` / `VoiceRecognitionSettings` / `SpeakerProfileStore` /
`ObservedVoiceSnapshots` out of the worktree. Against unfixed `main`:

- diarization off mid-recording → `assign` still returned `SPEAKER_00`, pool
  still held `profileName == "Alice"` — the bug;
- opt-out → correctly minted a new nameless speaker (#204's path, working).

The write gates always held, so nothing was ever persisted. What survived was the
**read** path: stored voices could still put a name on the in-flight transcript
after the feature stopped being configured. The inconsistency between the three
revocation triggers is the actual defect here rather than the impact.

## The fix — observe the gate, not the toggle

`VoiceRecognitionSettings` gains `addConfiguredObserver(_:)` /
`refreshConfiguredState()`: a second observer channel keyed on `isConfigured`,
firing on **transitions only**. It is driven by `isEnabled`'s `didSet` (so opt-out
still lands synchronously) and by assignments to `diarizationReady`.
`trackDiarizationReadiness(_:)` subscribes to the `objectWillChange` publishers
behind `diarizationReady`. `MilaApp` moves the `forgetSeededProfiles()`
registration onto the new channel.

**The main-actor hop is load-bearing.** `objectWillChange` fires on *willSet*, so
reading `isConfigured` inside the sink would see the pre-change value — the exact
trap the existing `isEnabled` comment documents. It still lands in time: the
pool's only reader (`process` → `assign`) is enqueued as a main-actor task too,
so an utterance captured *after* the change queues behind the refresh, while one
captured before it is legitimately labelled.

**`addEnabledObserver` deliberately keeps its two remaining registrants.**
`SpeakerProfileStore` (load/unload) and `ObservedVoiceSnapshots` answer to the
user's *consent*, not to diarization's availability. Moving them onto the gate
would empty the Settings profile-management list whenever the Python pipeline was
merely unready — a regression well outside this issue.

Both readiness publishers are subscribed, because `DiarizationSettings.isConfigured`
reads `bootstrap.isReady` on the bundled-runtime path and `nuclearRepair` clears
that without touching the settings object.

## Tests

`VoiceRecognitionGateTests` gains a `SeededRecording` fixture and a `PoolWiring`
enum (`.none` = pre-#204, `.toggleOnly` = #204, `.gate` = shipping) so each
control names the shape it reconstructs. The fixture sends `objectWillChange`
**before** the value settles, exactly as `@Published` does, so an implementation
that skipped the hop fails it.

Covered: the diarization-off case for a not-yet-heard speaker; a **negative
control** on `.toggleOnly` that still matches `SPEAKER_00` as "Alice", so the
main assertion is provably discriminating; the already-heard nuance (keeps
`SPEAKER_00`, loses the name — pinned against a future "fix" that mints a new id
and forks one person across two speakers mid-transcript); opt-out still landing
synchronously through the new channel; and transition-only notification, so an
`objectWillChange` firehose cannot re-run the handler on a loop.

## Verification

`xcodebuild build-for-testing` clean twice, no new warnings in the touched files.
The XCTest target was **not** run locally (app-hosted tests are not run on the
dev machine); instead every new assertion was mirrored into the harness and run
against the real objects — all pass, and both negative controls reproduce their
pre-fix shapes. CI is the first real XCTest run.

## Checklist

- [x] Builds locally (`make build`); tests written, left for CI to run
- [ ] README changelog — N/A, entries are written per release

## Summary by CodeRabbit

* **Bug Fixes**
  * Voice recognition data is now cleared when diarization becomes unavailable during an active recording, not only when users disable the feature.
  * Speaker continuity during recording is preserved while the feature remains configured.
  * Configuration change notifications now occur only when the effective feature state changes.

* **Documentation**
  * Updated guidance to clarify when voice recognition profiles are removed.

---

> [!NOTE]
> **Medium Risk**
> Changes live voice-recognition revocation during recording (privacy-sensitive read path), but scope is narrow and heavily regression-tested.
> 
> **Overview**
> Fixes **#215**: during an active recording, turning **diarization** off (or losing pipeline readiness) no longer leaves the live speaker pool matching stored voice centroids copied at record-start. Writes were already gated; this closes the **read** path that kept auto-labeling the transcript after the feature stopped being usable.
> 
> **`VoiceRecognitionSettings`** adds a second observer channel on **`isConfigured`** (`isEnabled && diarizationReady`): `addConfiguredObserver`, transition-only `refreshConfiguredState()`, and `trackDiarizationReadiness` on diarization `objectWillChange` publishers (with a deferred main-actor refresh so reads see settled state). **`MilaApp`** registers `LiveSpeakerDiarizer.forgetSeededProfiles()` on that gate instead of the opt-out toggle. **`addEnabledObserver`** stays for consent-scoped holders (in-memory profiles and snapshots).
> 
> **`VoiceRecognitionGateTests`** adds a `SeededRecording` / `PoolWiring` fixture and coverage for diarization-off mid-recording, toggle-only negative control, transition-only notifications, and opt-out still clearing the pool synchronously.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 770011805c8b29613ed7737f0c7663b80f840e21. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -641,21 +641,37 @@ struct MilaApp: App {
         // …and the third holder of copied voice data: the live pool. The two
         // lines above unload the profiles (`SpeakerProfileStore`'s own
         // observer) and drop the snapshots, but `seedPool` handed the
-        // diarizer its own copy of every centroid at record-start, so an
-        // opt-out mid-recording otherwise leaves `assign` matching against
-        // stored voices for the rest of it. The write gates stop the result
-        // being persisted; they do not stop it being *read*, and the user
-        // still watches their transcript auto-fill with names from a feature
-        // they just switched off.
+        // diarizer its own copy of every centroid at record-start, so the
+        // feature becoming unusable mid-recording otherwise leaves `assign`
+        // matching against stored voices for the rest of it. The write gates
+        // stop the result being persisted; they do not stop it being *read*,
+        // and the user still watches their transcript auto-fill with names
+        // from a feature that is no longer on.
         //
         // Exactly the deletion path's problem, so it gets the deletion
-        // path's remedy — see `addDeletionObserver` below. Opt-out is the
+        // path's remedy — see `addDeletionObserver` below. This one is the
         // broader of the two, so it forgets every seeded entry rather than
         // named ones.
-        voiceSettings.addEnabledObserver { [weak liveDiar] nowEnabled in
-            guard !nowEnabled else { return }
+        //
+        // Registered on the **gate**, not the toggle. `isConfigured` is
+        // `isEnabled && diarizationReady`, and either half closing revokes
+        // the feature just as completely: turning diarization off in
+        // Settings mid-recording used to fire nothing at all, so the pool
+        // went on recognising stored voices (#215). `trackDiarizationReadiness`
+        // below is what lets the settings object notice the second half.
+        voiceSettings.addConfiguredObserver { [weak liveDiar] nowConfigured in
+            guard !nowConfigured else { return }
             liveDiar?.forgetSeededProfiles()
         }
+        // The readiness half of that gate lives on `DiarizationSettings`, and
+        // `diarizationReady` above only *reads* it — nothing tells this
+        // object when the answer changes. Both publishers are needed:
+        // `isConfigured` there is `isEnabled && (hasBundledRuntime ?
+        // bootstrap.isReady : status.isGood)`, so the bootstrap's readiness
+        // is a genuinely separate source of truth (`nuclearRepair` clears it
+        // without touching the settings object at all).
+        voiceSettings.trackDiarizationReadiness(diarSettings.objectWillChange,
+                                                diarSettings.bootstrap.objectWillChange)
         // Auto-naming + the per-recording snapshot, driven by the finalize
         // drain rather than by observing `isRecording`. The drain hands over
         // the id of the recording that actually finished, and calls
--- a/Mila/Models/SpeakerProfileStore.swift
+++ b/Mila/Models/SpeakerProfileStore.swift
@@ -174,8 +174,9 @@ final class SpeakerProfileStore: ObservableObject {
     /// recording's centroid, which is the same person's fingerprint to
     /// within rounding. `MilaApp` wires this to
     /// `LiveSpeakerDiarizer.forgetSeededProfiles`, so the deletion takes
-    /// effect immediately and everywhere, exactly as opting out does through
-    /// `VoiceRecognitionSettings.addEnabledObserver`.
+    /// effect immediately and everywhere, exactly as the feature ceasing to
+    /// be configured does through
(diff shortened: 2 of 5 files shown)


urisland:
bugbot run

Issue #206, opened 2026-08-27
Tune the seed anchor weight for returning-speaker matching


urisland:
`seedPool` starts a seeded entry at `min(persistedSampleCount, 3)`. That cap is
the sole knob controlling how strongly a returning speaker's stored centroid
anchors matching, versus adapting to today's acoustics (different mic, room,
headset). The value `3` was picked without measurement.

Now that seeded weight never reaches persistence (fixed in #204 — seeded weight
feeds `centroid`/`sampleCount` for matching only, while `observedCentroid`/
`observedCount` carry the persisted delta), this cap affects **matching only**.
It can therefore be tuned in isolation, with no risk of corrupting stored
profiles — which was not true before #204.

A related question worth evaluating at the same time: the stored centroid is a
plain running mean over every observation ever recorded for that speaker, so it
never forgets old acoustics. A decayed or windowed mean may recognise returning
speakers better, particularly for someone whose setup has changed since their
profile was first built.

Both need real multi-recording audio with the same speakers across sessions to
evaluate — unit tests can pin the arithmetic but cannot tell us whether
recognition actually improves.

Origin: CodeRabbit's second Major finding on #204, whose correctness half is
fixed there.

urisland:
I looked at this closely enough to sweep the value, and concluded I should **not**
change it. Below: what the knob actually does (with the arithmetic), what moving it
would trade, a protocol that would settle it, and my read on the decayed-mean question.

## What the knob does, mechanically

`LiveSpeakerDiarizer.seedPool` turns each stored profile into a pool entry with
`sampleCount = min(entry.sampleCount, 3)` — call it **n₀ ∈ {1,2,3}** — while
`observedCentroid`/`observedCount` start empty/zero. Since #204 that split is clean:
n₀ never reaches persistence, so it is purely a matching parameter.

In `assign`, a confident match (`sim >= similarityThreshold` **and** equal dimensions)
folds the utterance into the matching centroid as a running mean:

```swift
centroid[i] = (centroid[i] * Float(n) + embedding[i]) / Float(n + 1)
sampleCount = n + 1
```

Unroll that over *k* confident matches in one recording and it telescopes exactly:

```
centroid_k = (n₀·c₀ + Σ e_j) / (n₀ + k)
```

So n₀ is literally **"how many of this session's utterances the stored centroid is
worth"**. The weight the stored voice still carries when the (k+1)-th utterance is
compared against it:

[...]

PR #226, opened 2026-08-31
fix(speakers): naming a speaker mid-recording now creates a voice profile

### LINKED ISSUE #209: Naming a speaker during a recording never creates a voice profile
Naming a speaker **during** a recording, from the live transcript, never creates
or updates a voice profile. It looks like it worked — the name appears — but
nothing is learned, so the same speaker is unrecognised next time.

**Cause.** `LiveAIRecordingView` names speakers by writing
`transcriber.speakerNames[raw]` directly. Voice-profile persistence hangs off
`RecordingStore.setSpeakerName`, which synchronously fires `store.onSpeakerNamed`
— the hook `MilaApp.init` wires to `SpeakerProfileStore.updateProfile`. The live
path never reaches `setSpeakerName`, so the hook never fires.

Naming the *same* speaker afterwards from `RecordingDetailView` does go through
`setSpeakerName` and does persist, which is why this is easy to miss: the
feature works, just not from the place a user is most likely to use it.

**Not a regression.** This predates the cross-recording voice recognition work
(#204) and is unchanged by it — the `ObservedVoiceSnapshots` fix keys snapshots
by recording id and would serve this path correctly if it ever called
`setSpeakerName`. Found while auditing every `rawID` consumer for that change.

**Why it is worth fixing rather than documenting.** Mid-recording is exactly
when the user knows who is speaking — they are looking at the person. Being
asked again after the fact, having already labelled them, reads as the feature
being broken. The silence is the problem: there is no error, the label sticks in
the transcript, and only the profile is missing.

**Direction.** Route live naming through `RecordingStore.setSpeakerName` so it
gets the same hook, rather than adding a second persistence call at the live
site — a parallel call is what caused the double-merge bug fixed in #204, and
one persistence trigger is the invariant that PR deliberately established.

Note the live recording may not have a `Recording` in the store yet at that
point; if so, the fix needs to either defer the profile write until finalisation
or reconcile it against the recording id once it exists. Worth checking before
picking an approach.

Related: #206 (seed anchor weight).

urisland:
Closes #209

## What & why

Naming a speaker **during** a recording, from the live transcript, never created or
updated a voice profile. The label appeared and stuck, no error was shown, and the same
person was unrecognised in the next recording. Naming them afterwards from
`RecordingDetailView` *did* persist, which is what made this easy to miss.

### The finding the issue asked for: there is no store row at naming time

Confirmed by reading the code, not inferred. `QuickActionsController.stopRecording`
creates the `Recording` and calls `store.add(recording)` **inside stop**, after
`await session.stop()` returns the output URL — there is no `Recording` in the store at
any point while the recording is running. So `RecordingStore.setSpeakerName(_:forSpeaker:
recordingID:)` is not merely unused from the live pane, it is *unreachable*: its first
line is `guard let idx = recordings.firstIndex(where: { $0.id == recordingID }) else
{ return }`.

That is why `LiveAIRecordingView` writes `transcriber.speakerNames[raw]` directly, and
it is the reason the fix has to defer rather than reroute at the live site.

### Why nothing was learned

Voice-profile persistence hangs off exactly one trigger: `setSpeakerName` fires
`store.onSpeakerNamed`, which `MilaApp.init` wires to
`SpeakerProfileStore.updateProfile`. One trigger is a **deliberate invariant** — a
parallel `updateProfile` call alongside it is what caused the double-merge fixed in
#204, where every recognised speaker folded in twice, `sampleCount` doubled, and the
profile went progressively rigid with no visible failure.

The live map reached the row without ever passing through that trigger, twice over:

* `store.add(Recording(… speakerNames: liveTranscriber?.speakerNames ?? [:]))`
* `updated.speakerNames = liveTranscriber?.speakerNames ?? [:]` before `store.update`

Both run *before* `onRecordingFinalized?(recording.id)`. So by the time
`RecognisedSpeakerAssigner.finish` called `setSpeakerName`, the row already carried the
name and the no-change guard (`guard recordings[idx].speakerNames[rawID] != trimmed else
{ return }`) returned before firing the hook.

### The fix

Route the live names through the same single trigger, at the one moment it is reachable
and the embeddings are final.

* **`QuickActionsController`** — both pre-writes are gone. `onRecordingFinalized` is now
  `(UUID, [String: String]) -> Void` and carries `liveTranscriber?.speakerNames`, read
  before the live-singleton teardown.
* **`RecognisedSpeakerAssigner.finish(recording:liveSpeakerNames:)`** — after snapshotting
  the pool (which must happen before anything can fire the hook), applies each live name
  with `store.setSpeakerName`. No second persistence call anywhere.
* **`MilaApp`** — passes the map through to `finish`.
* **`LiveAIRecordingView`** — comment only, recording why the write goes to the
  transcriber and where persistence happens, so the next reader doesn't "fix" it by
  adding a profile write at the live site.

Three consequences, all deliberate:

1. **Live names are applied before the auto-naming loop, and that loop now skips any raw
   id the user named.** The user's label beats a recogniser's guess, and each raw id
   fires the hook at most once — letting the auto-name overwrite "Bob" with a seeded
   "Alice" would fire `onSpeakerNamed` twice and fold this recording's centroid into
   *both* profiles. This also fixes a smaller adjacent bug: previously the auto-name
   silently clobbered a mid-recording label.
2. **Live names are not behind the voice-recognition gate**; the snapshot and the
   auto-naming loop still are. A display name is not voice data and has to keep working
   with recognition off (`VoiceRecognitionGateTests.
   test_while_off_naming_a_speaker_persists_no_embedding` already pins that for the
   detail-view path). Gating it would have dropped every mid-recording label for the
   majority of users who never opt in — a worse bug than the one being fixed.
   `isConfigured` is read once into a local so the two gated regions cannot disagree.
3. **`stopRecording` re-reads the row after the hook.** `setSpeakerName` writes into the
   store, not into the local `updated` snapshot the finalize tail carries; without the
   re-read, `finalizeTail`'s `TranscriptExporter.writeSRT(for: updated, …)` would
   overwrite the named `.srt` that `setSpeakerName` had just written with an unnamed one.

Nothing reads the row's `speakerNames` in the window where they are briefly absent: the
post-record sheet and `PostRecordingCoordinator.awaitTranscript` both render the
transcript only once the row leaves `.pending`/`.running`, and the live MCP sidecar
carries its own copy of the map (`LiveTranscriptSidecarWriter.update(segments:
speakerNames:)`), so the MCP surface is unaffected.

### Not done, on purpose

The issue's own comment thread proposed calling the hook **directly** after
`store.update(updated)`. That is the second persistence call #204 exists to prevent, and
it would also bypass `setSpeakerName`'s idempotency — a duplicated stop would merge
twice. Deferring the *name* instead of duplicating the *write* keeps one trigger.

## How it was tested

`make build` / `make test` could not be run: no Xcode, no Swift toolchain on this
machine. Verification is by reading the actual call paths (cited above) plus CI on this
PR. New tests, in `MilaTests/LiveSpeakerNamingTests.swift`:

**`LiveSpeakerNamingTests`** (real `RecordingStore`, `SpeakerProfileStore`,
`LiveSpeakerDiarizer`, `ObservedVoiceSnapshots`, `RecognisedSpeakerAssigner`, hook wired
exactly as `MilaApp.init` does, with a tap that logs every fire):

* `test_a_name_assigned_during_the_recording_creates_a_voice_profile` — the fix.
* `test_the_pre_fix_shape_learns_nothing` — **negative control.** Reconstructs the
  removed copy (writes the live map onto the row, then finalizes) and measures the
  silence: label present, no profile, hook never fired. Proves the test above
  discriminates — it is *not having pre-written the row* that makes the profile appear.
* `test_running_finalization_twice_folds_the_live_name_in_once` — the anti-double-merge
  invariant.
* `test_a_live_name_wins_over_the_recognised_profile_name` — precedence, and exactly one
  fire; the seeded profile must not learn a voice the user says isn't theirs.
* `test_an_unnamed_speaker_is_still_auto_named` — the skip is per raw id, not global.
* `test_a_live_name_is_applied_with_voice_recognition_off` — label lands, nothing stored,
  no profiles file, no embeddings held.
* `test_a_live_name_for_an_unobserved_speaker_stores_nothing` — fail-closed on a raw id
  this recording never observed.

**`LiveSpeakerNamingStopRecordingTests`** — drives the real
`QuickActionsController.stopRecording` (fake-session seam + `StubWhisperEngine`, as
`LiveSidecarHandoffOrderingTests` does):

* `test_naming_a_speaker_mid_recording_learns_the_voice_exactly_once` — the reported bug
  end to end. **Fails on the pre-fix code**: with either copy in place the hook never
  fires, so `fires` is empty and no profile exists. Re-adding either pre-write turns it
  red again.
* `test_naming_nothing_leaves_the_saved_recording_unnamed` — the hook's new argument is
  the live map, not a synthesised one.

Both classes assert the **fire count**, not merely that a profile exists, because "one
persistence trigger" is the invariant that a passing profile-exists assertion would hide.

No pre-existing test asserted the buggy behaviour (checked every `speakerNames` /
`setSpeakerName` / `onSpeakerNamed` reference under `MilaTests/`), so nothing had to be
inverted. `VoiceRecognitionGateTests.test_stopping_right_after_an_opt_out_names_nothing_and_stores_nothing`
and the whole of `RecognisedSpeakerAssignerTests` keep compiling and passing unchanged —
`liveSpeakerNames` defaults to `[:]`, which is the "no live names" no-op, and the
opt-out test passes none.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — **not possible on this
      machine (no Xcode / no Swift toolchain); relying on CI**
- [x] User-facing change is noted in the README changelog (if applicable) — n/a, the
      changelog is per release and there is no version bump here

## Related

#206 (seed anchor weight) was evaluated alongside this and deliberately **not** changed —
analysis and an evaluation protocol are posted on that issue. No code change there, so it
is not closed by this PR.

## The build is unverified

Stated plainly because it changes how the first CI run should be read: this was written
on Linux with **no Xcode, no `xcodebuild` and no Swift toolchain**, so `make build` and
`make test` were never run and CI is the first compiler to see this diff. What was done
instead, by hand:

* Every project symbol referenced by the new test file was grep-confirmed to exist with
  the signature used (`TestSupport.*`, `StubWhisperEngine.setDefaultCanned`,
  `SpeakerProfileStore.Deletion.all`/`.named`, `VoiceProfile.sampleCount`/`.embedding`,
  `QuickActionsController.startFakeRecordingForTesting`/`awaitFinalizeTails`,
  `TranscriptionService.waitForIdle`, `LiveSpeakerDiarizer.assign`/`currentProfiles`/
  `seedPool`/`forgetSeededProfiles`, `ObservedVoiceSnapshots.observation(forSpeaker:in:)`/
  `heldRecordingCount`, `RecordingStore.freshAudioURL`/`update`/`setSpeakerName`). The
  only unresolved names left over were Foundation APIs and string literals.
* Brace / paren / bracket balance checked on all four changed Swift files.
* `UserDefaults` isolation follows the convention in `CLAUDE.md`: the assigner-level class
  uses a fresh `"LiveSpeakerNamingTests.<UUID>"` suite per fixture (collected and removed
  in `tearDown`, mirroring `RecognisedSpeakerAssignerTests`); the controller-level class
  uses `"LiveSpeakerNamingStopRecordingTests.{diarization,language,llm,voice}"` and
  `TestSupport.isolatedRemoteSettings(label:)`, all removed in `tearDown`. No test reads
  or writes `.standard` except the `selectedModelName` save/restore the sibling suites
  already do.
* Every existing caller of the two changed signatures was enumerated: `.finish(recording:`
  has one production call site (`MilaApp`, updated) and 15 test call sites, all of which
  keep compiling because `liveSpeakerNames` defaults to `[:]`; `onRecordingFinalized` has
  exactly three references, all updated.
* The blast radius of removing the two `speakerNames` pre-writes was traced to every
  reader of `Recording.speakerNames` between `store.add` and the finalize hook
  (`RenameRecordingSheet`, `SendToLLMSheet`, `PostRecordingCoordinator.awaitTranscript`,
  `TranscriptExporter`, `SpeakerNameRemapper` in `finalizeTail`, and the MCP
  `MilaStoreReader` path) — none of them observe the gap, for the reasons given above.

---

---

> [!NOTE]
> **Medium Risk**
> Changes finalize-time speaker naming and voice-profile persistence on the recording stop path; mistakes could regress learning, double-merge profiles, or corrupt SRT sidecars, but the flow is narrow and heavily tested.
> 
> **Overview**
> Fixes **island-io/mila#209**: labels typed in the **live transcript** during a recording appeared on the saved transcript but never created or updated a voice profile, because `stopRecording` copied `LiveTranscriber.speakerNames` onto the new row **before** `onRecordingFinalized` ran. That made `RecordingStore.setSpeakerName` hit its no-change guard and skip `onSpeakerNamed` — the only path that writes embeddings.
> 
> **`QuickActionsController`** stops seeding or copying live speaker names at `store.add` and in the finalize drain. **`onRecordingFinalized`** now passes `(recordingID, liveSpeakerNames)` read from the transcriber before teardown, then **re-reads the row** so `finalizeTail`’s SRT export does not overwrite named subtitles.
> 
> **`RecognisedSpeakerAssigner.finish(recording:liveSpeakerNames:)`** applies live names first via `setSpeakerName` (including when voice recognition is off), then auto-names from stored profiles only for speakers the user did not label. User-chosen names win over recogniser guesses and each raw speaker id triggers persistence at most once.
> 
> **`MilaApp`** wires the new callback signature. **`LiveAIRecordingView`** adds comments documenting why naming stays on the transcriber until finalize.
> 
> New **`LiveSpeakerNamingTests`** and **`LiveSpeakerNamingStopRecordingTests`** cover profile creation, the pre-fix negative control, idempotency, precedence, recognition-off labeling, and end-to-end `stopRecording`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78e6efe078fdcace87115f9817fd81be2edca9ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -166,7 +166,20 @@ final class QuickActionsController: ObservableObject {
     /// recovering one from `store.recordings.first` is unreliable because
     /// `add` inserts at index 0 with no re-sort, so any recording added in
     /// between (a Voice Memos import, say) becomes `first`.
-    var onRecordingFinalized: ((UUID) -> Void)?
+    ///
+    /// The second argument is the speaker names the user assigned from the
+    /// LIVE transcript pane while the recording ran (`LiveTranscriber.
+    /// speakerNames`). They travel through this hook rather than being copied
+    /// onto the row here because that copy was the bug in island-io/mila#209:
+    /// a name already on the row makes `RecordingStore.setSpeakerName`
+    /// no-change-guard out before it can fire `onSpeakerNamed`, the one hook
+    /// that persists a voice profile — so a mid-recording label stuck to the
+    /// transcript and taught the recogniser nothing, silently. There is no
+    /// row to name while the recording runs (`store.add` happens in
+    /// `stopRecording`), so deferring to here is what makes that single
+    /// persistence trigger reachable at all. See `RecognisedSpeakerAssigner.
+    /// finish(recording:liveSpeakerNames:)`.
+    var onRecordingFinalized: ((_ recordingID: UUID, _ liveSpeakerNames: [String: String]) -> Void)?
 
     /// Late-bound by MilaApp. When the live-transcript path saves a
     /// recording directly (skipping `transcription.enqueue` because the
@@ -679,6 +692,20 @@ final class QuickActionsController: ObservableObject {
         // and chunk modes produce live segments we want to preserve,
         // so the initial-status gate is segment-presence, not mode.
         let initialStatus: TranscriptionStatus = initialSegments.isEmpty ? .pending : .running
+        // `speakerNames` is deliberately NOT seeded from `liveTranscriber`
+        // below, and neither is it copied onto `updated` in the drain. Both
+        // copies used to exist, and between them they made mid-recording
+        // speaker naming teach the recogniser nothing at all
+        // (island-io/mila#209): the name was already on the row by the time
+        // `onRecordingFinalized` fired, so `setSpeakerName`'s no-change guard
+        // returned before firing `onSpeakerNamed` — the only hook that writes
+        // a voice profile. The live map now travels through that hook instead,
+        // which applies it via `setSpeakerName` and so keeps ONE persistence
+        // trigger (a parallel write is what caused the double-merge fixed in
+        // #204). Nothing reads the row's names in the meantime: the
+        // post-record sheet renders the transcript only once the row leaves
+        // `.pending`/`.running`, and the live MCP sidecar carries its own copy
+        // of the map.
         let recording = Recording(
             title: title,
             duration: duration,
@@ -691,8 +718,7 @@ final class QuickActionsController: ObservableObject {
             appName: appName,
             appBundleID: appBundleID,
             summary: initialSummary.isEmpty ? nil : initialSummary,
-            actionItems: initialItems.isEmpty ? nil : initialItems,
-            speakerNames: liveTranscriber?.speakerNames ?? [:]
+            actionItems: initialItems.isEmpty ? nil : initialItems
         )
         store.add(recording)
         // The live sidecar is deliberately NOT closed here. `finish()`
@@ -867,11 +893,15 @@ final class QuickActionsController: ObservableObject {
             return
         }
         updated.segments = finalTranscriptSegments
-        // Mid-recording speaker renames from the live pane. Snapshotted
-        // here (before `liveTranscriber?.stop()` below) alongside the
-        // segments they label; `finalizeTail` remaps them if the offline
-        // re-diarize pass re-keys the speaker IDs.
-        updated.speakerNames = liveTranscriber?.speakerNames ?? [:]
+        // Mid-recording speaker renames from the live pane are NOT copied
+        // onto the row here — see the note at `store.add` above. They are
+        // handed to `onRecordingFinalized` below (still before
+        // `liveTranscriber?.stop()`, so the map is intact) and applied with
+        // `store.setSpeakerName`, which is what lets them persist a voice
+        // profile. `finalizeTail` still remaps whatever the row ends up
+        // holding if the offline re-diarize pass re-keys the speaker IDs — it
+        // re-reads the row rather than using this snapshot, precisely so a
+        // name assigned after this line is not lost.
         // Always preserve fullText when we have live segments — the
         // sheet should show what the user just saw on screen, even
         // for chunk mode while the batch diarization pass is still
@@ -927,7 +957,21 @@ final class QuickActionsController: ObservableObject {
         // which carries no id and had to guess with `store.recordings.first`
         // — and `add` inserts at index 0, so a Voice Memos import landing in
         // between made the memo "the recording that just stopped".
-        onRecordingFinalized?(recording.id)
+        //
+        // It also carries the live pane's speaker names, which is what makes
+        // naming a speaker mid-recording learn a voice profile at all
+        // (island-io/mila#209). Read from `liveTranscriber` here — before the
+        // teardown below — and applied on the far side through
+        // `store.setSpeakerName`, the single persistence trigger.
+        onRecordingFinalized?(recording.id, liveTranscriber?.speakerNames ?? [:])
+        // Those names are written into the STORED row by `setSpeakerName`, not
+        // into this local copy, so re-read it. Without this the tail below
+        // carries a `speakerNames`-less snapshot and its
+        // `TranscriptExporter.writeSRT(for: updated, …)` would overwrite the
+        // named `.srt` that `setSpeakerName` just wrote with an unnamed one.
+        if let named = store.recordings.first(where: { $0.id == recording.id }) {
+            updated = named
+        }
 
         // ---- END OF THE LIVE-PIPELINE-OWNING PHASE.
         //
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -681,8 +681,8 @@ struct MilaApp: App {
             case .named(let names): liveDiar?.forgetSeededProfiles(named: names)
             }
         }
-        actions.onRecordingFinalized = { [assigner] recordingID in
-            assigner.finish(recording: recordingID)
+        actions.onRecordingFinalized = { [assigner] recordingID, liveSpeakerNames in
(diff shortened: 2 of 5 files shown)


urisland:
bugbot run

---

Issue #208, opened 2026-08-27
unit-and-ui-tests is ~20% flaky, and it hid main going red


urisland:
`unit-and-ui-tests` (the **App Tests** workflow) fails roughly one run in five,
and it takes a *different* test with it each time. That is enough to make a red
tick meaningless — it trains everyone to rerun on failure, which is exactly how
a real regression gets waved through.

### Evidence

Last ten runs of the job, most recent last:

| commit | result | failing tests |
|---|---|---|
| `e7dc3c6` Release 1.9.3-beta.1 | pass | — |
| `8841fe1` #192 | pass | — |
| `19d3e7d` #191 | pass | — |
| `4ad4368` #197 | pass | — |
| `6b64e8b` revert of #194 (**main**) | **fail** | `LLMRunnerTests.test_runner_surfaces_nonzero_exit_code`, `PostRecordingCoordinatorSendTests.test_second_send_for_same_id_replaces_the_first` |
| #183, #202, #196, #204 heads | pass | — |
| `c9de9d3` #204 | **fail** | `LLMRunnerTests.test_runner_spawns_child_in_isolated_empty_directory`, `ObsidianGitSyncerIntegrationTests.test_sync_rebases_nonconflicting_remote_change_and_pushes` (all 3 retries) |

The two failing sets are disjoint apart from *which suite* they belong to, and
the commits involved do not touch the failing code — the revert of #194 changed
`SettingsView` only, yet took down an `LLMRunner` test and a
`PostRecordingCoordinator` test. So this is environmental/timing, not a broken
assertion.

### Suspected causes, by suite

- **`LLMRunnerTests`** — these spawn the real `claude` CLI. Two different tests
  in this suite have now failed on unrelated commits. There is a known
  wedging problem where `waitUntilExit` can hang after `SIGKILL` on macOS 26,
  so any unbounded wait here is a candidate.
- **`ObsidianGitSyncerIntegrationTests`** — drives real `git` against a
  temporary remote. `test_sync_rebases_nonconflicting_remote_change_and_pushes`
  failed on **all three** retry iterations, so within a single run it is
  deterministic; across runs it is not. That smells like leftover state or a
  port/path collision with a concurrent run rather than randomness.
- **`PostRecordingCoordinatorSendTests`** — timing around replacing an
  in-flight send.

### Why this matters now

`main` has been red since 08:53 today and nobody noticed, because the failure
looked like the usual noise. Separately, the suite's flakiness meant four PRs
[...]

PR #227, opened 2026-08-31
fix(llm): restore action items on the one-shot summary (relay of #200)

### LINKED ISSUE #199: One-shot summary lost action items and can dump raw JSON as the summary

urisland:
Closes #199

## Credit: this fix is @ValeroK's work

**The one-shot summary fix was written by @ValeroK. This PR relays it — it
supersedes #200, which is where the work actually happened.** His commit is
cherry-picked unmodified; everything I added sits in a single commit on top,
and it is three follow-up fixes plus tests, not the fix.

It is relayed rather than merged for the usual reason (the #198 → #204 and
#129 → #164 precedent): a fork PR can't get an automated review under the
contributor's own quota — #200 never got one, CodeRabbit refused it twice for
"review limit reached" on his account — and fork CI needs per-run approval on
every run.

> **On authorship.** The relay was prepared with `git cherry-pick`, so his
> commit keeps `Author: KobiValero-Pagaya <kobi.valero@pagaya.com>` — the
> address recorded on #200's head. My commit carries
> `Co-authored-by: ValeroK <8961499+ValeroK@users.noreply.github.com>`, which
> is the form that resolves to his GitHub account.

## What & why

After #87 the post-recording summarizer only stored plain text: the one-shot
call never asked for action items, and any JSON envelope a model returned out
of habit landed in `Recording.summary` verbatim — a raw `{...}` blob in the
detail view, the Obsidian note and the MCP surface.

**@ValeroK's commit** (`Mila/Actions/RecordingSummarizer.swift`,
`Mila/Models/LiveAISettings.swift`):

- The one-shot prompt is wrapped at send time with an output contract: the
  plain-text summary, then `###ACTION_ITEMS###` on its own line, then a JSON
  array. Keeping the summary *outside* JSON is the point — a long multi-line
  Hebrew summary full of unescaped quotes cannot break the parse, because only
  the compact item array has to be valid JSON. The contract is appended, not
  persisted, so a customised `summaryPrompt` keeps working.
- Parse order: sentinel → decodable envelope → salvage a broken envelope →
  prose. Items land on the recording; an empty list is treated as "keep what's
  there", so a glitchy response can't wipe the live snapshot taken at stop.
- The default summary prompt regains its pre-#87 depth, with users still on
  the old concise default migrated up (`legacySummaryPromptV1`) and custom
  prompts left alone. I verified `legacySummaryPromptV1` is byte-identical to
  the default on `main`, so the match can't silently miss, and the migration is
  a pure read (property observers don't fire during `init`), so it is
  idempotent as its comment claims.

**On top of that (my commit), three things found by reading the paths this
change now owns:**

1. **Dictated action items were being replaced by the model's guess.** For a
   Live-AI recording, `QuickActionsController` snapshots the live items onto
   the recording at stop — voice-command items included, the ones the user
   said out loud — and then immediately calls `regenerate`. The one-shot
   prompt is never shown that list and only ever asks for
   `source: "inferred"`, so replacing wholesale traded an explicit instruction
   for the model's re-derivation of it, or lost it. `ActionItem.source` is
   mirrored all the way out to the MCP surface precisely so a client can tell
   a spoken commitment from an inferred one, which is the answer that
   degraded. `merged(fresh:keepingDictatedFrom:)` keeps every dictated item
   and folds the fresh set in after it, deduped by id and by normalised text.
2. **The salvage leaked the items array on pretty-printed JSON.** Its end
   anchor matched the separator literally (`", "items"` / `","items"`), so a
   multi-line malformed envelope fell through to "everything up to the last
   quote" and dragged the whole items array into the summary — the same
   scaffolding leak the function exists to prevent, one whitespace variation
   away. It now finds the structural `"items"` key and walks back to the last
   quote before it.
3. **The fallback is now honest and observable.** A JSON object with no
   readable summary (no `"summary"` key, or an empty one) still landed
   verbatim, and that failure is self-perpetuating: a stored blob satisfies the
   "already has a summary" gate, so the recording is never retried. It now
   yields *no* summary — the same treatment empty CLI output has always had —
   so the next backfill sweep picks it up. `parseSummaryAndItems` also reports
   which shape arrived (`sentinel` / `envelope` / `salvaged-envelope` /
   `plain` / `unreadable-envelope`) and both log lines carry it. It is a fixed
   keyword, never the model's text: summary output is derived from the
   transcript and is user content, which must not reach a `.public` log field.

Also: the Settings prompt editor now says Mila asks for the action items
alongside the summary, since the appended contract isn't visible in the box
the user is editing.

**Persisted contracts:** nothing to mirror. `Recording.encode(to:)` and
`ActionItem` are untouched — the change is what *value* lands in
`actionItems`, not its shape — and `MilaKit.StoredRecording.ActionItem`
already carries all six fields including `source`.

### One design question for you, not changed here

The one-shot prompt is not shown the recording's existing items, so the model
cannot consolidate or update them the way the live tick does (`encodeForPrompt`
feeds the current list back in). Preserving dictated items covers the data-loss
half; feeding the list in would let the model merge instead of re-derive. Left
alone deliberately — it is a prompt-design change, not a bug fix.

## How it was tested

- **Not built or run.** There is no Xcode, no `xcodebuild` and no Swift
  toolchain on the machine this relay was prepared on, so the code was verified
  by reading it against the real call sites (`QuickActionsController.stopTail`
  → `regenerate`, `LiveAISession.parseEnvelope` / `parseActionItems` /
  `makeItem`, `LiveAISettings.summaryPrompt`'s `didSet`,
  `AIOverviewSection`/`RecordingDetailView` item rendering) and by checking
  every symbol referenced actually exists. CI is the first compile.
- @ValeroK reports `RecordingSummarizerTests` + `LiveAISettingsMigrationTests`
  green on #200 (36/36).
- **Tests that fail on `main`** — his: a one-shot response yielding populated
  action items (`test_summarize_parses_envelope_and_stores_summary_and_action_items`,
  `test_summarize_parses_sentinel_format`) and the Hebrew malformed-envelope
  salvage. On `main` the raw output is stored verbatim and `actionItems` is
  never written, so all three fail there.
- **Negative controls** — mine: `test_parse_salvages_pretty_printed_malformed_envelope_without_leaking_scaffolding`
  reproduces the old literal anchor in the test file and asserts it *leaks*, so
  the assertion isn't vacuous; `test_summarize_never_stores_an_unreadable_json_object_as_the_summary`
  feeds a JSON object the parser can't read and asserts the summary stays nil —
  specifically not a raw `{...}` blob — with the retry gate still open.
- No test on `main` pinned the buggy behaviour (all 18 in
  `RecordingSummarizerTests` and the 4 migration tests checked), so nothing had
  to be un-asserted.
- Not verified: no manual end-to-end run with a real CLI and a real Hebrew
  transcript.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — could not be
      run in the relay environment; CI is the first build

---

---

> [!NOTE]
> **Medium Risk**
> Touches the core post-recording summarization path and persisted `summary`/`actionItems` data, with complex LLM output parsing; risk is mitigated by extensive tests and conservative fallbacks (no blob storage, preserve dictated items).
> 
> **Overview**
> Restores **summary + action items** from the post-recording one-shot LLM pass (behavior lost after #87) and stops malformed model output from being saved as the recording summary.
> 
> **`RecordingSummarizer`** appends an output contract at send time (`promptWithActionItems`): plain-text summary, then `###ACTION_ITEMS###`, then a JSON array—so long Hebrew/multiline summaries are not embedded in JSON. CLI output is split via `parseSummaryAndItems` (sentinel → envelope → salvage broken JSON → prose). Unreadable JSON objects yield **no** summary (retry/backfill still possible) instead of a permanent `{...}` blob. Parsed items update `actionItems` when non-empty; empty lists keep existing items. **`merged(keepingDictatedFrom:)`** keeps voice-command items ahead of inferred ones on `regenerate`.
> 
> **`LiveAISettings`** ships a richer default summary prompt and migrates users still on the old concise default (`legacySummaryPromptV1`); custom prompts are unchanged. **Settings** copy notes that action items are requested alongside the summary.
> 
> Large test additions cover parsing edge cases (fenced JSON, salvage, dictated-item merge).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c73f45adad545fa9d9295d352bf52cdc54c1bc11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Mila/Actions/RecordingSummarizer.swift
+++ b/Mila/Actions/RecordingSummarizer.swift
@@ -581,9 +581,11 @@ final class RecordingSummarizer: ObservableObject {
     ///   2. A JSON envelope `{"summary":...,"items":[...]}` the model may have
     ///      returned out of habit — decoded via `LiveAISession.parseEnvelope`.
     ///   3. Malformed JSON that LOOKS like an envelope (the failure the user
-    ///      hit: unescaped quotes broke the object decode). Salvage the
-    ///      summary text so a raw `{...}` blob is never shown, and keep any
-    ///      items the lenient array parse recovered.
+    ///      hit: unescaped quotes broke the object decode) — bare, inside
+    ///      ```` ```json ```` fences, or behind a one-line label, since
+    ///      `parseEnvelope` accepts all three. Salvage the summary text so a
+    ///      raw `{...}` blob is never shown, and keep any items the lenient
+    ///      array parse recovered.
     ///   4. Plain prose — the whole output is the summary, no items.
     ///
     /// A JSON object that reaches step 3 and yields nothing readable returns
@@ -603,10 +605,11 @@ final class RecordingSummarizer: ObservableObject {
         if !parsed.summary.isEmpty {
             return (parsed.summary, parsed.items, .envelope)
         }
-        // Object decode produced no summary. If it still looks like an
-        // envelope, salvage the summary value rather than surfacing the blob.
-        if trimmed.hasPrefix("{") {
-            let salvaged = Self.salvageSummaryValue(from: trimmed)
+        // Object decode produced no summary. If the output still IS an
+        // envelope — bare, fenced, or behind a label — salvage the summary
+        // value rather than surfacing the blob.
+        if let envelope = Self.envelopeBody(in: trimmed) {
+            let salvaged = Self.salvageSummaryValue(from: envelope)
             if !salvaged.isEmpty {
                 return (salvaged, parsed.items, .salvagedEnvelope)
             }
@@ -625,7 +628,72 @@ final class RecordingSummarizer: ObservableObject {
             // lenient array parse recovered are still returned.
             return ("", parsed.items, .unreadableEnvelope)
         }
+        // Genuine prose. Items stay empty: every shape from which the lenient
+        // parse can actually recover an items array is an envelope, and those
+        // now route through the branch above instead of falling to here.
         return (trimmed, [], .plain)
+    }
+
+    /// The envelope-shaped payload inside the wrappers
+    /// `LiveAISession.parseEnvelope` accepts — a ```` ```json ```` fence, or a
+    /// one-line "Here is the JSON:" label — or nil when the output is prose
+    /// that merely CONTAINS a brace.
+    ///
+    /// Why this exists: the salvage and `unreadableEnvelope` branches were
+    /// gated on `trimmed.hasPrefix("{")`, but `parseEnvelope` does not require
+    /// its object to be at position 0 — it plucks the first balanced block out
+    /// of the string (`findFirstJSONStructure`) precisely because "the CLI
+    /// occasionally wraps output in prose or ```json fences". A fenced or
+    /// prefixed BROKEN envelope therefore failed the object decode *and* the
+    /// prefix test and landed in `.plain`, which stores the raw blob as the
+    /// summary — the exact bug this fix exists to remove, and the
+    /// self-perpetuating one, since a stored blob satisfies
+    /// `shouldSummarize`'s "already has a summary" gate and is never retried.
+    /// Fences are the single most likely shape for an LLM to emit, so the
+    /// promise held only for the tidiest case.
+    ///
+    /// Deliberately NOT "contains a `{`": a legitimate prose summary can quote
+    /// braces, and blanking a good summary is a worse outcome than showing the
+    /// blob. The object must be what the output IS, not something it mentions
+    /// — so only a short single-line label may precede it and nothing but
+    /// whitespace may follow it.
+    static func envelopeBody(in trimmed: String) -> String? {
+        var body = trimmed
+        // Unwrap a ``` / ```json fenced block. Matched wherever it starts, not
+        // just at position 0, because a chatty model puts a label line ahead
+        // of the fence — the commonest shape of all.
+        if let opening = body.range(of: "```") {
+            // Only a short label may precede the fence and nothing but
+            // whitespace may follow it; otherwise this is prose that quotes a
+            // code block, not an envelope in a wrapper.
+            guard body[..<opening.lowerBound].count <= 80 else { return nil }
+            let afterOpening = body[opening.upperBound...]
+            // Drop the info string (`json`) up to the end of the fence line.
+            var inner = afterOpening
+            if let newline = afterOpening.firstIndex(of: "\n") {
+                inner = afterOpening[afterOpening.index(after: newline)...]
+            }
+            if let closing = inner.range(of: "```") {
+                guard inner[closing.upperBound...]
+                    .trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+                else { return nil }
+                inner = inner[..<closing.lowerBound]
+            }
+            body = inner.trimmingCharacters(in: .whitespacesAndNewlines)
+        }
+        // The object IS the payload. Accepted without a balanced-block check,
+        // because the malformed input this path exists for can defeat the
+        // depth counter (an odd number of unescaped quotes leaves `inString`
+        // inverted) while `salvageSummaryValue` still reads the summary fine.
+        if body.hasPrefix("{") { return body }
+        guard let range = LiveAISession.findFirstJSONStructure(
+            in: body, opening: "{", closing: "}"
+        ) else { return nil }
+        guard body[..<range.lowerBound].count <= 80,
+              body[range.upperBound...]
+                .trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        else { return nil }
+        return String(body[range])
     }
 
     /// Which branch of `parseSummaryAndItems` produced the result. Logged so
--- a/Mila/Actions/RecordingSummarizer.swift
+++ b/Mila/Actions/RecordingSummarizer.swift
@@ -652,48 +652,75 @@ final class RecordingSummarizer: ObservableObject {
     /// Fences are the single most likely shape for an LLM to emit, so the
     /// promise held only for the tidiest case.
     ///
-    /// Deliberately NOT "contains a `{`": a legitimate prose summary can quote
-    /// braces, and blanking a good summary is a worse outcome than showing the
-    /// blob. The object must be what the output IS, not something it mentions
-    /// — so only a short single-line label may precede it and nothing but
-    /// whitespace may follow it.
+    /// Unwrapping is an OPTIMISTIC normalisation, judged separately from the
+    /// envelope/prose decision, and it must never be able to make the parse
+    /// worse than not having attempted it. So: if a fence is confidently
+    /// identifiable and what it wraps is envelope-shaped, that is the payload;
+    /// otherwise the ORIGINAL text is judged on its own merits. An
+    /// unrecognised wrapper — an over-long prefix, content after the closer,
+    /// no closer at all, or a shape neither of us has thought of — costs
+    /// nothing beyond the failed attempt.
+    ///
+    /// The first cut of this stripped the fence *before* the shape test and
+    /// returned nil when the wrapper looked wrong, which re-broke the very
+    /// case it was added for: a malformed envelope whose summary TEXT mentions
+    /// ```` ``` ```` (a meeting that discussed code) was hijacked by the fence
+    /// branch and aborted, though its leading `{` alone should have been
+    /// enough to salvage it.
     static func envelopeBody(in trimmed: String) -> String? {
-        var body = trimmed
-        // Unwrap a ``` / ```json fenced block. Matched wherever it starts, not
-        // just at position 0, because a chatty model puts a label line ahead
-        // of the fence — the commonest shape of all.
-        if let opening = body.range(of: "```") {
-            // Only a short label may precede the fence and nothing but
-            // whitespace may follow it; otherwise this is prose that quotes a
-            // code block, not an envelope in a wrapper.
-            guard body[..<opening.lowerBound].count <= 80 else { return nil }
-            let afterOpening = body[opening.upperBound...]
-            // Drop the info string (`json`) up to the end of the fence line.
-            var inner = afterOpening
-            if let newline = afterOpening.firstIndex(of: "\n") {
-                inner = afterOpening[afterOpening.index(after: newline)...]
-            }
-            if let closing = inner.range(of: "```") {
-                guard inner[closing.upperBound...]
-                    .trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
-                else { return nil }
-                inner = inner[..<closing.lowerBound]
-            }
-            body = inner.trimmingCharacters(in: .whitespacesAndNewlines)
+        if let unwrapped = Self.unfencedBody(in: trimmed),
+           let body = Self.objectPayload(in: unwrapped) {
+            return body
         }
-        // The object IS the payload. Accepted without a balanced-block check,
-        // because the malformed input this path exists for can defeat the
-        // depth counter (an odd number of unescaped quotes leaves `inString`
-        // inverted) while `salvageSummaryValue` still reads the summary fine.
-        if body.hasPrefix("{") { return body }
-        guard let range = LiveAISession.findFirstJSONStructure(
-            in: body, opening: "{", closing: "}"
-        ) else { return nil }
-        guard body[..<range.lowerBound].count <= 80,
-              body[range.upperBound...]
+        return Self.objectPayload(in: trimmed)
+    }
+
+    /// The contents of a ``` / ```json fenced block. Nil means only "no
+    /// wrapper I can be confident about" — NOT "not an envelope"; the caller
+    /// falls back to the unmodified text, never to a failed parse.
+    private static func unfencedBody(in trimmed: String) -> String? {
+        guard let opening = trimmed.range(of: "```") else { return nil }
+        // Only a short label may precede the fence, and nothing but
+        // whitespace may follow its closer. Otherwise the backticks are far
+        // more likely part of the content than a wrapper around it.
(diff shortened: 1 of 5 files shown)


urisland:
### Bugbot finding "Wrapped JSON blobs stored as summaries" — real, fixed in 9e8a053

Verified against the code before acting, and it is correct on all three counts:

1. **The `{`-prefix condition gated both paths.** `salvageSummaryValue` *and* the `unreadableEnvelope` return were inside the same `if trimmed.hasPrefix("{")` block.
2. **`parseEnvelope` does document the wrappers, and it never strips them.** Its doc comment says "the CLI occasionally wraps output in prose or ```json fences, so we extract the first balanced JSON block we find" — it locates the object with `findFirstJSONStructure` rather than removing the wrapper from the string. So my prefix test ran against the still-wrapped output and there was no stripping, before or after.
3. **Yes, items were discarded.** The fall-through was `return (trimmed, [], .plain)` — a hardcoded `[]`, dropping whatever the lenient array parse had recovered.

So the PR's promise held only for envelopes beginning with `{`, and failed for a fenced or label-prefixed one — the shape a model emits most often. The retry half is the worse part, exactly as described: a stored blob satisfies `shouldSummarize`'s "already has a summary" gate, so the recording is never retried and the blob is what the detail view, the Obsidian note and the MCP surface report forever.

**Fix:** a new `envelopeBody(in:)` unwraps the wrappers *before* the shape test — a ``` / ```json fence (matched wherever it starts, since the label line usually precedes it) and/or a leading label.

One deliberate limit, worth flagging because the obvious version of this fix is wrong: it is **not** "the output contains a `{`". A real summary can quote braces or a fenced code block, and blanking a good summary is a worse outcome than showing a blob. The object therefore has to be what the output *is*, not something it mentions — at most a short (≤80 char) label before it, nothing but whitespace after it. A bare leading `{` is still accepted without the balanced-block check, because an odd number of unescaped quotes can leave `findFirstJSONStructure`'s `inString` inverted while `salvageSummaryValue` still reads the summary fine.

**Tests** (all four fail on `5abd335`, not merely on `main`):

- `test_parse_salvages_a_malformed_envelope_inside_fences_and_a_label` — fenced and label-only variants salvage, with no fence or `"items"` scaffolding in the summary and the recovered item intact.
- `test_summarize_never_stores_a_fenced_unreadable_json_object` — end-to-end: no summary stored and `shouldSummarize` stays `true`.
- `test_parse_keeps_prose_that_merely_quotes_braces_or_a_code_block` — the two counterweights, so the fix cannot be "blank anything with a brace in it".

No other `hasPrefix("{")` gate remains in `Mila/`. Nothing about what gets persisted changed, so `StoredRecording` and `StoredRecordingDriftTests` are still unaffected.


---

urisland:
### Bugbot round 2, "Backticks abort envelope salvage" — real, and my own regression. Fixed in db22619

Correct, and worth being blunt about what happened: this is round 1's bug reappearing *through the fix for it*. The fence handling I added stripped the wrapper **before** the shape test and returned `nil` when the wrapper looked wrong — over-long prefix (L669) or content after the closer (L679) — and both of those `return nil`s sit above the `hasPrefix("{")` acceptance. So a malformed envelope whose summary **text** mentions ``` had its parse hijacked by the fence branch and aborted, landing in `.plain` as a raw blob with the retry gate shut, even though its leading `{` alone had always been enough to salvage it.

So I fixed the class rather than the instance: **unwrapping is an optimistic normalisation and must never make the parse worse than not having attempted it.**

- `unfencedBody(in:)` reports the fenced contents, or `nil` meaning only "no wrapper I can be confident about".
- `envelopeBody(in:)` falls back to judging the **original** text whenever the unwrapped candidate is absent or not envelope-shaped.

The worst outcome becomes "salvage tries and fails on text it could not have parsed anyway" — the pre-existing behaviour — instead of a fresh blob. Every unrecognised-wrapper case is covered at once: over-long prefix, trailing content, no closer at all, and shapes nobody has enumerated.

**One correction to the suggested patch, since it matters.** The fail-open could not be pushed all the way down as "`envelopeBody` never returns `nil`". I checked the call site first, and it does depend on `nil` meaning "definitely not an envelope":

```swift
if let envelope = Self.envelopeBody(in: trimmed) { … return (…, .salvagedEnvelope / .unreadableEnvelope) }
return (trimmed, [], .plain)          // ← the only route to .plain
```

A blanket accept would send **every plain-prose summary** into the salvage branch, where `salvageSummaryValue` finds no `"summary"` key, returns `""`, and the summary gets blanked as `.unreadableEnvelope` — turning a cosmetic bug into total data loss on the normal path, and breaking `test_parse_origin_reports_which_shape_arrived`. So the discriminator keeps its `nil` in `objectPayload(in:)`, and the fallback is expressed one level up in `envelopeBody` — the call-site expression you flagged as the alternative. `parseEnvelope`'s contract and the origin labelling are untouched, so this needed no escalation.

**Tests** (both fail on `9e8a053`, which is the point):

- `test_parse_salvages_a_malformed_envelope_whose_summary_mentions_backticks` — the exact shape named, salvaged with `` ```retry``` `` preserved in the text and the item recovered.
- `test_summarize_never_stores_a_blob_whose_text_mentions_backticks` — end-to-end: no summary stored, `shouldSummarize` stays `true`.
- The two prose counterweights still hold, so the fix cannot have drifted into "blank anything with a brace in it".

Note for the record: the macOS CI runners were backlogged, so `9e8a053` never got a compile signal before this push — one run now covers both commits.


---

urisland:
### Bugbot round 3, "Unclosed fence stores raw summary blob" — real, mine again. Fixed in 67463de

Right, and it is the previous commit's own principle applied incompletely rather than a new defect. I said unwrapping must never fail closed, then left two fail-closed vetoes in `unfencedBody`:

```swift
guard let closing = inner.range(of: "```"),
      inner[closing.upperBound...].trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
else { return nil }        // ← veto on a missing closer, AND on trailing content
```

Round 1's version had no veto for the missing closer — it kept `inner` as-is, which preserved the leading-`{` bypass — so the round-2 refactor regressed that specific path. Bugbot's mechanism is exactly right, including why it bites hardest on a **labelled** envelope: that is the case with no leading `{` of its own to fall back on, and `findFirstJSONStructure` gives up on the odd unescaped quote this salvage path exists for, so both fallbacks are gone at the same time.

`unfencedBody` now only ever **offers** a candidate. A missing closer keeps everything after the fence line; content after a closer is dropped rather than vetoing. `envelopeBody` already re-tests the original text when the candidate is not envelope-shaped, so being generous here cannot cost anything.

That also closes the trailing-sign-off shape ("Hope that helps!" after the closer), which was never a regression — it behaved this way in round 1 and on the original head too — but is fixed by the same change rather than needing its own case.

**The one restriction I kept, deliberately:** the ≤80-char label guard before the fence. It is load-bearing in the other direction, and I have documented it as such. Without it, a long prose summary that quotes a fenced JSON object would have that object substituted for it and then be blanked — losing a good summary, which is the worse outcome. The two prose counterweight tests pin that behaviour, and I re-checked both against this change.

**Test:** `test_parse_salvages_a_wrapped_envelope_when_the_fence_is_unclosed_or_trailed` covers both veto shapes. Each payload carries an **odd** number of unescaped quotes, which leaves `findFirstJSONStructure`'s `inString` inverted so the balanced-block search fails outright — that is what makes the fence candidate the only route to the bypass, and the assertions discriminating rather than incidentally true. Fails on `db22619`.

For the record on this parser, three rounds in: rounds 2 and 3 were both my own regressions introduced while fixing round 1, and both were the same shape — a normalisation step that could veto a payload the next step would have handled. The invariant is now stated in the code where the veto used to be, which is the only place it would have been read.

Still true that nothing here changes what gets persisted, so `StoredRecording` and `StoredRecordingDriftTests` remain unaffected. Also noting for the record: CodeRabbit posted only a rate-limit notice on this PR (free OSS tier, 2 reviews/hour exhausted), so there is no CodeRabbit review to address — Bugbot is the reviewer of record here.


---

urisland:
### Bugbot round 4 — both findings real, both mine, fixed in c73f45a

Both are consequences of the previous commit removing the trailing-content veto, and both point at the same thing I got wrong: I wrote that being generous in the unwrapper "cannot cost anything". That is false, and it is now corrected in the code where I wrote it.

**1. "Fence unwrap blanks prose summaries"** — confirmed, including the part about my own test. A short lead-in plus a quoted fenced object wins the unwrap, `objectPayload` accepts any leading `{`, salvage finds no `"summary"` key, and `unreadableEnvelope` **blanks a real prose summary** — discarding the prose on both sides of the block:

```
Notes:
```json
{"retries": 3}
```
The team agreed to raise it next sprint after the incident review.
```

And the observation that my counterweight fixture "is saved by that long prefix, not by the closer check" is correct — that fixture was weaker than I claimed. The new test uses a deliberately **short** lead-in so the 80-char guard cannot be what saves it.

The asymmetry I missed: a candidate that is object-shaped but has no summary is not merely *rejected*, it is **blanked**. So unwrapping too eagerly destroys data, while unwrapping too timidly only shows a blob. The trailing-content veto is therefore restored — content after the closer means the block is material *quoted inside* a summary, not a wrapper around the output.

**2. "Inner backticks truncate fenced salvage"** — confirmed, and this one is the worst outcome in the whole PR. The closer was matched anywhere, so a malformed envelope whose summary mentions ``` was clipped at the first inner run and salvage stored `We fixed the` — a short, plausible-looking **wrong** summary. Worse than the blob it replaced, exactly as the finding says: nothing about it looks broken, and the retry gate shuts on it just the same.

**The concept missing behind both: a fence delimits a block only at the START of a line.** Backticks mid-line are content. `lineLeadingFence(in:from:)` now finds openers and closers that begin a line, so an inner run can neither open a block nor close one. That is also what makes the round-2 case (`` ```retry``` `` inside a bare `{…}` envelope) safe *structurally* rather than by luck of the shape test.

Kept from round 3: a **missing** closer is still not a veto — that case is unambiguous, there is no trailing prose to lose.

**One test removed rather than adjusted,** and I want that visible: the round-3 "trailed fence" assertion, which said a chatty sign-off after the closer should win the unwrap. That is precisely what sacrifices the prose in finding 1, and it was never a regression — that shape stored a blob on the original head too. I had added it as a "fix" and it was wrong.

New tests, both failing on `67463de`: `test_parse_keeps_prose_that_continues_after_a_quoted_fenced_object` and `test_parse_does_not_let_inner_backticks_truncate_a_fenced_salvage` (its fixture carries unescaped quotes so the decode genuinely fails and the salvage is actually reached — otherwise `parseEnvelope` succeeds and `envelopeBody` is never called, which would have made the test vacuous).

**Standing note on this parser, four rounds in:** rounds 2, 3 and 4 were all my own regressions while fixing round 1, and the tension is now explicit in the code — eager unwrapping blanks good summaries, timid unwrapping shows blobs, and only the *envelope-shaped, nothing-else-around-it* case is safe to take. If a fifth round finds another shape here, my recommendation to the maintainer is to stop patching the tolerant parser and make the prompt's sentinel contract the only accepted format, treating everything else as `.plain` — the salvage exists for output that predates the contract, and it has now cost more than it protects. Flagging rather than doing it, as it changes behaviour well beyond #199.

Still no change to what gets persisted, so `StoredRecording` / `StoredRecordingDriftTests` remain unaffected.


---

urisland:
### Status: green at `c73f45a`

**CI:** all 16 checks pass — `unit-and-ui-tests`, `build-and-test`, `Analyze (swift)`, CodeQL, `diarize`, `e2e`, `remote-transcription-e2e`, `UI layout tests`, both mock-meeting jobs, `Linked issue`; `mila-client` / `speaches-server` skipped by path filter.

One re-run was needed: the first pass failed only `MicrophoneRecorderTests.test_timeout_fires_even_when_bring_up_is_not_cancellation_responsive()`. That is a timing test in the microphone bring-up path, nothing this PR touches, and it passed on re-run — consistent with the known flakiness tracked in #208. All 1094 tests pass, including every summarizer test added here.

**Cursor Bugbot:** clean at this head — "Bugbot completed review — no issues found", and all five threads from the earlier rounds are resolved. Four rounds of findings, all verified against the code and all fixed; the reasoning for each is in the comments above and in the commit messages.

**CodeRabbit:** posted only a rate-limit status (free OSS tier), so there is no CodeRabbit review on this PR.

Not merging — leaving that to a maintainer. The one deliberate follow-up is still open and described in the PR body: the one-shot prompt is not shown the existing action items, so the model cannot consolidate them the way the live tick does. And the standing recommendation from the round-4 comment holds — if this tolerant parser draws further findings, the better move is to make the sentinel contract the only accepted format rather than keep patching the salvage.

Credit for the original diagnosis and fix remains with @ValeroK (#200, closed as superseded); commit `23abe76` here is theirs, unmodified.


---

PR #229, opened 2026-08-31
feat(live): delete individual lines from the live transcript (relay of #125)

### LINKED ISSUE #222: No way to remove a bad line from the live transcript while recording
## What's missing

While a recording is running, the live transcript pane is read-only. Every line whisper emits stays there until the recording ends.

That is a problem in three distinct situations, and they are not cosmetic:

- **Whisper misfires.** A cough, a door, an air-conditioning hum that cleared the RMS gate — whisper does not return "nothing", it returns words. A hallucinated phrase during silence looks exactly like something a participant said.
- **Cross-talk.** A snippet of a neighbouring conversation, or someone else's call bleeding through the mix, lands in the transcript of *this* meeting.
- **Something said off the record.** A participant asks for a remark not to be minuted, or says something personal before remembering they are being recorded. Right now the only remedy is to let it be transcribed, saved, summarised and exported, and then go and edit the saved transcript afterwards.

The third case is the one that matters most, and it is why this deserves doing properly rather than as a UI convenience: today the user has no way to act at the moment they want to act.

## Why "fix it afterwards" isn't a workaround

The live transcript is not just a display — it is fed onward *while the recording is still going*:

- `LiveAISession` is fed `formattedTranscript` on every tick, so the line reaches the LLM summary and action items.
- `LiveTranscriptSidecarWriter` mirrors every tick to `live/current.json`, which the `mila-mcp` helper serves through `get_live_transcript` — so an external Claude session following the meeting has already read it.
- The mid-recording "Export SRT…" button snapshots whatever is in the pane.
- At Stop, the live segments become the saved `Recording` (in auto-segment mode they *are* the authoritative transcript), which is what `recordings.json`, the `.txt` sidecar, the `.srt` and the summary are built from.

By the time the recording ends the line has been copied into several places. Removing it at the source is the only thing that keeps it out of them.

## Intended shape

A per-line delete affordance in the live transcript pane — a trash control revealed on hover, plus a right-click "Delete line" item — backed by a `LiveTranscriber.removeSegment(id:)` that removes the segment and recomputes `fullText`. Because `segments` is `@Published`, the removal propagates to the sidecar, the LLM feed and the Stop-time snapshot for free.

The non-obvious part, and the reason this is not a one-line array removal: the fixed-window (non-VAD) path **re-transcribes a rolling buffer on every tick**, so a deleted line reappears a second later unless its absolute time range is remembered and incoming output overlapping it is suppressed.

## What "done" has to mean here

`bugbot-rules/deleted-data-stays-deleted.md` applies directly — this is a privacy action, not a cache eviction, so after the delete no code path may write the line back. Each of these needs checking, not assuming:

- the live segment array, across a subsequent re-transcription tick;
- `live/current.json`, **including the incremental `get_live_transcript` cursor** — it is append-only by construction (a count plus a "re-read the last entry" rule), so a poller that already fetched the line has to be told, or it keeps it and its cursor drifts;
- the `.txt` transcript sidecar and `recordings.json` written at Stop;
- the `.srt` export, and the mid-recording SRT export;
- the automatic post-Stop batch pass — in chunk mode `finalizeTail` enqueues the WAV and the pass overwrites the saved segments wholesale;
- the in-flight Live-AI summary payload.

Two things also want an explicit decision: whether a mis-tap should be undoable, and whether the affordance should appear at all in modes where the deletion would not survive Stop.

## Credit

The feature was written by **@ValeroK** in #125, which predates this issue — that PR is where the design and the suppression logic came from. This issue exists because every PR needs a linked issue (`.claude/rules/pull-requests.md`), and because the durable, searchable record of *what was wrong and why* belongs somewhere other than a PR description.

urisland:
Closes #222

## Credit: this feature is @ValeroK's work

**Deleting a line from the live transcript was designed and written by @ValeroK in #125.** Both of his commits are cherry-picked unmodified onto today's `main`, so he is the `Author` on each of them; everything I added sits in three separate commits on top and is a guard, a protocol fix and tests — not the feature.

The cherry-pick was **clean — zero conflicts** — even though #125 branched from `53ddcfe` and only nine files have changed in `main` since. Nothing of his was rewritten, reflowed or reimplemented. `git diff` between his two commits as relayed here and `origin/main...pr/125` is byte-identical.

It is relayed rather than merged for the usual reason (the #129 → #164 and #198 → #204 precedents): a fork PR can't get an automated review under the contributor's own quota, and its CI needs per-run approval on every run. The relay is purely mechanical. Please close #125 as superseded by this once it's green.

> His commit email (`kobi.valero@pagaya.com`) is a real address rather than a placeholder, so it is preserved verbatim as the `Author` on his commits. My three commits additionally carry `Co-authored-by: KobiValero-Pagaya <8961499+ValeroK@users.noreply.github.com>` so the attribution resolves to the `@ValeroK` handle as well.

## What & why

Each line in the live transcript pane gets a delete affordance — a trash button revealed on hover plus a right-click "Delete line" item. `LiveTranscriber.removeSegment(id:)` records the segment's absolute time range in `deletedRanges`, removes it from `segments` and recomputes `fullText`; incoming whisper output overlapping a deleted range is then skipped.

That suppression is the load-bearing part, and it is why this isn't just an array removal: the fixed-window path **re-transcribes a rolling buffer on every tick**, so without it the deleted line reappears a second later. (The VAD path emits each utterance once, so there it's a safety net.) His follow-up commit also clamps a VAD segment's `absEnd` to the *unpadded* audio duration, so deleting a line whose whisper end ran into the trailing silence pad cannot suppress the next real utterance.

## Where a deleted line could have survived, and what I found

Deletion is a privacy action (`bugbot-rules/deleted-data-stays-deleted.md`), so I traced every path the live transcript is copied into rather than trusting the in-memory array. Read the two **`SURVIVED`** rows as the review, not as nits.

| Path | Deleted line survives? |
|---|---|
| `LiveTranscriber.segments`, across a later tick | **No.** `isSuppressed` is applied in both `runOnce` (fixed-window, before `merge`) and `transcribeUtterance` (VAD, before append). |
| `live/current.json` — file contents | **No.** `MilaApp`'s feed loop maps `transcriber.segments` into `sidecarWriter.update(...)` on every `$segments` publish; the shorter list differs, so the revision bumps and it is rewritten atomically. |
| `live/current.json` — **incremental MCP read** | **SURVIVED.** Fixed in this PR — see below. |
| `.txt` transcript sidecar (written by `store.update` at Stop) | **No**, in auto-segment mode. `stopRecording` builds `finalTranscriptSegments` from the post-delete `segments`. Asserted through `MilaStoreReader.transcriptText`, which reads this sidecar *first*. |
| `recordings.json` | **No**, same path. Asserted against the raw JSON text. |
| `.srt` sidecar (`finalizeTail` → `TranscriptExporter.writeSRT(for: updated, …)`) | **No** — built from `updated.segments`. |
| Mid-recording "Export SRT…" button | **No** — it snapshots `transcriber.segments`. |
| `MilaStoreReader.transcriptText` / `namedTranscript` | **No.** |
| MCP `search_transcripts`, `get_transcript(id:)` | **No.** `recording(id:)` keeps its existing trashed filter, untouched. |
| Summary / LLM payload | **No** for every *new* payload: the feed sends post-delete `formattedTranscript`, and `finalizeTail`'s `regenerate` / `summarizeIfNeeded` read `recording.fullText`. See the caveat below. |
| **Automatic post-Stop batch pass (chunk mode)** | **SURVIVED.** Fixed in this PR — see below. |

### SURVIVED 1 — the automatic batch pass brought the line back (chunk mode)

When auto-segment (VAD) is **off**, `liveTranscriptIsAuthoritative` is false, the row is saved `.pending`, and `finalizeTail` calls `transcription.enqueue(updated)`. `TranscriptionService.process` then re-transcribes the whole WAV and overwrites `working.segments` / `working.fullText` before merging onto the live row — so the deleted line came back into `recordings.json`, the `.txt` sidecar and the `.srt`, along with a whole fresh transcript. That is exactly the shape the rule describes: an automatic completion path writing back data the user deleted while it was in flight, and the guard belongs on that path rather than on the store primitive.

**Fix (commit 3):** the delete affordance is only offered when the live transcript is the one that gets saved. `liveTranscriptIsSaved` in `LiveAIRecordingView` is deliberately the *same expression* the controller calls `vadActive`, not a second rule that could drift. In chunk mode the entire displayed transcript is provisional, so there is nothing to edit either way — and a destructive control that silently undoes itself is worse than no control.

**Maintainer decision, not mine to make:** the alternative is to plumb the deleted time ranges into the batch pass so the deletion survives re-transcription in *both* modes. I didn't do it — it needs the ranges persisted to outlive a relaunch, and it would have to *not* apply to a user-requested "Re-transcribe" — and smuggling a change that size into a relay seemed wrong. Say the word and it's a follow-up.

### SURVIVED 2 — a live MCP poller was never told the line went away

`get_live_transcript`'s delta is append-only by construction: `next_segment_index` is a count, and `segments(sinceIndex:)` returns from one before it so the client can replace a trailing segment whisper rewrote. Nothing in that vocabulary can say "the line you already hold is gone".

Concretely: a poller holding three lines asks `since_segment_index: 3` against a now-two-entry list, gets `Array(segments[min(2, 2)...])` = `[]`, and reads that as *"changed, nothing new"*. It keeps the deleted line in its own context for the rest of the meeting — and its cursor is now one ahead of reality, so every line after the deletion is mis-stitched too. Both halves are new with this feature; `segments` only ever grew before.

**Fix (commit 4):** `LiveTranscriptSnapshot.segmentsRemovedAtRevision` records the revision that carried a removal. A poll whose `since_revision` predates it — or which sends no revision, and so cannot prove it saw the removal — gets the **complete** current set plus `segments_removed: true`, and the tool description now tells the client to discard rather than append. Once the cursor is at or past that revision it is back on the cheap delta: a delete must not turn every remaining poll into a full resend.

Detection is precise in both directions, which is why it isn't a count check. `segmentsWereRemoved(from:to:)` compares start/end/text over the prefix *before* the trailing entry, so a mid-list delete is caught even when a new utterance lands in the same tick and the count is unchanged; a trailing rewrite is not flagged (the protocol already re-sends that entry, which is also why deleting the **last** line needs no marker); and a late speaker label landing on an already-published line is not flagged, or a multi-speaker meeting would full-resend on most ticks.

The new field is `Int?` on purpose. `live/current.json` is a cross-process *and* cross-version contract (`bugbot-rules/persisted-contracts-and-mirrors.md`): a non-optional `Int` would make a newer helper throw on an older app's snapshot, `read` would swallow it as `nil`, and `get_live_transcript` would answer "not recording" in the middle of a live meeting. `docs/mcp.md` is updated.

### Also checked

- **The stop-recording handoff is unaffected.** `finish(recordingID:transcriptIsFinal:)` is still published only after `store.update` and only when it reported success. Nothing here moves that ordering, and `LiveSidecarHandoffOrderingTests` still covers it. The failure it guards against — a handoff naming a row whose sidecar holds *older* text — is precisely what a delete applied to the view but not to `store.update` would have produced, so I checked it explicitly rather than assuming.
- **Logging is clean.** `removeSegment` logs only `start`, `end` and the remaining count, all `.public` — no transcript text. Per `bugbot-rules/no-user-content-in-logs.md` those are purely diagnostic, so redacting them would be over-redaction. My new sidecar log line is the same shape (revision + count). Nothing here collides with the log-privacy sweep in #218.
- **`StoredRecording` needs no mirror change** — `Recording.encode(to:)` is untouched, so `StoredRecordingDriftTests` has nothing new to catch. There is no key-set drift test for `LiveTranscriptSnapshot`; if you want one, that's a good follow-up independent of this PR.
- **No test had to be un-pinned.** I searched `MilaTests/` and the MilaKit tests for anything asserting the live segment set is append-only, or that the sidecar mirrors every segment. There is nothing — so unlike the two past cases here, no fixture was quietly encoding the old behaviour.

### Two things left as maintainer decisions

1. **A mis-tap is immediate and irreversible.** No confirmation, no undo; the text is gone from `segments` and only the time range is kept (to suppress re-emission). The right-click route exists because the hover button is easy to miss, which reduces accidental *triggering* but not accidental *loss*. I deliberately did not invent a confirmation or an undo stack on the contributor's behalf — but for a mid-meeting destructive action with no recovery, an undo is worth considering.
2. **"Deleted" means deleted from the transcript, not from the recording.** The WAV/m4a still contains the audio, and the right-click "Re-transcribe" will bring the text back. That is a user *asking* for a fresh transcript rather than an automatic write-back, so I left it alone — but it's worth nobody reading "delete" as "redact".

## How it was tested

**The build is UNVERIFIED locally.** This relay was prepared on Linux with no Xcode, no `xcodebuild` and no Swift toolchain, so nothing here has been compiled or run — CI is the first compiler to see it. What I could do instead: read every touched call site, grep-confirm every symbol referenced by the new tests exists (`TestSupport.makeTempRoot` / `installFakeModel` / `isolatedRemoteSettings` / `writeStereo48kSineWav`, `RecordingStore.freshAudioURL`, `QuickActionsController.startFakeRecordingForTesting` / `awaitFinalizeTails`, `StubWhisperEngine.setDefaultCanned`, `MCPAccessGate.set`, `MilaStoreReader.recording(id:)` / `transcriptText(for:)` / `namedTranscript(for:)`), mirror the idioms of `LiveSidecarHandoffOrderingTests` and `MilaMCPToolHandlersTests` exactly, and check brace/paren balance across all eleven files. Please treat a red first run as mine, not the contributor's.

`project.yml` needs no change — the `Mila`, `MilaTests` and MilaKit targets are directory scans, so the three new files are picked up automatically. The `.xcodeproj` was not touched.

### What fails on `main`

All of it: `LiveTranscriber.removeSegment(id:)` does not exist there, so the suites don't compile. The assertions that discriminate *within* the feature are the contributor's own — delete the `isSuppressed` guard and `test_deleted_line_does_not_reappear_on_next_chunk_tick` fails; delete the `absEnd` clamp and `test_vad_padded_segment_delete_does_not_suppress_next_utterance` fails.

### Negative controls (new — `MilaTests/LiveTranscriptLineDeleteTests.swift`)

An in-memory-array assertion passes just as happily when every persisted copy still holds the deleted text, so these drive a real `stopRecording()` through `QuickActionsController` and then go looking for a distinctive marker line in all six places: the store row, `recordings.json`, the `.txt` sidecar via `MilaStoreReader.transcriptText`, `namedTranscript`, the `.srt`, and `live/current.json` — finishing with a sweep of **every readable file under the store root**, which is the assertion that fails for a delete that only ever reached the UI. A second test asserts `search_transcripts` returns `count: 0` while `get_transcript(id:)` still serves the surviving line, and a third drives the live-poll handoff end to end.

Every negative assertion is paired with a precondition (the line really was published to the sidecar; the sidecar chain really resolved to text rather than an empty string; search really can see this recording), because otherwise each of them would also pass against an empty file or a store the reader never found.

Plus `Packages/MilaKit/Tests/MilaKitTests/LiveTranscriptRemovalTests.swift` (removal detection, both directions, including the cross-version decode), four new cases in `MilaMCPToolHandlersTests`, and three in `LiveTranscriptSidecarWriterTests`.

### Not verified

- No build, no test run, no manual pass — see above.
- The hover interaction itself has had no human eyes on it. @ValeroK noted in #125 that his XCUITest runner wouldn't launch locally; I couldn't run it either, for a different reason. Worth a sanity-check before release.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — **not done, no toolchain available; see above**
- [ ] User-facing change is noted in the README changelog — N/A, entries are written per release

---

---

> [!NOTE]
> **Medium Risk**
> Changes how live transcripts are finalized and persisted across stop, batch, and MCP poll paths; mistakes could resurrect deleted text or leak it to pollers, though the PR adds targeted guards and broad integration tests.
> 
> **Overview**
> Adds **per-line deletion** in the live transcript pane when auto-segment (VAD) is on — hover trash and a destructive context menu call `LiveTranscriber.removeSegment`, which drops the segment and records **deleted time ranges** so fixed-window re-transcription cannot bring the line back.
> 
> **Stop/save behavior** now treats a user-emptied live transcript as authoritative (via `hasUserDeletedSegments`), avoiding a batch re-transcribe that would resurrect deleted text; fully clearing the transcript also clears saved **summary** and **action items** derived from that text. Delete controls stay hidden in chunk mode where the live text is not what gets saved.
> 
> **MCP live polling** gains `segmentsRemovedAtRevision` and `segments_removed: true` so stale incremental cursors get a full segment resend after a removal. Tests cover transcriber suppression, sidecar stamping, MCP handlers, and end-to-end persistence (store, sidecars, MCP tools).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ec811644096277a16d8cb0af468682ea989cc96a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete lines from live transcripts.
  - Deleted text remains removed from saved transcripts, exports, and live transcript integrations.
  - Live transcript updates now notify connected clients when previously received lines were removed.

- **Bug Fixes**
  - Prevented deleted lines from reappearing during transcription or recording finalization.
  - Clearing a transcript also removes summaries and action items based on its deleted content.

- **Documentation**
  - Documented live transcript deletion behavior for integrations.

### DIFF
--- a/Mila/Actions/LiveAISession.swift
+++ b/Mila/Actions/LiveAISession.swift
@@ -185,6 +185,16 @@ final class LiveAISession: ObservableObject {
         self.liveAISettings = liveAISettings
     }
 
+    /// Inject rolling AI output without invoking an LLM. Same idiom (and same
+    /// justification) as `LiveTranscriber.seedForTesting`: `summary` and
+    /// `actionItems` are `private(set)`, so a test asserting on what Stop
+    /// snapshots out of them has no other way in. Production paths must not
+    /// call this.
+    func seedForTesting(summary: String, actionItems: [ActionItem] = []) {
+        self.summary = summary
+        self.actionItems = actionItems
+    }
+
     /// Begin a session — clears any previous state. Generates a fresh
     /// session id when the user's LLM tool supports session continuity
     /// (Claude does; Cursor doesn't), so each recording gets its own
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -898,6 +898,27 @@ final class QuickActionsController: ObservableObject {
         updated.fullText = hasLiveSegments ? finalFullText : ""
         updated.summary = finalSummary.isEmpty ? nil : finalSummary
         updated.actionItems = finalItems.isEmpty ? nil : finalItems
+        if userEmptiedLiveTranscript {
+            // Emptying the live transcript is a DISCARD of the transcript, so
+            // the artifacts DERIVED from it go with it. Both were produced by
+            // the Live AI loop from text that no longer exists, and nothing
+            // downstream would replace them: `RecordingSummarizer.regenerate`
+            // bails on an empty `fullText` (and only ever writes `summary`,
+            // never `actionItems`), and an authoritative row owes no batch
+            // pass, so there is no completion hook either. Keeping them would
+            // leave the transcript reading clean while the deleted words sit
+            // in the `summary` field of `recordings.json` and in everything
+            // that reads the row — MCP `get_transcript` included, which serves
+            // summary + action items by default. (Cursor Bugbot on #229.)
+            //
+            // Scoped to a FULL emptying on purpose. A partial delete leaves a
+            // non-empty transcript, and dropping a meeting's whole summary
+            // because one line was removed would be its own data loss; that
+            // case is discussed on the PR (the summary is regenerated from the
+            // clean text, but `actionItems` are not).
+            updated.summary = nil
+            updated.actionItems = nil
+        }
         updated.status = liveTranscriptIsAuthoritative ? .completed : .pending
         // `update` reports whether the `.txt` sidecar AND recordings.json
         // actually landed on disk. Both used to fail silently, so the handoff
--- a/Mila/Actions/QuickActionsController.swift
+++ b/Mila/Actions/QuickActionsController.swift
@@ -789,6 +789,11 @@ final class QuickActionsController: ObservableObject {
         // handler is skipping its `transcriber.stop()` /
         // `diarizer.stop()` while `isFinalizingRecording` is true.
         let finalLiveSegments = liveTranscriber?.segments ?? []
+        // Snapshotted here for the same reason as the segments themselves —
+        // before `liveTranscriber?.stop()` below — because it is what tells an
+        // emptied live transcript apart from one that never existed. See
+        // `liveTranscriptIsAuthoritative`.
+        let userDeletedLiveLines = liveTranscriber?.hasUserDeletedSegments ?? false
         let finalSummary = (liveAISession?.summary ?? "")
             .trimmingCharacters(in: .whitespacesAndNewlines)
         let finalItems = liveAISession?.actionItems ?? []
@@ -835,7 +840,21 @@ final class QuickActionsController: ObservableObject {
         // `vadActive` here is whatever was passed in by the caller —
         // typically `liveAISettings.useVAD && liveAISettings.enabled`.
         let hasLiveSegments = !finalLiveSegments.isEmpty
-        let liveTranscriptIsAuthoritative = hasLiveSegments && vadActive
+        // A live transcript the user EMPTIED by deleting every line is not the
+        // same thing as one that was never produced, and `hasLiveSegments`
+        // cannot tell them apart on its own. It exists for the second case: no
+        // live text means the live path gave us nothing, so fall back to the
+        // batch pass. Run that fallback after a hand-emptied transcript and the
+        // batch pass re-transcribes the WAV and writes every deleted line back
+        // into recordings.json, the `.txt` sidecar and the `.srt` — the
+        // resurrection `bugbot-rules/deleted-data-stays-deleted.md` describes,
+        // reached by the one route the UI-side gate cannot cover, because
+        // emptying `segments` is itself what flips this gate. Deleting every
+        // line is the strongest statement a user can make that this must not be
+        // recorded, so a deliberate emptying stays authoritative and saves
+        // empty. (Cursor Bugbot on #229.)
+        let userEmptiedLiveTranscript = userDeletedLiveLines && !hasLiveSegments
+        let liveTranscriptIsAuthoritative = (hasLiveSegments || userEmptiedLiveTranscript) && vadActive
         let finalTranscriptSegments: [TranscriptSegment] = finalLiveSegments.map { ls in
             TranscriptSegment(start: ls.startSeconds, end: ls.endSeconds,
                               text: ls.text, speaker: ls.speaker)
--- a/Mila/Audio/LiveTranscriber.swift
+++ b/Mila/Audio/LiveTranscriber.swift
@@ -249,6 +249,15 @@ final class LiveTranscriber: ObservableObject {
         liveLog.log("LiveTranscriber.removeSegment start=\(seg.startSeconds, privacy: .public) end=\(seg.endSeconds, privacy: .public) remaining=\(self.segments.count, privacy: .public)")
     }
 
+    /// Whether the user deleted at least one line from THIS recording's live
+    /// transcript (reset by `start()` along with `deletedRanges`).
+    ///
+    /// Read at Stop, where it is the only thing that tells an *emptied* live
+    /// transcript apart from one that was never produced. Those two look
+    /// identical in `segments` and must be handled oppositely — see
+    /// `liveTranscriptIsAuthoritative` in `QuickActionsController.stopRecording`.
+    var hasUserDeletedSegments: Bool { !deletedRanges.isEmpty }
+
     /// Whether an incoming segment's absolute time range overlaps a range
     /// the user deleted. Any positive overlap counts — the intent of a
     /// delete is "drop what was said in that span," so re-emitted content
--- a/Mila/Audio/LiveTranscriber.swift
+++ b/Mila/Audio/LiveTranscriber.swift
@@ -112,6 +112,13 @@ final class LiveTranscriber: ObservableObject {
     /// amount so `windowAbsoluteStart` in `runOnce` keeps producing
     /// correct absolute timestamps for the segment merge.
     private var samplesDropped: Int = 0
+    /// Absolute time ranges of segments the user deleted from the live
+    /// pane. Incoming whisper output overlapping any of these is dropped so
+    /// a just-deleted line can't reappear on a later tick — the fixed-window
+    /// path re-transcribes a rolling buffer and would otherwise re-emit the
+    /// deleted content. The VAD path emits each utterance once, so this is
+    /// mostly a safety net there. Cleared on `start()`.
+    private var deletedRanges: [(start: Double, end: Double)] = []
     private var inFlight: Task<Void, Never>?
     private var tickTask: Task<Void, Never>?
     private var language: String = "he"
@@ -140,6 +147,7 @@ final class LiveTranscriber: ObservableObject {
         self.language = language
         self.buffer.removeAll(keepingCapacity: true)
         self.samplesDropped = 0
+        self.deletedRanges = []
         self.fullText = ""
         self.segments = []
         self.speakerNames = [:]
@@ -224,6 +232,32 @@ final class LiveTranscriber: ObservableObject {
     /// never arrives via `ingest`).
     func pauseBoundary() {
         detector?.flushForPause()
+    }
+
+    /// Delete a line from the live transcript. Records the segment's time
+    /// range so a later whisper tick (fixed-window mode re-transcribes a
+    /// rolling buffer) can't re-emit it, removes it from `segments`, and
+    /// recomputes `fullText`. Because `segments` is `@Published`, the app's
+    /// feed loop picks the change up automatically (MCP sidecar + LLM
+    /// re-feed), and the Stop-time snapshot excludes the deleted line.
+    func removeSegment(id: UUID) {
+        guard let idx = segments.firstIndex(where: { $0.id == id }) else { return }
+        let seg = segments[idx]
+        deletedRanges.append((start: seg.startSeconds, end: seg.endSeconds))
+        segments.remove(at: idx)
+        fullText = segments.map(\.text).joined(separator: " ")
+        liveLog.log("LiveTranscriber.removeSegment start=\(seg.startSeconds, privacy: .public) end=\(seg.endSeconds, privacy: .public) remaining=\(self.segments.count, privacy: .public)")
+    }
+
+    /// Whether an incoming segment's absolute time range overlaps a range
+    /// the user deleted. Any positive overlap counts — the intent of a
+    /// delete is "drop what was said in that span," so re-emitted content
+    /// there should stay gone.
+    private func isSuppressed(start: Double, end: Double) -> Bool {
+        for r in deletedRanges where start < r.end && end > r.start {
+            return true
+        }
+        return false
     }
 
     func transcribeNow() async {
@@ -406,10 +440,14 @@ final class LiveTranscriber: ObservableObject {
             let text = s.text.trimmingCharacters(in: .whitespacesAndNewlines)
             guard !text.isEmpty else { continue }
             guard s.start < originalDuration else { continue }
+            let absStart = startSec + s.start
+            let absEnd = startSec + s.end
+            // Skip anything the user deleted from this time span.
+            guard !isSuppressed(start: absStart, end: absEnd) else { continue }
             segments.append(LiveSegment(
                 id: UUID(),
-                startSeconds: startSec + s.start,
-                endSeconds: startSec + s.end,
+                startSeconds: absStart,
+                endSeconds: absEnd,
                 text: text,
                 speaker: nil,
                 stable: true
@@ -464,10 +502,14 @@ final class LiveTranscriber: ObservableObject {
         let absoluteSegs: [LiveSegment] = whisperSegs.compactMap { s in
             let text = s.text.trimmingCharacters(in: .whitespacesAndNewlines)
             guard !text.isEmpty else { return nil }
+            let absStart = windowAbsoluteStart + s.start
+            let absEnd = windowAbsoluteStart + s.end
+            // Don't resurrect a line the user deleted from this span.
+            guard !isSuppressed(start: absStart, end: absEnd) else { return nil }
             return LiveSegment(
                 id: UUID(),
-                startSeconds: windowAbsoluteStart + s.start,
-                endSeconds: windowAbsoluteStart + s.end,
+                startSeconds: absStart,
+                endSeconds: absEnd,
                 text: text,
                 speaker: nil,
                 stable: true
--- a/Mila/Audio/LiveTranscriber.swift
+++ b/Mila/Audio/LiveTranscriber.swift
@@ -441,7 +441,10 @@ final class LiveTranscriber: ObservableObject {
             guard !text.isEmpty else { continue }
             guard s.start < originalDuration else { continue }
             let absStart = startSec + s.start
-            let absEnd = startSec + s.end
+            // Whisper can extend a last segment into the silence we padded
+            // onto short utterances. Clamp to the real audio so a delete
+            // cannot mark later utterances as suppressed.
+            let absEnd = min(startSec + s.end, startSec + originalDuration)
             // Skip anything the user deleted from this time span.
             guard !isSuppressed(start: absStart, end: absEnd) else { continue }
             segments.append(LiveSegment(
--- a/Mila/Audio/LiveTranscriber.swift
+++ b/Mila/Audio/LiveTranscriber.swift
@@ -117,7 +117,9 @@ final class LiveTranscriber: ObservableObject {
     /// a just-deleted line can't reappear on a later tick — the fixed-window
     /// path re-transcribes a rolling buffer and would otherwise re-emit the
     /// deleted content. The VAD path emits each utterance once, so this is
-    /// mostly a safety net there. Cleared on `start()`.
+    /// mostly a safety net there. Cleared on `start()` AND on `stop()` — it is
+    /// per-recording state, and `start()` is not the only way a new recording
+    /// begins (see `stop()`).
     private var deletedRanges: [(start: Double, end: Double)] = []
     private var inFlight: Task<Void, Never>?
     private var tickTask: Task<Void, Never>?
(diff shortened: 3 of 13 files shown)


urisland:
## Bugbot round 1 — one fixed, one flagged

Both findings were checked against the code rather than taken on trust. The first is real and is fixed in `ed2c1fb`; the second is real in substance but its stated mechanism is off, and I'm deliberately not implementing it in a relay. Detail below.

### 1. "Emptying the live transcript resurrects it" — High — **valid, fixed in `ed2c1fb`**

Correct, and it's the same resurrection I fixed for chunk mode in `69767d4` arriving through the one door that fix can't cover.

```swift
let hasLiveSegments = !finalLiveSegments.isEmpty
let liveTranscriptIsAuthoritative = hasLiveSegments && vadActive   // before
```

My UI gate keys on the *mode*, but emptying `segments` flips the authoritative test **by itself**, whatever the mode. So delete the last remaining line and: the row saves `.pending`, `finalizeTail` takes the `else` branch and calls `transcription.enqueue(updated)`, and the batch pass re-transcribes the WAV and writes every deleted line into `recordings.json`, the `.txt` sidecar and the `.srt`. Deleting everything — the strongest statement a user can make that something must not be recorded — was the one input that brought all of it back.

The root cause is that `hasLiveSegments` is doing two different jobs with one bit. As a fallback for *"the live path produced nothing, so the WAV still needs transcribing"* it is right and has to stay. It just cannot tell that apart from *"the user emptied this by hand"* — both are `[]`.

Fix: make them distinguishable. `deletedRanges` already records the fact, per recording, reset on `start()`; it's now readable as `LiveTranscriber.hasUserDeletedSegments`, and

```swift
let userEmptiedLiveTranscript = userDeletedLiveLines && !hasLiveSegments
let liveTranscriptIsAuthoritative = (hasLiveSegments || userEmptiedLiveTranscript) && vadActive
```

so a deliberate emptying stays authoritative and saves empty. The never-had-a-transcript fallback is untouched.

Two notes on the shape of the fix:

- **The delete is not disabled on the last line instead.** That would have been the smaller diff, but emptying the transcript is a legitimate thing to want, and refusing the *user's* action to protect an internal invariant is the wrong trade. The gate belongs on the write-back path, which is where the rule puts it.
- **The empty-authoritative path was traced, not assumed.** `shouldRediarize(liveSpeakerCount: 0)` is false so no pyannote pass runs; `writeSRT` gets an empty segment list; and both summarizer entry points bail on an empty transcript (`regenerate` on its explicit `transcript.isEmpty` guard, `summarizeIfNeeded` via `shouldSummarize`) — so nothing downstream tries to make something out of the empty row. The `finish` / `store.update` handoff ordering is unchanged.

New negative control in `MilaTests/LiveTranscriptLineDeleteTests.swift`: delete both lines, drive a real `stopRecording()`, then **`await service.waitForIdle()`** before sweeping every readable file under the store root — draining the queue first is the point, because on the old head the resurrection arrives from a background batch pass and a sweep that ran too early would have gone green. It fails on `415ff84` on both the `.completed` assertion and the sweep. The stub engine is still canned with both lines, so the pass it must not provoke would write both back.

### 2. "Live AI keeps deleted speech" — Medium — **real, but not fixed here; mechanism below**

The conclusion holds and the disk-reaching half is worse than the caveat I'd left in the body, so let me state it exactly. The **mechanism as described is wrong in one detail**, and the correction matters because it rules out a text-corruption bug:

> the remainder is shipped as additional text on the same `--resume` session

The delta is guarded (`Mila/Actions/LiveAISession.swift`):

```swift
if snapshot.hasPrefix(lastTranscriptSent) {
    delta = String(snapshot.dropFirst(lastTranscriptSent.count))…
} else {
    delta = snapshot            // ← a delete lands here
}
```

A mid-list delete breaks the prefix, so the fallback ships the **whole post-delete snapshot**, not a misaligned remainder. No garbled text, and the deleted line is not re-sent.

What *is* real is the history: the tick is `claude --resume <uuid>`, and the earlier turn that carried the deleted line is still in that conversation. The model can keep quoting it — and that reaches disk, which is the part my flag understated:

```swift
updated.summary = finalSummary.isEmpty ? nil : finalSummary
updated.actionItems = finalItems.isEmpty ? nil : finalItems
```

in `stopRecording`. So the transcript in `recordings.json` is clean (tests 1–3 pin that) while `summary` / `actionItems` on the same row can still quote the deleted words. That is a `deleted-data-stays-deleted` violation in the derived fields, not merely "the LLM session remembers". Claude-session mode only: stateless mode re-sends the full clean snapshot with a `CURRENT STATE` block the prompt explicitly lets the model drop items from, so it self-corrects on the next tick.

**Why I'm not fixing it in this PR.** A complete fix is two changes, and the second is a product decision:

1. Rotate the session on a deletion — drop `sessionID`, mint a fresh UUID, clear `lastTranscriptSent` — so the history no longer holds the line. The machinery exists: `feed` already discards a stale session when `context != lastContextSent`.
2. Clear the already-emitted `summary` / `actionItems`, because a rotation doesn't retract what the model already said. That blanks the user's live notes mid-meeting, and a delete followed immediately by Stop would save a recording with **no** summary at all — trading one data-loss surface for another.

That is a change to a component the contributor never touched, on behaviour nobody has agreed to, in a relay, on a branch I cannot compile locally. Smuggling it in seemed clearly worse than naming it. Happy to do it as a follow-up if you want (1) alone, or (1) + (2); worth its own issue either way.

Also unaffected by both of these: `Recording.encode(to:)` is still untouched, so `StoredRecording` needs no mirror change.

CodeRabbit posted only a rate-limit notice on this PR (free-tier quota exhausted), so there is nothing to address from it.


---

urisland:
## Bugbot round 2 — both fixed in `ec81164`, plus one design question for you

Both findings are consequences of my own round-1 fix, and both are real. Fixed. But the second one is the round-1 Medium coming back in a new shape, which is the signal in this repo's rules to stop patching and say something, so there's a question at the end I'd like your call on before anything else lands.

`unit-and-ui-tests` passed at `ed2c1fb`, so the round-1 fix and its test compile and pass; this round builds on that.

### 1. "Stale deletions skip the next batch" — Medium — **valid, and my regression. Fixed.**

Confirmed, and the lifecycle claim is exactly right. `deletedRanges` was cleared only by `start()`, and `start()` is not the only way a recording begins — `wireLiveAIPipeline`'s hardware-gated branch calls `_ = transcriber.stop()` and deliberately never calls `start()`:

```swift
guard aiSettings.isLiveAIAvailable else {
    // BUT: clear the transcriber's `segments` / `useVAD` first. Without this, a stale
    // live transcript from a previous recording … would still be sitting in memory;
    // `stopRecording` would snapshot it onto the new recording and could mark it
    // `.completed` without ever running batch transcription. Cursor flagged on 62e1c3b.
    _ = transcriber.stop()
```

That comment is describing **this same leak, for the sibling field**, and it's the reason the fix belongs in `stop()`: that call site's whole purpose is to reset per-recording live state.

The failure mode is worse than the one my flag prevents. After an emptied recording, the next gated capture has an empty transcript *for real*, the stale flag makes it look authoritative, no batch pass is enqueued, and it saves with **no transcript at all** — silent, total, and on a recording the user did nothing to.

`stop()` now clears the ranges. Every call site is a teardown (end of recording, the failed-stop early return, the gated re-arm, dictation), so none of them wants the ranges kept, and suppression is unaffected: `stopRecording` snapshots `hasUserDeletedSegments` before its own `stop()` — the same ordering it already relies on for `segments`.

New test: empty the transcript, stop, then start the next recording the gated way (`transcriber.stop()`, no `start()`), and assert the new transcript actually lands. Fails on `ed2c1fb`.

> While verifying this: **that comment's claim about `segments` is not true today.** `LiveTranscriber.stop()` cancels the tasks and the detector and returns `fullText` — it does not clear `segments` or `fullText`. So the stale-live-transcript leak described at `MilaApp.swift:1457` looks like it is still open on `main`, independent of this PR: a previous recording's *non-empty* transcript can be snapshotted onto a gated capture. I did not fix it here — it's pre-existing, out of this PR's scope, and clearing `segments` in `stop()` has UI consequences (the live pane reads it) that want their own change. Flagging it because it's adjacent and I only found it by pulling on this thread.

### 2. "Deleted text survives in the summary" — High — **valid. Fixed for the emptied case in `ec81164`; the partial case is the question below.**

Verified end to end. `stopRecording` snapshots the Live AI output onto the row unconditionally:

```swift
updated.summary = finalSummary.isEmpty ? nil : finalSummary       // liveAISession?.summary
updated.actionItems = finalItems.isEmpty ? nil : finalItems
```

and on a fully emptied transcript nothing ever replaces it: `RecordingSummarizer.regenerate` returns early on `transcript.isEmpty`, and an authoritative row owes no batch pass so there's no completion hook either. Net effect — the transcript reads clean while the deleted words sit in the `summary` field of `recordings.json` and in MCP `get_transcript`, which serves summary + action items by default. My round-1 fix is what made that snapshot final, so I own it.

**Your framing is the one I've implemented.** Emptying the transcript is a *discard of the transcript*, so the artifacts derived from it go with it: `summary` and `actionItems` are cleared when the transcript was emptied. It's contained — four lines in `stopRecording`, at the point that already computes the fact. It touches no LiveAISession lifecycle, no `regenerate`, and nothing mid-meeting (this is Stop). New test seeds real Live AI output through a `seedForTesting` seam (both fields are `private(set)`; same idiom and justification as `LiveTranscriber.seedForTesting`) and asserts the marker is absent from the row, from `recordings.json`, from `get_transcript`, and from every file under the store root. Fails on `ed2c1fb`.

### The question — the partial-delete case, which I am deliberately not patching

While verifying the above I found the hole is **wider than the emptied case**, and one half of it is not new with this PR:

| | `summary` | `actionItems` |
|---|---|---|
| Transcript emptied | stale, nothing regenerates → **fixed here** | stale, nothing regenerates → **fixed here** |
| One line deleted, others kept | regenerated from the clean `fullText` by `finalizeTail` → clean | **never regenerated by anything.** `RecordingSummarizer` only ever writes `summary`. |

So after a partial delete, an action item that quotes the deleted line persists onto the row. That is a `deleted-data-stays-deleted` violation, it reaches `recordings.json` and `get_transcript`, and it is **pre-existing in shape** — `actionItems` have never had a regeneration path — but this feature is the first thing that makes the transcript and the items disagree.

I am not fixing it, and I don't think I should be the one to decide it, because every option is a product call:

1. **Drop all action items on any delete.** Contained, but destroys items the user wants over one removed line. Trades a privacy leak for data loss.
2. **Re-derive the items at Stop from the clean transcript.** Correct, but needs a Live-AI round trip on the Stop path — latency on every delete-then-stop, a failure mode when the CLI is unreachable, and it pulls in the session lifecycle (the `--resume` history still holds the line, so the rotation from round 1's finding is a prerequisite).
3. **Give `actionItems` a regeneration path** the way `summary` has one, and let it run whenever the transcript changed. Probably the right long-term answer, and clearly larger than this PR.

This is also the third round on the same underlying issue — round 1's "Live AI keeps deleted speech", then this — so per `.claude/rules/pull-requests.md` I'm stopping the patch cycle here rather than guessing a fourth time. **Which of 1/2/3 do you want, and as a follow-up issue or in this PR?** I'll implement whichever you name; my own preference is 3 as a follow-up, with this PR shipping the emptied-case discard that's already in.

Nothing here changes `Recording.encode(to:)`, so `StoredRecording` still needs no mirror update. The `finish` / `store.update` handoff ordering is untouched in both commits.


---

PR #230, opened 2026-08-31
feat(recording): choose a folder and meeting name for a recording (relay of #127)

### LINKED ISSUE #223: No way to choose a destination folder or name a meeting at record time
## What's missing

Two things you can only do *after* a recording exists, both of which you usually know *before* you press Record:

1. **Choose which folder the recording is filed into.** Folders already exist (`Recording.folder`, `RecordingStore.assign` / `createFolder`, the sidebar's context-menu assignment), but the only way to use one is to record first and then move the row. Every recording lands unfiled and has to be filed by hand.
2. **Name the meeting.** Every recording is saved under the auto-generated date-stamped default (`Recording · Aug 31, 2026 at 2:31 PM`). You can rename it in the post-stop sheet, but you already knew it was the Q3 board call when you pressed Record — and if LLM name suggestion is on, it proposes its own title over the top of that default.

## Why it matters

The filing step is per-recording manual work that scales with usage, and it happens at the worst moment: right after a meeting, when the user wants to walk away. Anyone who keeps folders (a client per folder, a project per folder) pays that cost on every recording, and recordings that aren't filed promptly tend never to get filed at all.

Naming has a sharper edge: the post-stop rename sheet is the only place a title can be set, so a user who dismisses it ends up with a library of date stamps.

There is also an existing defect in this area: **a folder chosen for a recording is lost when the recording is renamed after stop.** The rename sheet snapshots the recording at `init`, and SwiftUI reuses that `@State` across presentations, so the sheet can write a stale folder back over the current one — and `applyFolder()` did not run on every exit path.

## Intended shape

Set both *before* the recording starts, and let them stay editable *up to* Stop — they are applied when the recording is saved, so a change at any point still takes effect:

- **Home** — a "Save to" folder picker and an optional meeting-name field.
- **Live recording header** — the same pair, editable mid-recording.
- **Recording detail view** — an inline folder menu in the meta line, mirroring the sidebar's context-menu assignment but reachable from the open recording.
- **Post-stop rename sheet** — a folder picker with an inline "New Folder…" field, and the stale-folder bug above fixed.

Two properties worth deciding deliberately rather than by accident:

- **The pick should be per-session, not persisted.** A folder chosen for one meeting silently becoming the default for the next is the worse failure mode — you'd only notice weeks of recordings landed in the wrong place after the fact. Same for the name: clear it after each save, including on a *failed* stop.
- **A user-typed meeting name must not be handed to the auto-title job.** That job protects a user's title by only writing the LLM suggestion when the recording still carries the default title it captured as its baseline. If the user typed the name before recording, that name *is* the title at save time, so it becomes the baseline, the guard sees no divergence, and the suggestion replaces the name the user chose.

Note for whoever implements it: the folder is a **metadata label** in `recordings.json`, not a directory on disk. Keeping it that way is what stops this touching `store-location.json` and `MilaStoreReader`'s single-root assumption.

## Credit

Reported and implemented by **@ValeroK** in #127. This issue exists so the work has the durable linked record CI requires (`.claude/rules/pull-requests.md`); the implementation is being relayed onto an in-repo branch because fork PRs can't be reviewed under a contributor's own quota and fork CI needs per-run approval on every run.

urisland:
Closes #223

**All of the feature work here is @ValeroK's** (#127, part of the #99 split).
Five of the seven commits are his, cherry-picked with authorship intact — nothing
was rewritten. The relay is mechanical: fork PRs can't be reviewed under a
contributor's own quota and fork CI needs per-run approval on every run, so the
established pattern here is to recreate the commits on an in-repo branch (same
as #198 → #204 and #129 → #164). #127 is closed with a pointer to this PR.

The old PR applied to today's `main` **without a single conflict** — every commit
cherry-picked clean and the resulting diff is byte-identical to the fork diff.
Nothing had to be reimplemented.

## ⚠️ Scope note: the fork moved after I branched, so this PR carries one unrelated fix

I cut this branch from fork head `b4bce8b`. Before closing #127 I re-fetched its
head and found @ValeroK had pushed one more commit, `74fe137`
**"fix(recordings): stop discarded and empty captures from coming back"** (7 files,
+183/−5). Closing #127 without it would have discarded his work, so it is
cherry-picked here too — authorship intact, those seven files byte-identical to
his.

It is **not** part of the folder/meeting-name feature, so this PR is now two
things. **Say the word and I'll split it into its own PR + issue.** I left it here
because the alternative was dropping it on the floor. What it fixes, both real:

- **Empty captures came back as phantom rows.** `AVAudioFile` reserves a 4 KB
  header block on open, so a capture that never received a frame is 4096 bytes —
  not the 44 the old `size < 512` guard in `recoverOrphanRecordings` was written
  against. The guard therefore never fired: every empty capture was adopted by the
  launch-time orphan sweep, enqueued, transcribed to nothing, and left as a failed
  "Recovered recording". Now compared against `RecordingStore.emptyWAVByteCount`.
  The threshold stays a **byte count** on purpose — a genuinely crashed recording
  has an unfinalized header claiming zero frames, so asking `AVAudioFile` for a
  duration would discard exactly what the sweep exists to save. Both sides pinned
  by new tests (`test_empty_orphan_wav_is_not_recovered`,
  `test_orphan_wav_with_audio_is_still_recovered`).
- **Discard left the Obsidian note in the vault.** Export runs off the summarizer's
  completion hook, which can land either side of a discard, so clearing the pending
  gate alone left a note behind whenever the summary got there first. New
  `ObsidianExporter.discardNote(for:)` clears the gate *and* removes an
  already-written note; `PostRecordingCoordinator.cancelAndDiscard` calls it,
  against the **stored** row (its title is what named the file on disk; the
  `pending` snapshot's may predate a rename made in the sheet). The removal reuses
  the write path's rule — never resolve a stored path back out of the vault
  (`ObsidianPathSanitizer.isContained`) — so a relocated vault can't turn this into
  a delete of an unrelated file. Both orderings tested, plus the `MilaApp` wiring.

Risk-surface check on this extra commit specifically: it adds **no** `Recording`
field (so no `StoredRecording` mirror change, no `StoredRecordingDriftTests`
touch), and it moves nothing in `recordingsDirectory` — the one file it deletes is
a note in the user's Obsidian vault, which is not the transcript store, so
`store-location.json` and `MCPAccessGate` are untouched here as well.

## What & why (the feature)

Destination controls in the three places you'd naturally name or file a
recording, plus a fix for the folder being lost on rename after stop:

- **Home** — a "Save to" folder picker and an optional meeting-name field, so
  both are set *before* you press Record.
- **Live recording header** — the same pair, editable mid-recording. Both are
  applied when the recording is saved, so a change at any point up to Stop takes
  effect.
- **Recording detail view** — an inline folder menu in the meta line, mirroring
  the sidebar's context-menu assignment but reachable from the open recording.
- **Post-stop rename sheet** — a folder picker with an inline "New Folder…"
  field.

`QuickActionsController` gains `nextRecordingFolder` / `nextRecordingTitle`, both
per-session and deliberately **not** persisted, and both cleared on the failed-stop
path as well as the successful one. `resolvedRecordingTitle` is pure and `static`
so the fallback-to-date-stamped-default logic is testable without a session.

Two things worth reading the diff for, both his:

**The NSMenu isolation.** A SwiftUI `Menu` is backed by an NSMenu that runs a
nested modal event-tracking runloop while open. Every one of these folder menus
sits in a body that re-renders at transcript or progress cadence, so adding the
feature naively *introduces* a hang. The controls are pulled into
`Mila/Views/RecordingFolderControls.swift` as leaves observing only
`RecordingStore` and `QuickActionsController` — never `LiveTranscriber` /
`TranscriptionService` / `LiveAISession`. Same rationale as the existing
`RecordingElapsedLabel`. For the same reason the rename sheet no longer holds
`@EnvironmentObject transcription`; the status row is extracted into a private
`RenameTranscriptionStatusRow` so progress ticks re-render that leaf instead of
the whole sheet including its `Menu`.

**The auto-title interaction.** `armAutoSuggestTitle` protects a user's title by
only writing the LLM suggestion when the recording still carries the baseline
title it captured at add time. When the user types a meeting name before
recording, that name *is* the title at save time — so it becomes the baseline,
`applyAutoSuggestedTitle`'s `recording.title == baseline` guard passes, and the
CLI suggestion overwrites the name they chose. `present(_:titleWasUserProvided:)`
skips the job in that case; the Suggest button still offers it on demand. I read
this one against the code and his diagnosis and fix are both exactly right.

## What I changed on top of his work

Two commits.

`test(recording): negative controls …`:

- **Three negative-control tests** (below) — the PR had none for the destination
  edges.
- **Rewrote the comment on the `store.assign` call** in `stopRecording`. It
  claimed the call covers "a brand-new folder or a case-insensitive match", which
  is not reachable from the picker (the picker only offers names already in
  `store.folders`). The one way the *UI* reaches it is a folder deleted or renamed
  mid-recording. The comment now says that, names the invariant the call keeps,
  and records that honouring a stale pick rather than silently unfiling is
  deliberate. No behaviour change.

`fix(recordings): unfuse the orphan-sweep doc comment …` — a one-hunk cleanup on
his late commit. `emptyWAVByteCount` landed *between* `recoverOrphanRecordings`'
doc comment and the function, which silently reassigned both: the crash-recovery
prose became the constant's documentation and the constant's prose became the
function's. Moved the constant above so each `///` block documents the declaration
it describes. Comments only, no behaviour change.

## Findings on the risk surfaces

**1. Recording storage paths and the MCP contracts — verified clean, and the
reason is load-bearing.** `Recording.folder` is a **metadata label inside
`recordings.json`**, not a directory. `RecordingStore.assign` mutates the string
and the `folders` list and calls `persist()`; it moves no files. The audio, the
`.txt`, `.summary.txt` and `.srt` sidecars all stay in `recordingsDirectory`,
whose name each recording derives from `audioFileName` — set at record *start*.
So there is **no interaction at all** with `store-location.json`: it still points
at one directory, `MilaStoreReader.resolvesActiveStore` still resolves it, and
`MCPAccessGate`'s `enabled && storeLocationIsDiscoverable` mirroring is untouched.
`MilaStoreReader` in fact already reads and filters `folder`
(`Filter.folder`, `list_recordings`' `folder` arg), so a recording filed at record
time is fully visible to `mila-mcp`. There's now a test asserting exactly that
through the real reader, so if a destination folder ever became a real directory
this fails loudly rather than silently leaving the helper resolving
`recordingsDirectory/` and finding nothing.

**2. Persisted schema drift — nothing to mirror.** This PR adds **no** field to
`Recording`. Both fields it writes at save time (`title`, `folder`) already exist,
and `folder` is already mirrored in MilaKit's `StoredRecording` with lenient
`decodeIfPresent` decoding. `StoredRecordingDriftTests` is unaffected and its
allowlists don't need touching. The one-shot state (`nextRecordingFolder` /
`nextRecordingTitle`) is in-memory only and deliberately never persisted — nothing
new lands in `recordings.json` or in UserDefaults. The late commit adds no field
either; `RecordingStore.emptyWAVByteCount` is a `static let` constant.

**3. Logging — this diff adds one log site, and it does raise the stakes on
existing ones.** The only new one is the cherry-picked `discardNote` failure path,
which prints a vault *path* on a failed `removeItem` — an existing pattern in that
file. No user-chosen path or meeting name reaches any log field in the feature
commits. What the feature *does* change is the nature of `recording.title` at
creation time: previously always a machine `"Recording · "`, now potentially a
client or employer name. `TranscriptionService` `print`s `recording.title` on
roughly a dozen lines, and `MilaApp` prints the title-derived `audioFileName` —
pre-existing, but materially more sensitive after this. **#218 (`fix/log-privacy`)
fixes exactly those files**, so I deliberately did not duplicate the work here.
Overlap with #218 in this diff is limited to `PostRecordingCoordinator.swift`,
where #218's hunk is the send path around line 400 and mine are `present` (~131)
and `cancelAndDiscard` (~500) — no textual conflict either way round.

**4. Feature gates — one gap, deliberately left as a design question rather than
"fixed".** There is no re-validation of `nextRecordingFolder` between selection
and Stop, so a user who deletes or renames that folder in the sidebar
mid-recording reaches `store.assign` with a name `store.folders` no longer holds,
and `assign` auto-creates it — the deleted folder reappears. I looked hard at
turning that into a readiness check (`.claude/rules/feature-gates.md`) and
concluded his behaviour is the better of the two: "Save to Work" was on screen
right up to Stop, so filing it in Work is what the user was told would happen,
and silently unfiling would make the UI a liar. Honouring the pick is also what
keeps the invariant the sidebar depends on — a recording's `folder` is always a
name present in `store.folders`. **Say the word if you'd rather it dropped the
pick and fell back to All Transcriptions and I'll wire an observer on the
controller** (it can't live in the view: both pickers can be off-screen when the
folder is deleted). Either way the behaviour is now pinned by a test and
explained in the comment.

**5. Sandbox / entitlements — not applicable.** There is no filesystem picker
here: no `NSOpenPanel`, no user-chosen directory, no security-scoped bookmark, no
entitlement change. The "folder" is a Mila folder, so nothing needs to survive
relaunch as a bookmark and there is no permission to forget. (The one place the
app *does* take a real directory — `RecordingStorageSettings` / `relocateRecordings`
— is untouched.)

One observation I did **not** act on, for the maintainer: both new fields are
`@Published` on `QuickActionsController`, so every keystroke in the meeting-name
field publishes an `objectWillChange` that anything observing `actions` sees —
including `HomeView` and `LiveAIRecordingView`. The leaf isolation protects the
*menus* from the transcript-cadence churn (which is the hang), but the reverse
edge means typing a name during a long recording re-renders those parents. I
can't measure it without a Mac, and moving the state out would be a redesign of
his API, so I'd rather flag it than guess.

## How it was tested

I could not compile or run anything: this relay ran on Linux with no Xcode, no
`xcodebuild` and no Swift toolchain (`download.swift.org` is egress-blocked). So
verification was (a) reading every changed file end to end against `main`, grepping to
confirm every symbol referenced actually exists, and (b) CI on this PR. Called out
plainly because it means CI is doing the compiling here, not me. Specifically
checked by hand:

- `project.yml` globs `- path: Mila`, so the new `RecordingFolderControls.swift` is
  picked up with no project change (and no `.xcodeproj` hand-edit).
- Every environment object the new leaves need (`store`, `actions`, `transcription`)
  is injected on the main window in `MilaApp`, and the rename sheet is presented
  via `.sheet(item:)` from `ContentView`, so it inherits them.
- No `transcription.` reference survives in `RenameRecordingSheet` outside the new
  `RenameTranscriptionStatusRow`.
- `PostRecordingCoordinator.dismiss(savingTitle:)` re-reads the row from the store
  rather than writing back the `pending` snapshot, so `applyFolder()` → `dismiss()`
  ordering in the sheet can't have the folder clobbered by the title save. Same for
  the inline drain in `stopRecording`, which re-reads via
  `store.recordings.first(where:)` before `store.update`.
- `runLLM` closure arity (14), `startFakeRecordingForTesting`, `awaitFinalizeTails`,
  `TestSupport.writeStereo48kSineWav`, `store.recordings(inFolder:)`,
  `store.storeURL`, `MilaStoreReader.recording(id:)` / `listRecordings(filter:)` all
  exist as used, and MilaKit is already a `MilaTests` dependency in `project.yml`.
- For the cherry-picked commit: `clearPending`, `writtenIndex()`, `setWrittenIndex`,
  `indexKey(vault:id:)`, `migrateLegacyIndexEntry`, `kickGitSync`, `singleLine`,
  `markPending` / `isPending`, `ObsidianExporter.fileName(for:)` / `export(_:)`,
  `ObsidianPathSanitizer.isContained(_:in:fileManager:)`,
  `ObsidianVaultSettings.setVault` / `subfolder` / `gitSyncEnabled`,
  `RecordingSummarizer.awaitInFlight`, `store.consumePendingRecoveryIDs()`,
  `store.recordingsDirectory`, and the `tempRoot` / `addRecording` /
  `addCompletedRecording` fixtures in the three touched test classes all exist as
  used. I also searched `MilaTests/` for a fixture pinning the old 512-byte
  threshold — there is none, so no existing test had to change.

### Tests

His: `test_nextRecordingFolder_defaults_to_all_transcriptions` (non-persistence),
`test_stopRecording_clears_nextRecordingFolder`,
`test_failed_stop_clears_nextRecording_title_and_folder`,
`test_stopRecording_registers_a_brand_new_folder`,
`test_stopRecording_does_not_auto_title_a_user_named_recording`, both branches of
`resolvedRecordingTitle`, and
`test_present_skips_auto_title_when_the_user_named_the_recording`. These are
discriminating — `test_stopRecording_clears_nextRecordingFolder` asserts a recording
lands in a *chosen* folder under a *chosen* name, which is the behaviour that does
not exist on `main` at all. From the late commit:
`test_empty_orphan_wav_is_not_recovered`,
`test_orphan_wav_with_audio_is_still_recovered`,
`test_discardNote_removes_an_already_exported_note`,
`test_discard_before_the_summary_lands_blocks_the_export`,
`test_discardNote_is_a_noop_when_nothing_was_exported`, and
`test_cancelAndDiscard_removes_the_obsidian_note` (which covers the `MilaApp`
wiring — without it the coordinator deletes the recording and leaves the note).

I searched `MilaTests/` for fixtures pinning the old behaviour, since one surviving
into `main` has twice made a defect look intentional here. The `XCTAssertNil(…folder)`
assertions in `RecordingStoreTests` and `TranscriptionServiceTests` are about
`assign` semantics on fixture-created rows, not about recordings always landing
unfiled, so none of them had to change. `StoredRecordingDriftTests` needs no change
for the reason in finding 2.

Mine, all three negative controls:

- `test_meeting_name_with_path_separators_cannot_escape_the_recordings_directory`
  — drives a meeting name containing both `..` and `/` and asserts every sidecar
  the store resolves still lands *directly* inside `recordingsDirectory`, while the
  title survives verbatim. On this path the WAV is named at record start so the name
  never reaches a path — but nothing said so, and `freshAudioURL(suggestedName:)`
  appends its argument as a path component with no sanitizing of its own
  (`FileTranscriber` has to strip `/` and `:` itself for exactly this reason). "Name
  the file after the meeting" is an obvious-looking follow-up; this is the tripwire
  for it.
- `test_folder_deleted_mid_recording_still_saves_and_stays_reachable` — deletes the
  chosen folder while the recording is running and asserts the recording is still
  saved and is never left tagged with a folder absent from `store.folders`. This is
  also the test that justifies the `store.assign` call: `Recording(folder:)` already
  persists the name, which makes the `assign` look redundant and tempting to delete.
  Delete it and this fails.
- `test_a_chosen_folder_and_meeting_name_stay_resolvable_to_an_external_reader` —
  reads back through `MilaStoreReader`, the same reader `mila-mcp` uses, and asserts
  both `recording(id:)` and `listRecordings(filter: .init(folder:))` see it.

The lookups in all three match the audio file's *stem*, not its full name: the tail
enqueues for batch transcription and completion can hand the row to
`compressRecordingAudio`, which rewrites `.wav` to `.m4a`. Matching the
whole name would race that transcode.

## Automated review

Bugbot's one finding on `b5dbaca` ("Rename folder menu can hang") is **skipped with
a stated reason** — see the comment below. Short version: its true sub-claim (the
`Menu` is built in the sheet body reading `store.folders`) isn't sufficient, and the
inference carrying it — that this PR treats a `store` publish as the hazard — is
contradicted by the pattern it cites, since all three extracted leaves in
`RecordingFolderControls.swift` also observe `store` by design. The
transcript-cadence publisher (`TranscriptionService.progress`, `@Published` on the
service, not the store) is what the diff removed from that sheet. CodeRabbit posted
only a rate-limit notice (free-tier quota), so nothing is outstanding from it.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — **not possible in
      this environment** (Linux, no Xcode/Swift toolchain); CI is the check. The
      contributor reported `make build` and the unit suite green on his machine, and
      that he could not run the XCUITest target locally ("Timed out while enabling
      automation mode", reproducing on a clean `main`, so environmental).
- [x] User-facing change is noted in the README changelog (if applicable) — the
      changelog there is per *release*, not per PR (cf. #204, which added none), so
      this lands with the next version bump.

Worth a human confirming by hand, since no automated test can: open the folder menu
on the live recording screen **while a transcription is running**. That's the
beachball the isolation exists to prevent, and it is the one claim in this PR I
could neither compile nor reproduce.

---

---

> [!NOTE]
> **Medium Risk**
> Touches post-stop discard, Obsidian vault deletes, and recording metadata at save time; race handling is well documented but vault removal and late export hooks deserve a quick manual pass.
> 
> **Overview**
> Adds **per-session** “Save to” folder and optional **meeting name** on Home and the live recording UI, applied at stop via `nextRecordingFolder` / `nextRecordingTitle` (cleared after save and on failed stop). Recordings get the resolved title and folder; a mid-recording sidebar folder delete is reconciled with `store.assign` so rows stay reachable. **Detail view** and the **rename sheet** gain folder assignment; user-typed names skip the background auto-title job (`titleWasUserProvided`).
> 
> New **`RecordingFolderControls`** and **`RenameTranscriptionStatusRow`** isolate folder menus and transcription progress from high-frequency publishers so open menus do not beachball during transcription.
> 
> **Discard / Obsidian:** `ObsidianExporter` keeps a session **discarded** tombstone, gates `markPending`/`export`, and adds **`discardNote(for:)`** to remove vault notes (with safe path checks and git sync). **`cancelAndDiscard`** wires the exporter and uses the stored row for the note title.
> 
> **Recovery:** orphan WAV sweep skips files at or below **`emptyWAVByteCount` (4096)** so empty captures are not resurrected as failed “Recovered recording” rows.
> 
> Extensive unit tests cover folder/title wiring, Obsidian discard orderings, orphan thresholds, MCP reader visibility, and path-separator negative controls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 181eeb78b9248028f30d338e260bcab4fc49cb15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Choose a meeting name and destination folder before or during recording.
  - Organize recordings into folders from recording details and rename screens.
  - User-provided titles are preserved without automatic replacement.
  - Discarding a recording now removes its associated exported note.

- **Bug Fixes**
  - Prevented discarded recordings from reappearing after delayed processing.
  - Empty audio captures are no longer restored as failed recordings.
  - Improved cleanup and retry behavior when removing exported notes.


urisland:
## Bugbot finding on `b5dbaca`: "Rename folder menu can hang" — not applying it, and the reason is worth recording

I took this one seriously because of its strongest claim: that this PR *already
recognises* the beachball and isolates the other pickers from it while leaving the
rename sheet exposed. That would be an internal inconsistency in the diff, not a
nit. I checked all three links in the chain. The first holds; the other two don't,
and the second is the one that decides it.

**1. "The folder `Menu` is built inside `RenameRecordingSheet`'s body, which reads
`store.folders` there" — TRUE.** `Mila/Views/RenameRecordingSheet.swift:371,373`
read `store.folders` inside the `Menu` content, in the sheet's own `body`, and the
sheet holds `@EnvironmentObject private var store` (line 14). No argument.

**2. "…the failure this PR isolates in the other pickers" — FALSE, and this is the
crux.** The other pickers are *not* isolated from `store`. All three leaves in
`Mila/Views/RecordingFolderControls.swift` keep `@EnvironmentObject private var
store` and read `store.folders` inside their `Menu` content —
`NextRecordingFolderPicker` (lines 25, 34-35) and `RecordingFolderMenu` (lines 72,
84-85). `store` is deliberately *inside* the boundary, because the folder list has
to be live. The file's own header comment says which side each object is on:

> These deliberately observe ONLY `RecordingStore` (for the folder list) and
> `QuickActionsController` (for the bindings) — never the high-frequency
> `LiveTranscriber` / `TranscriptionService` / `LiveAISession`.

So if a single `store` publish mid-tracking were enough to reconcile the NSMenu
host and beachball, the PR's own "isolated" leaves would hang identically, on Home
and in the detail view. The remedy the finding points at does not exclude the
hazard it describes. That mismatch is what makes it unactionable as written —
extracting the rename sheet's menu into `RecordingFolderControls.swift` would be
consistent-looking churn that changes this exposure by exactly zero, since the
extracted leaf would still observe `store` and its `Menu` body would still be
re-invoked by the same publish.

**3. "the inline drain and later transcription writes still call `store.update`
while the sheet is up" — true as stated, but they are discrete events, not the
flood.** The hazard is cadence, and no transcript-cadence writer reaches the store
on this path:

- Progress does **not** go through the store. `TranscriptionService.progress` is
  `@Published` on the *service* (`TranscriptionService.swift:28`), fed by a
  coalescing progress callback (~line 734). That is the high-frequency publisher,
  and it is exactly what extracting `RenameTranscriptionStatusRow` removed from
  this sheet — the sheet no longer holds `@EnvironmentObject transcription` at all.
- The batch pass writes the store **once, terminally**, via `mergePassResult`
  (`TranscriptionService.swift:963` — `store.update(live)`), on the completion
  branches around lines 672 and 822. There is no partial-segment streaming into
  the store on the post-record path.
- `RecordingSummarizer` publishes only `inFlightIDs` (start/finish,
  `RecordingSummarizer.swift:71`) and writes the store once at line 458.
- The inline drain's folder write lands *before* the sheet exists:
  `store.assign` is `QuickActionsController.swift:776`, `postRecording.present(…)`
  is line 788. It cannot reconcile a menu that isn't on screen yet.
- The remaining late write is the offline re-diarize merge
  (`QuickActionsController.swift:1129`) — one `store.update`, once, for a Live-AI
  recording.

That leaves the sheet's body driven by a handful of discrete, event-level publishes
— which is the same footing as the three leaves the PR does extract. The
transcript-cadence exposure the sheet genuinely had (`transcription`) was the one
the diff removed.

**Verdict: skipped, mechanism doesn't hold.** The one true sub-claim (the `Menu` is
in the sheet body) is not sufficient on its own, and the inference that carried it
— that the PR treats `store` as the hazard — is contradicted by the pattern it
cites. What I am *not* claiming is that a rename-sheet re-render is free: a
discrete store publish does rebuild more of the sheet than it would a leaf. That's
a cost question, not a hang, and I'd rather not add an extraction I can't measure
on this machine to a diff whose menu isolation is already the part most in need of
a human's eyes on real hardware. That manual check is still the one open item in
the PR description: **open the folder menu on the live recording screen while a
transcription is running.**

Happy to be overruled with a repro — if a single `store` publish really does
beachball a tracking NSMenu here, the fix is bigger than this PR and lands on all
three pickers plus `RecordingElapsedLabel`'s neighbours, not just the sheet.

CodeRabbit posted only a rate-limit notice on this PR (free-tier OSS quota
exhausted), so there is nothing outstanding from it.


---

urisland:
## Bugbot round 2 on `4e185d3`: 3 findings, all real, all fixed in `8e83d1e`

Worth stating plainly: **all three landed in the late commit I relayed** (`2eccee2`,
@ValeroK's `74fe137`), none in his original four. That's the argument for having
re-fetched #127's head instead of closing it blind — the alternative wasn't "a
tidier PR", it was this work silently not existing, bugs and all.

### 1. HIGH — "Discard can still export later" — real, and worse than a race

Confirmed, and the mechanism is more ordinary than "race". Chain:

- `PostRecordingCoordinator.cancelAndDiscard` has no reach into
  `QuickActionsController.finalizeTasks`, so **`finalizeTail` is not cancelled**.
  On the VAD path it `await`s the offline re-diarize before doing anything else,
  so it is routinely still running when the user hits Discard on the sheet —
  `stopRecording` presents that sheet at line 788 and launches `finalizeTail` at
  1046. The window is tens of seconds, not milliseconds.
- `finalizeTail` ends by calling `markPending(updated.id)`
  (`QuickActionsController.swift:1153` and `:1156`), re-arming the gate
  `discardNote` had just cleared. `clearPending` is not a tombstone — the later
  mark wins. Bugbot has this exactly right.
- The part that makes it *not* a narrow race: the summary that follows can be
  **skipped**, and every skip path in `RecordingSummarizer`
  (`summarizeIfNeeded` ~227, `regenerate` ~247 and ~254) fires
  `onSummaryFinished` **synchronously**, with
  `latestRecording(id, fallback: recording)` — which falls back to the caller's
  snapshot when the row is gone from the store. That snapshot isn't trashed, so
  `export`'s own trashed guard doesn't fire, and `MilaApp`'s hook
  (`MilaApp.swift:571`) gates on nothing but `isPending`. With **summaries off
  (or the LLM unconfigured) and Obsidian on, this is the ordinary path, not a
  race** — discard a recording, and its note appears in the vault a minute later.

What was already safe, and why it hid this: the *success* path bails on
`runSummary`'s "is the row still in the store?" lookup
(`RecordingSummarizer.swift:443`), which is what the existing
`test_recording_deleted_mid_flight_is_not_exported_on_the_success_path` pins. So
the suite looked like it covered this. It covered the half that was fine.

**Fix:** a session-scoped `discarded: Set<UUID>` tombstone on `ObsidianExporter`,
consulted by **both** `markPending` and `export`. Gating `markPending` alone
would not have been enough: `MilaApp.swift:614` has a second export entry point
that skips the pending gate entirely (`guard llm?.summaryEnabled ?? true else {
obsidian?.export(rec); return }`), so the choke point had to be `export` too.
Not persisted, deliberately — a discard only has to beat work already in flight,
and a fresh launch has none. `exportAll` is left alone: it's the "Sync existing
transcripts" backfill over `store.recordings`, which cannot contain a
permanently-deleted row. That reasoning is in the property's doc comment so the
next reader doesn't "tidy up" the asymmetry.

Pinned by `test_a_late_finalize_tail_cannot_re_export_a_discarded_recording`,
which drives the real ordering — discard, hard-delete, then the late
`markPending` plus a skipped summary — and fails without the tombstone.

### 2. MEDIUM — "Failed delete forgets vault note" — real

`discardNote` dropped the written-index entry before `removeItem` succeeded. That
entry is the **only** record of where the note lives, so a failed delete left the
discarded transcript in the vault permanently unreachable — no later discard could
find it. Same family as (1).

**Fix:** the entry is dropped only after the removal actually succeeds. A stale
entry pointing at a file that's already gone is still cleared (that's a genuine
tidy-up) but keeps returning `nil`, so it can't kick a git sync for a file that was
never removed — which my first cut of the reorder accidentally introduced and the
return-value contract caught. The index *migration* is now persisted up front so it
lands on every path rather than only the success one.

Pinned by `test_a_failed_removal_keeps_the_index_entry_for_a_retry`: a `FileManager`
subclass whose `removeItem` always fails, then a second exporter over the same
index that still finds and removes the note — only possible if the entry survived.

### 3. MEDIUM — "Delete failure logs vault path" — real

A bare `print` interpolating `note.path` and the error. Public by construction, and
it leaked twice: the vault is user-chosen and the filename is title-derived, and
Cocoa quotes the path again inside `localizedDescription` — mechanism 2 in
`bugbot-rules/no-user-content-in-logs.md`.

**Fix:** an `os.Logger` line on a new `obsidianLog`, following the idiom already in
`RecordingStore` (e.g. `:979`): `NSError.domain` and `code` stay `.public` so a
permissions failure is still distinguishable from a read-only volume, path and
`localizedDescription` are `.private`. Per the rule's "do not over-redact" section.
I touched **only** the new site — the four pre-existing `print`s in this file are
#218's, and I'm not reaching into its branch.

### Still skipped, unchanged

The earlier Medium on `b5dbaca` (rename sheet folder `Menu` rebuilding on a `store`
publish) remains skipped, for the reason in the comment above: all three extracted
leaves in `RecordingFolderControls.swift` also observe `store` by design, so the
remedy it cites doesn't exclude the hazard it describes. Nothing in these three
findings changes that analysis.


---

urisland:
Addressed CodeRabbit's review of `15f5c26` in `181eeb7`.

**Fixed — stale title on sheet reuse (`RenameRecordingSheet`).** This one is real, and it's the same bug this PR already fixes for the folder. The comment above `onChange(of: initialRecording.id)` states the premise outright — SwiftUI can reuse the sheet's `@State` for a later recording — which is why the folder pick is dropped there. `title` / `userEditedTitle` have identical exposure and weren't.

The path: user edits the title in one presentation → `userEditedTitle` becomes `true` → SwiftUI reuses the state → the id-change handler clears the folder but not the title → both title mirrors (`onAppear` at :223, `onChange(of: liveRecording.title)` at :239) are gated on `!userEditedTitle` and never reseed → `save()` writes the *previous* recording's name onto the new one. Worse than the folder version, since the title is what shows in the list.

**Fixed — test/transcode race.** Three of this PR's new tests matched on the full `audioFileName` after `awaitFinalizeTails()`. The existing `savedRecording(for:)` helper exists precisely because the finalize tail can rewrite `<stem>.wav` → `<stem>.m4a`, so those three were racing the transcode and could fail intermittently. Switched to the helper, which matches on the stem.

**Skipped, with reason — deleting skipped empty orphan WAVs.** Automatically deleting a user's audio file during a startup scan is a destructive action on a file the session didn't create, and the leak it avoids is a few KB per empty capture. That trade is a maintainer's call rather than something to take on a review-bot's suggestion inside a relay. Worth a follow-up issue if the accumulation turns out to matter.

**Skipped, with reason — production-generated WAV boundary fixture.** The existing test writes 4,097 bytes to pin the `> 4096` boundary, which is what the threshold actually keys on. Generating a real one-frame `AVAudioFile` would additionally prove a sample survives, but it tests `AVAudioFile`'s header size more than this code, and I can't run it here to confirm the generated size lands where expected.

**On the out-of-scope warning:** correct that the Obsidian discard lifecycle and the empty-orphan filter go beyond what #223 describes. Both come from @ValeroK's later commit `2eccee2` ("stop discarded and empty captures from coming back") plus the fixes layered on for Bugbot's findings — so it's the contributor's own work, not scope creep added during the relay. Splitting it would mean carving up their commit. Better resolved by widening #223 to cover the discard and orphan-recovery behaviour, if the maintainer agrees.

**Build status unchanged: UNVERIFIED locally** — no Swift toolchain in this environment, so CI remains the first compiler for these two edits.


---

urisland:
`unit-and-ui-tests` failed on `181eeb7`. Recording what failed and why I don't currently think it's this PR's, since I've now spent the one re-run confirming it.

**Failing tests:**
```
MicrophoneConfigurationChangeTests.test_a_storm_of_configuration_changes_is_bounded()   ×3
MicrophoneConfigurationChangeTests.test_change_after_the_grace_window_still_rebuilds()  ×1
TranscriptionServiceTests.test_retranscribe_after_audio_compressed_reads_current_file_not_stale_wav()  ×3
```
(The ×3s are the three in-job attempts from `-retry-tests-on-failure -test-iterations 3`, not three distinct tests.)

**Why this doesn't look like this PR:**

- `MicrophoneConfigurationChangeTests` is one of the suites named in #208 as intermittently failing under CI load. This PR touches nothing in the microphone-configuration path.
- `TranscriptionServiceTests.test_retranscribe_after_audio_compressed_reads_current_file_not_stale_wav` is **pre-existing** — it's present on `c3f7821`, before any of today's merges — and this PR touches neither `TranscriptionService` nor `AudioCompressor`. Its only file in that neighbourhood is `RecordingStore`, and that change is the startup orphan scan, which has no bearing on re-transcribing after a compress.
- Two unrelated suites failing in the same run, one of them a known-flaky one, reads as a bad/loaded runner rather than a regression. There are currently three CI pipelines plus a `main` push all competing for macOS runners.

**Caveat I'm not glossing over:** `TranscriptionServiceTests` is *not* on the #208 list, and it failed all three in-job retries. That's the same signature as `MicrophoneRecorderTests` earlier today — failed 3/3 retries, then passed cleanly on a fresh runner — which suggests some failures here are machine-state-dependent rather than randomly timing-sensitive, and survive in-job retry by construction. So the retry count alone doesn't settle it.

The real control is `main`'s own `unit-and-ui-tests` run on `e472466`, which is still queued. `main` already carries #218 — the one merged PR that touched `TranscriptionService` and `AudioCompressor`. If `main` fails the same test, this is `main`'s problem and not this PR's, and I'll fix it there. If `main` passes and the re-run here fails again, I'll treat it as this PR's and root-cause it properly rather than re-running again.

For the record, `build-and-test` is green on both this head and on `main`, so nothing here is a compile problem.


---

ValeroK:
Yes — please split the Discard / empty-capture recovery work out of this PR.

It was found while dogfooding the folder + meeting-name branch, but it is a separate bug (orphan sweep adopting 4 KB empty WAVs, plus discard racing Obsidian export). The folder/name feature should land without it.

Take the related follow-ups with it: `emptyWAVByteCount` doc-comment unfuse, and the discard tombstone so a late `finalizeTail` `markPending` cannot re-file a thrown-away recording. Those belong with the recovery/export fix, not with choosing a folder.

Thanks for offering.

PR #233, opened 2026-08-31
fix(llm): empty the shared LLM sandbox on the idle transitions

### LINKED ISSUE #190: LLM sandbox is now shared, so one invocation's leftover files are visible to the next

urisland:
Closes #190

## What & why

#187 pointed every LLM invocation at one stable working directory (`~/Library/Application Support/Mila/llm-sandbox/`) because the `claude` CLI derives its project identity from a slug of the cwd, so a per-run cwd minted a fresh `~/.claude/projects/` directory each time and made those conversations unresumable. The cost was that the directory is no longer empty at the start of a run: any CLI that writes a regular file into its cwd leaves it there for the next invocation to find.

Scope, honestly: this is not an external exposure. The directory sits under the user's own Application Support, readable only by them, and Mila writes no transcripts there. The narrow concern is that content **derived** from one recording's LLM call could be discovered by an unrelated later call, including one for a different recording. Separately — and this is the part a future reader would have been misled by — `LLMRunner`'s doc comment still promised "an empty directory", which stopped being true in #187.

### The option chosen: reference-count the invocations, sweep on both idle transitions

`acquireSandbox()` / `releaseSandbox(_:)` track how many children are currently chdir'd into the sandbox and empty it when the count is (or reaches) zero — **before the first child enters, and after the last one leaves**. `executeProcess` takes the lease with a `defer` that spans the child's whole life, not just its launch.

Against the two things #187 deliberately preserves:

1. **The stable project path.** Only the directory's *contents* are removed; the directory itself is never deleted or recreated. The path is unchanged, so claude still resolves one `~/.claude/projects/<slug>` for every run and #181 stays fixed. Nothing durable lives in the cwd to lose: conversations are stored under `~/.claude/projects/<slug>` (never touched, `--resume` still works), and workspace trust is granted per-invocation in argv (`cursor-agent -f`, `gemini --skip-trust`), not by state in the working directory.
2. **Concurrent invocations.** A sweep can only run when the count is zero, so it can never delete a file a live child is using. A Live AI tick and a one-shot Send still share the one cwd for as long as they overlap; the one-shot's entry and exit are both no-ops for the sweep while the tick is inside. This is exactly the hazard that rules out a naive sweep-before-each-run, and it is pinned by two tests.

**Why entry as well as exit.** The exit sweep cannot be trusted to have happened — a crash or force-quit skips it. Sweeping on entry when idle is what makes "a child that starts while nothing else is running starts in an empty directory" hold unconditionally, and it subsumes a sweep-at-app-launch hook (considered and dropped: it would clear crash residue slightly earlier while adding a second entry point enforcing the same invariant). It also keys the sweep to the directory the child is *about to enter*, which matters on the temp-fallback path where `sandboxDirectory()` can resolve somewhere else.

Hidden entries are swept too (`contentsOfDirectory` without `.skipsHiddenFiles`). A CLI's own dot-directory — `.claude/`, `.cursor/` — is precisely where a cached answer about the user's meeting would sit; sparing it would leave the fix covering only the naive case.

### On the per-invocation-subdirectory option

The issue says a subdirectory of a stable parent does not work, and that holds when checked against the code. `LLMRunner.sandboxDirectory()`'s doc comment and the #181 report both describe the project directory as `~/.claude/projects/<slug-of-cwd>` — the slug is of the **full** cwd path, not of a parent — and `LLMSandboxDirectoryTests.test_sandbox_path_has_no_per_run_component` already pins "the path must contain nothing run-specific: no UUID, no PID, no timestamp". A per-run child directory is a per-run cwd path, so it re-mints a project directory per run. Rejected, and now recorded in the doc comment so the next reader doesn't re-derive it.

### What the corrected doc comment promises

`sandboxDirectory()` no longer claims empty isolation as a property of the path. It states the guarantee the reference count actually keeps:

* A child that starts while no other invocation is running always starts in an empty directory — including the first run after a crash, because the entry sweep does not trust the exit sweep to have happened.
* An artifact an invocation writes here is visible only to invocations whose lifetime **overlaps** it. A later, non-overlapping call — a different recording's summary, a Settings → AI Provider test — never sees it.
* It is explicitly **not** a private directory per invocation. Overlapping runs share the cwd by design and can see each other's files while they overlap, because no sweep may run while any of them is inside.

Failures are visible rather than silent: a sweep that cannot unlink something logs at `error` ("leftovers stay visible to the next invocation"), with counts only — never filenames, which are derived from what the CLI was asked about.

## How it was tested

Cannot build on this machine (Linux, no Xcode/Swift toolchain), so the build and suite are unverified until CI runs. Tests added/changed:

**`MilaTests/LLMRunnerTests.swift`**

* `test_artifact_from_one_invocation_is_not_visible_to_the_next` — the test the issue asks for. Invocation 1 writes a uniquely-named marker into its cwd and re-checks it with the same `[ -e ]` predicate before exiting; invocation 2 probes for it and must report `ABSENT`. It deliberately probes for **one named file rather than asserting the directory is empty**: emptiness is a whole-directory property a concurrent invocation is entitled to break, whereas "the file A wrote is gone before B starts" is the property that matters and holds regardless of what else is in there. The writer's self-check is what stops `ABSENT` passing for the wrong reason (a mis-quoted path, a write that never landed).
* `test_leftover_in_the_shared_sandbox_is_gone_before_the_next_child_starts` — **negative control**, in the shape the bug actually took on disk: a file planted directly into the shared directory with `FileManager`, i.e. what a crashed session (or, before this change, any earlier invocation) leaves behind. It uses nothing but pre-#190 API (`run`, `sandboxDirectory()`), so it compiles against the unfixed runner and **fails there** — that runner swept nothing and the child prints `PROBE:FOUND`. The end-to-end test above fails on today's code too, for the same reason.
* `test_runner_spawns_child_in_isolated_empty_directory` — **this is the test that pinned the current behaviour and had to change.** It counted visible entries only (`ls`, not `ls -A`), on the grounds that a shared sandbox lets a CLI leave dot-files behind. That made the count pass by luck rather than by construction, and spared exactly the directory a CLI would cache in. It now uses `ls -A`, because the idle sweep makes emptiness true by construction for a run that starts with nothing else in flight.

**`MilaTests/LLMSandboxDirectoryTests.swift`**

* `test_sweep_waits_for_the_last_concurrent_invocation` — the concurrency constraint, driven through the lease API directly so the overlap is exact and there is no subprocess timing in it: a file written by the first tenant survives a second tenant's `acquire` **and** its `release`, and is gone only once the last tenant leaves.
* `test_a_oneshot_run_does_not_sweep_a_concurrent_tick` — the same property with real children, which the lease test cannot show: a long-running invocation (Live AI tick shape, `session: .new`) writes into the cwd and parks on a barrier while a one-shot call starts and finishes around it; the tick's file must still be there when it wakes. Both handshake files live outside the sandbox so the synchronisation writes nothing into the directory under test.
* The existing `test_concurrent_runs_share_one_working_directory` and `test_runner_leaves_the_shared_directory_in_place` are unchanged and must keep passing — the shared-cwd property from #187 is preserved, and the directory itself is never removed.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — not possible in this environment; relying on CI
- [x] User-facing change is noted in the README changelog (if applicable) — n/a, no user-visible behaviour change

---

---

> [!NOTE]
> **Medium Risk**
> Changes subprocess cwd lifecycle with locked file sweeps during overlapping LLM invocations; mistakes could delete in-use files or leave cross-run artifacts, though behavior is heavily regression-tested.
> 
> **Overview**
> Fixes **issue #190**: the stable shared LLM working directory (`llm-sandbox`) could retain scratch files or CLI dot-directories between runs, so a later unrelated invocation might see artifacts from an earlier recording.
> 
> **`LLMRunner`** now treats the sandbox with a reference-counted lease: `acquireSandbox()` / `releaseSandbox(_:)` sweep directory **contents** (including hidden entries) when the tenant count transitions on idle—before the first child enters and after the last one leaves. `executeProcess` holds that lease for the full child lifetime via `defer`, so concurrent Live AI and one-shot runs still share one cwd (#181) without a sweep deleting files a live process is using.
> 
> Documentation for `sandboxDirectory()` is updated to describe maintained emptiness rather than “empty by construction.” Failed sweeps log at error with counts only (no filenames).
> 
> Tests cover sequential isolation (`ls -A`), planted leftovers, lease/concurrency behavior, and an end-to-end overlapping tick vs one-shot scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 462707a9fde48cdfe4e01ed72e5a3d6166a2655c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -1389,20 +1389,30 @@ enum LLMRunner {
         // surprises users whose claude/cursor wrappers source nvm/asdf/etc.
         process.environment = childEnvironment(for: executable)
 
-        // CRITICAL: spawn in an empty directory Mila owns — never $HOME, `/`,
-        // or a user folder. macOS attributes any file access by the child
-        // process to *our* bundle ID, so if claude / cursor-agent walks the
-        // cwd looking for project files (which both do — particularly
+        // CRITICAL: spawn in a directory Mila owns that holds nothing — never
+        // $HOME, `/`, or a user folder. macOS attributes any file access by the
+        // child process to *our* bundle ID, so if claude / cursor-agent walks
+        // the cwd looking for project files (which both do — particularly
         // cursor-agent in `-f` mode), the user sees scary TCC prompts saying
         // "Mila would like to access Desktop / Downloads". Launching from an
-        // isolated, empty directory guarantees there's nothing for the LLM
-        // CLI to discover and reach for, so no permission prompts fire.
+        // isolated directory with nothing in it guarantees there's nothing for
+        // the LLM CLI to discover and reach for, so no permission prompts fire.
         //
         // The directory is SHARED and STABLE across every invocation — see
         // `sandboxDirectory()` for why (issue #181: claude derives its
         // `~/.claude/projects/<slug-of-cwd>` project directory from the cwd,
         // so a per-run cwd leaked a per-run project directory forever).
-        process.currentDirectoryURL = sandboxDirectory()
+        //
+        // Because it is shared, it is NOT empty by construction: a CLI that
+        // writes into its cwd leaves the file there for whoever comes next
+        // (issue #190). `acquireSandbox()` takes a reference-counted lease and
+        // empties the directory on the idle transitions, so a child that starts
+        // while nothing else is running still starts in an empty directory. The
+        // `defer` must span the child's whole life — the lease is what stops a
+        // concurrent invocation's sweep from deleting a file this one is using.
+        let sandbox = acquireSandbox()
+        defer { releaseSandbox(sandbox) }
+        process.currentDirectoryURL = sandbox
 
         // Close stdin immediately. Some CLIs (claude) read both stdin AND
         // argv; some (cursor-agent) ignore stdin entirely. We standardised
@@ -1540,9 +1550,10 @@ enum LLMRunner {
     /// Two requirements meet here, and a single stable directory is the only
     /// shape that satisfies both:
     ///
-    /// 1. **TCC isolation.** The child must start in an empty directory Mila
-    ///    owns, so a CLI that scans its cwd for project context finds nothing
-    ///    and no "Mila would like to access Desktop" prompts fire.
+    /// 1. **TCC isolation.** The child must start in a directory Mila owns
+    ///    that holds nothing, so a CLI that scans its cwd for project context
+    ///    finds nothing and no "Mila would like to access Desktop" prompts
+    ///    fire.
     /// 2. **One Claude Code project, not one per run** (issue #181). claude
     ///    stores its conversation transcripts in
     ///    `~/.claude/projects/<slug-of-cwd>/<session-uuid>.jsonl` — the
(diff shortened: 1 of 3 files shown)


urisland:
bugbot run


---

PR #236, opened 2026-09-01
feat(speakers): on-demand voice embedding from any recording

### LINKED ISSUE #235: On-demand voice embedding from any recording
## Problem

Voice profiles are only created when a speaker is named during or right after a live recording, because the embedding snapshot comes from the live diarizer pool. Older recordings, imports, and batch-transcribed files have no snapshot, so naming a speaker on them does nothing for voice recognition.

Similarly, after a batch transcription completes with diarized speakers, there is no automatic matching against stored voice profiles — the user has to manually name every speaker even if they are already known.

### On-demand embedding extraction (`embedSpeakers`)

A new `SpeakerDiarizer.embedSpeakers()` method that loads **only** the pyannote embedding model (~0.4s) and extracts a centroid from a known speaker segment. This is 100-1000x faster than the full diarization pipeline (which re-runs segmentation, clustering, etc.).

When a user names a speaker on a recording that has no snapshot, the longest segment for that speaker is extracted on demand and saved as a voice profile — same as if the speaker had been named during a live recording.

### Batch voice matching on transcription completion

After any transcription completes (live or re-transcribe), each speaker's longest segment is embedded and matched against stored voice profiles. Recognized speakers are auto-tagged immediately. The embeddings are also stored as snapshots so that manually correcting a name later still creates/updates a voice profile.

## Related

Extends the voice recognition work from #96 / PR #204.

noamleibovitch:
Closes #235

## Summary

- Adds `SpeakerDiarizer.embedSpeakers()` — loads only the pyannote embedding model (~0.4s) to extract voice centroids from known speaker segments, bypassing the full diarization pipeline (30-60min)
- When naming a speaker on an old recording with no snapshot, extracts the embedding on demand and saves the voice profile
- After any transcription completes, embeds all speakers and auto-matches against stored voice profiles for immediate auto-tagging

## Test plan

- [ ] Name a speaker on an old recording (never re-transcribed) — verify voice profile is created (~0.5s delay)
- [ ] Re-transcribe a recording with known speakers — verify they are auto-tagged
- [ ] Verify live recording flow is unchanged (snapshot path still preferred over on-demand)
- [ ] Verify opt-out (`voiceSettings.isConfigured = false`) blocks both paths

---

> [!NOTE]
> **Medium Risk**
> Changes how voice fingerprints are created, merged, and reversed across async Python subprocesses and persisted profiles; guards reduce double-learning and opt-out races but incorrect matching or profile math would affect speaker identity data.
> 
> **Overview**
> Adds a **fast embedding path** (`SpeakerDiarizer.embedSpeakers`) that runs only the pyannote embedding model on known segment time ranges (~sub-second) instead of full diarization.
> 
> **Naming speakers** still uses live/batch snapshots when present; otherwise it **extracts an embedding on demand** from the speaker’s longest segment in that recording (background utility task), records a snapshot, and updates the voice profile. Audio URLs are captured before compression can swap WAV for M4A, and `voiceSettings.isConfigured` is re-checked after the async work.
> 
> **Un-naming or renaming** now calls `onSpeakerUnnamed`, which **subtracts the merged observation** from the previous profile when a per-recording snapshot still exists (no adjustment if the snapshot was evicted).
> 
> After **transcription completes**, a chained hook (when voice recognition is on, profiles exist, and the recording has no names yet) embeds each speaker’s longest segment, **stores batch snapshots**, and **auto-applies stored profile names** via `setSpeakerName`. It **skips** recordings that already have live-stop snapshots so richer observations are not overwritten and profiles are not double-merged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca0a51c84b0f3a79ec3ef8a6767e038dc8696616. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Improved speaker identification by creating voice profiles automatically when a live or batch match is unavailable.
  - Added automatic voice matching after transcription completes.
  - Speaker names are now applied to recordings based on the closest stored voice profiles.
  - Renaming or removing a speaker name now correctly updates associated voice profile data.

- **Bug Fixes**
  - Improved handling of recordings and speaker segments that previously lacked usable voice-matching data.
  - Added more reliable processing across supported audio formats.

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -725,15 +725,153 @@ struct MilaApp: App {
         // weight already stored on disk, and folding it back counts it twice.
         // This is the only place voice profiles are written, so a recognised
         // speaker merges exactly once per recording.
-        store.onSpeakerNamed = { recordingID, rawID, name in
+        let voiceLog = os.Logger(subsystem: "io.island.whisper.IslandWhisper", category: "VoiceProfile")
+        store.onSpeakerNamed = { [weak store, weak diarSettings] recordingID, rawID, name in
             guard voiceSettings.isConfigured else { return }
-            guard let observed = voiceSnapshots.observation(forSpeaker: rawID,
-                                                           in: recordingID) else { return }
-            profileStoreRef.updateProfile(
-                name: name,
-                embedding: observed.observedCentroid,
-                sampleCount: observed.observedCount
-            )
+            if let observed = voiceSnapshots.observation(forSpeaker: rawID,
+                                                         in: recordingID) {
+                // Snapshot exists (live or batch) — save immediately.
+                profileStoreRef.updateProfile(
+                    name: name,
+                    embedding: observed.observedCentroid,
+                    sampleCount: observed.observedCount
+                )
+            } else if let store {
+                // No snapshot (old recording, never re-transcribed).
+                // Extract the embedding on demand (~0.5s) from the
+                // speaker's longest segment in this recording.
+                guard let rec = store.recordings.first(where: { $0.id == recordingID }) else { return }
+                var best: (start: Double, end: Double)?
+                for seg in rec.segments where seg.speaker == rawID {
+                    let dur = seg.end - seg.start
+                    if best == nil || dur > (best!.end - best!.start) {
+                        best = (seg.start, seg.end)
+                    }
+                }
+                guard let segment = best else { return }
+                let pythonPath = diarSettings?.pythonPath ?? "/usr/bin/python3"
+                let capturedName = name
+                let capturedRecID = recordingID
+                let capturedRawID = rawID
+                // Capture the audio URL now, before compression can
+                // change the recording's audioFileName to .m4a.
+                let wavURL = store.audioURL(for: rec)
+                Task.detached(priority: .utility) {
+                    guard FileManager.default.fileExists(atPath: wavURL.path) else { return }
+                    do {
+                        let embeddings = try await SpeakerDiarizer.embedSpeakers(
+                            wavURL: wavURL, pythonPath: pythonPath,
+                            speakerSegments: [capturedRawID: segment])
+                        guard let emb = embeddings[capturedRawID] else { return }
+                        await MainActor.run {
+                            // Re-check after await: user may have opted
+                            // out while the embedding was being extracted.
+                            guard voiceSettings.isConfigured else { return }
+                            voiceLog.log("on-demand voice profile saved for \(capturedRawID, privacy: .public)")
+                            voiceSnapshots.record(
+                                [(id: capturedRawID, observedCentroid: emb,
+                                  observedCount: 1, profileName: nil)],
+                                for: capturedRecID)
+                            profileStoreRef.updateProfile(
+                                name: capturedName,
+                                embedding: emb,
+                                sampleCount: 1)
+                        }
+                    } catch {
+                        voiceLog.log("on-demand embedding failed: \(error.localizedDescription, privacy: .private)")
+                    }
+                }
+            }
+        }
+
+        // Reverse a profile merge when a speaker is un-named or renamed.
+        // Looks up the snapshot for what was merged when the speaker was
+        // originally named; if available, subtracts that observation from
+        // the old profile. If the snapshot has been evicted, the name is
+        // still cleared but the profile is left untouched (safe default).
+        store.onSpeakerUnnamed = { recordingID, rawID, previousName in
+            guard voiceSettings.isConfigured else { return }
+            if let observed = voiceSnapshots.observation(forSpeaker: rawID,
+                                                         in: recordingID) {
+                profileStoreRef.subtractObservation(
+                    name: previousName,
+                    embedding: observed.observedCentroid,
+                    sampleCount: observed.observedCount)
+            } else {
+                voiceLog.log("un-name: no snapshot for \(rawID, privacy: .public) — profile not adjusted")
+            }
+        }
+
+        // Batch voice matching: after any transcription completes, embed
+        // each speaker's longest segment and match against stored voice
+        // profiles. Uses embedSpeakers (embedding model only, ~0.5s)
+        // instead of the full pyannote pipeline (30-60min).
+        //
+        // Skipped when the recording already has a snapshot from the live
+        // stop path (`RecognisedSpeakerAssigner.finish`), which stores
+        // richer multi-sample observations and already names recognised
+        // speakers. Without this guard the batch path would overwrite
+        // those observations with single-segment extracts and double-
+        // learn by firing `onSpeakerNamed` a second time.
+        let existingCallback = svc.onTranscriptionCompleted
+        svc.onTranscriptionCompleted = { [weak store, weak diarSettings] rec, wasRetranscription in
+            existingCallback?(rec, wasRetranscription)
+            guard voiceSettings.isConfigured,
+                  !profileStoreRef.profiles.isEmpty,
+                  rec.speakerNames.isEmpty,
+                  rec.segments.contains(where: { $0.speaker != nil }),
+                  // Skip if the live stop already snapshotted this
+                  // recording — its observations are richer and it
+                  // already named recognised speakers.
+                  voiceSnapshots.observation(forSpeaker: rec.segments.first(where: { $0.speaker != nil })!.speaker!, in: rec.id) == nil,
+                  let store else { return }
+            // Build a map of each speaker's longest segment.
+            var longest: [String: (start: Double, end: Double)] = [:]
+            for seg in rec.segments {
+                guard let sp = seg.speaker else { continue }
+                let dur = seg.end - seg.start
+                if let existing = longest[sp] {
+                    if dur > (existing.end - existing.start) {
+                        longest[sp] = (seg.start, seg.end)
+                    }
+                } else {
+                    longest[sp] = (seg.start, seg.end)
+                }
+            }
+            guard !longest.isEmpty else { return }
+            let pythonPath = diarSettings?.pythonPath ?? "/usr/bin/python3"
+            // Capture URL before the detached task — compression runs
+            // concurrently and may delete the WAV or change the path.
+            let wavURL = store.audioURL(for: rec)
+            Task.detached(priority: .utility) {
+                guard FileManager.default.fileExists(atPath: wavURL.path) else { return }
+                let embeddings = (try? await SpeakerDiarizer.embedSpeakers(
+                    wavURL: wavURL, pythonPath: pythonPath,
+                    speakerSegments: longest)) ?? [:]
+                guard !embeddings.isEmpty else { return }
+                await MainActor.run {
+                    // Re-check after await: user may have opted out.
+                    guard voiceSettings.isConfigured else { return }
+                    // Store embeddings as snapshots so naming a speaker
+                    // later (via onSpeakerNamed) can save the voice
+                    // profile even for batch-transcribed recordings.
+                    let entries = embeddings.map { (rawID, emb) in
+                        (id: rawID,
+                         observedCentroid: emb,
+                         observedCount: 1,
+                         profileName: nil as String?)
+                    }
+                    voiceSnapshots.record(entries, for: rec.id)
+
+                    for (rawID, emb) in embeddings {
+                        if let profile = profileStoreRef.match(embedding: emb) {
+                            store.setSpeakerName(
+                                profile.name, forSpeaker: rawID,
+                                recordingID: rec.id)
+                        }
+                    }
+                }
+            }
         }
 
         // Auto-drop accidental short+empty captures (issue #61). A recording
--- a/Mila/Transcription/SpeakerDiarizer.swift
+++ b/Mila/Transcription/SpeakerDiarizer.swift
@@ -356,6 +356,120 @@ enum SpeakerDiarizer {
         return try JSONDecoder().decode([SpeakerTurn].self, from: result.stdout)
     }
 
+    /// Embed specific speaker segments from an already-diarized recording.
+    /// Unlike `diarize` (which runs the full pyannote pipeline), this loads
+    /// ONLY the embedding model and extracts centroids from known speaker
+    /// segments. Takes ~0.4s to load + ~0.04s per speaker vs. 30-60+
+    /// minutes for full diarization.
+    ///
+    /// `speakerSegments` maps raw speaker IDs to their (start, end) time
+    /// ranges — typically the longest segment per speaker from the
+    /// already-completed transcription.
+    static func embedSpeakers(
+        wavURL: URL,
+        pythonPath: String,
+        speakerSegments: [String: (start: Double, end: Double)]
+    ) async throws -> [String: [Float]] {
+        guard let modelsPath = bundledModelsPath else { return [:] }
+        guard !speakerSegments.isEmpty else { return [:] }
+        var pythonInputURL = wavURL
+        var tempWAVToCleanUp: URL?
+        if wavURL.pathExtension.lowercased() != "wav" {
+            pythonInputURL = try AudioCompressor.decodeToTempWAV(wavURL)
+            tempWAVToCleanUp = pythonInputURL
+        }
+        defer {
+            if let temp = tempWAVToCleanUp { try? FileManager.default.removeItem(at: temp) }
+        }
+        let resolvedPython = resolvePython(userConfigured: pythonPath)
+        let extraEnv = pythonEnvironment()
+
+        // Encode speaker segments as JSON for the Python script.
+        let segmentsJSON = try JSONSerialization.data(
+            withJSONObject: speakerSegments.mapValues { ["start": $0.start, "end": $0.end] },
+            options: [])
+        let segmentsArg = String(data: segmentsJSON, encoding: .utf8) ?? "{}"
+
+        let script = """
+        import json, sys, os, types
+
+        try:
+            import speechbrain.utils.importutils as _sbiu
+            _orig_ensure = _sbiu.LazyModule.ensure_module
+            def _safe_ensure(self, *a, **kw):
+                try:
+                    return _orig_ensure(self, *a, **kw)
+                except ImportError:
+                    self.lazy_module = types.ModuleType(self.target)
+                    return self.lazy_module
+            _sbiu.LazyModule.ensure_module = _safe_ensure
+        except Exception:
+            pass
+
+        import torch
+        _orig_torch_load = torch.load
+        def _patched_torch_load(*args, **kwargs):
+            kwargs["weights_only"] = False
+            return _orig_torch_load(*args, **kwargs)
+        torch.load = _patched_torch_load
+
+        import soundfile as sf
+        import tempfile
+        from pyannote.audio import Pipeline
+
+        wav_path = sys.argv[1]
+        models_dir = sys.argv[2]
+        segments = json.loads(sys.argv[3])
+
+        config_path = os.path.join(models_dir, "config.yaml")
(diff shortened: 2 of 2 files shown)


urisland:
Thanks for this — the fast-embedding idea is a good one, and reusing `runPython` instead of hand-rolling a new subprocess call was the right instinct (details below). Review notes follow.

**How I reviewed this.** Static reading only. There is no Swift toolchain available to me, so nothing here was compiled or executed — every claim below is justified by reading the diff against the code on `main` (base `aec65fc`), and I say where I'm inferring rather than observing. CI has also not run: the only check on the PR is `Cursor Bugbot`; the workflow runs are sitting in `action_required` awaiting maintainer approval for an outside contributor, so there is no CI signal for either of us. I checked each CodeRabbit and Bugbot finding against the code rather than taking it on trust; some hold, some don't, and I've said which.

---

## Blocking

### 1. This branch does not compile on its own — it depends on #238

`MilaApp.swift:792` assigns `store.onSpeakerUnnamed` and `:796` calls `profileStoreRef.subtractObservation(...)`. Neither exists:

- `RecordingStore` on `main` declares exactly one such hook, `onSpeakerNamed` (`Mila/Models/RecordingStore.swift:29`). There is no `onSpeakerUnnamed`.
- `SpeakerProfileStore` has `updateProfile` / `deleteProfile` / `renameProfile` / `mergeProfiles` / `match`, and no `subtractObservation`.

I checked the head tree directly rather than inferring from the diff: at `ca0a51c`, `Mila/Models/RecordingStore.swift` contains zero occurrences of `onSpeakerUnnamed` and `Mila/Models/SpeakerProfileStore.swift` zero of `subtractObservation`. Both symbols are introduced by `eb962f3` — the second commit on the **#238** branch, which is a child of this PR's head. So #238 is a superset of #236, not a dependant of it, despite "Depends on #236" in its description.

A second, independent problem with the same block: even with the declaration in place, nothing on `main` would ever call it. `RecordingStore.setSpeakerName` (`:505-520`) fires `onSpeakerNamed` only on the non-empty branch, and the clear branch (`:512-515`) fires nothing at all. The invocation sites are also added by `eb962f3`.

Pick one:

- drop the `onSpeakerUnnamed` block (`MilaApp.swift:787-803`) from this PR so it stands alone, and let #238 add both the API and its wiring together; or
- close this PR and review the stack as #238 alone — but note #238 links only #237, and per `.claude/rules/pull-requests.md` you'd want #235 covered too.

Merge order matters either way: **merging #236 alone leaves `main` unbuildable.**

### 2. Batch auto-tagging overwrites a name the user typed while the embed was in flight

`MilaApp.swift:866-871`. The `rec.speakerNames.isEmpty` guard at `:821` is evaluated on the completion-time copy of the row, but the write at `:868` happens after an `await` that spans a Python process launch (torch + pyannote import, then inference). Nothing re-reads the live row.

Sequence:

1. Transcription completes. `rec.speakerNames` is empty — guaranteed, in fact: `TranscriptionService.swift:974` clears it on every pass that produced segments. The detached embed starts.
2. The user opens the recording and names `SPEAKER_00` "Dana". `setSpeakerName` writes it and fires `onSpeakerNamed`; no snapshot exists yet, so the on-demand path at `:739` starts a *second* embed.
3. The batch embed returns. `:867` matches `SPEAKER_00` to stored profile "Alice" and `:868` calls `setSpeakerName("Alice", …)`. "Alice" ≠ "Dana", so the no-change guard doesn't fire — the user's label is replaced, and `onSpeakerNamed` runs again, now finding the snapshot recorded at `:864`, and folds this recording's embedding into Alice.
4. The on-demand embed from step 2 returns and folds the same embedding into "Dana".

Net: the user's typed label is silently replaced, and one recording's single observation lands in two different profiles. That is exactly the failure the guard in `RecognisedSpeakerAssigner.finish` documents at length ("overwriting 'Bob' with 'Alice' here would fire `onSpeakerNamed` a second time and fold this recording's centroid into *both* profiles") — the auto-naming loop there skips any raw id the user already named, for this reason.

Fix is small: inside the `MainActor.run` at `:852`, re-read the live row and skip named speakers.

```swift
guard let live = store.recordings.first(where: { $0.id == rec.id }) else { return }
for (rawID, emb) in embeddings where live.speakerNames[rawID] == nil {
    …
}
```

(Bugbot flagged this as "Batch auto-tag overwrites user names". Confirmed.)

---

## Should fix before merge

### 3. `voiceSnapshots.record` replaces the whole per-recording map

Both bots raised this and both are right. `ObservedVoiceSnapshots.record` (`Mila/Models/ObservedVoiceSnapshots.swift:75-92`) does `byRecording.updateValue(observations, forKey: recordingID)` — a whole-dictionary replace, documented as "Re-snapshotting the same recording replaces its entry". Both new call sites pass a partial set: `:771-774` passes exactly one speaker, `:864` passes only the speakers that produced an embedding.

Reachable without any race, on an old recording with no snapshot:

1. Name `SPEAKER_00` "Alice" → on-demand path → `record([SPEAKER_00])`, `updateProfile("Alice", …)`.
2. Name `SPEAKER_01` "Bob" → `observation(forSpeaker: "SPEAKER_01", …)` is nil (that speaker has no entry, even though the recording does) → on-demand path again → `record([SPEAKER_01])` **replaces** the map. `SPEAKER_00`'s observation is gone.
3. Later un-name or rename Alice (once #238 lands): `onSpeakerUnnamed` finds no snapshot and logs "profile not adjusted" — the correction that block exists to perform silently doesn't happen.
4. Or re-name `SPEAKER_00`: no snapshot, so on-demand runs again and merges the same recording's embedding into a profile a second time — the double-merge that #204 fixed.

Two speakers named on one old recording is the ordinary case for this feature, so this isn't an edge. Needs an upsert on `ObservedVoiceSnapshots` (merge into the existing dict, keeping the recording's position in the eviction queue) rather than `record`.

### 4. The "skip if the live stop already snapshotted" guard is too weak, and it kills auto-tagging on re-transcription

`MilaApp.swift:826`. Two problems.

**It probes one speaker.** `voiceSnapshots.observation(forSpeaker: rec.segments.first(where: { $0.speaker != nil })!.speaker!, in: rec.id)` treats the first labelled speaker as a proxy for the whole recording. If that one speaker happens to have no entry while others do, the guard passes and `:864` replaces the richer live observations with single-sample extracts. (The force-unwraps themselves are safe — `guard` conditions short-circuit and `:822` establishes the same predicate — but the line is doing real logic inside a `guard` list and is hard to read; worth hoisting.)

**It permanently blocks re-transcription.** `TranscriptionService.swift:974` clears `speakerNames` and re-keys every `SPEAKER_NN` on each successful pass, so `:821` is always true. On a re-transcribe of a recording this hook already handled once, `:826` finds run 1's snapshot under the same `SPEAKER_00` string and bails — auto-tagging never runs again for that recording while it's in the LRU. Worse, run 1's snapshot is now keyed to run 2's re-minted ids, which pyannote assigns by first-appearance order and which may denote different people; naming one then persists the previous run's embedding. `ObservedVoiceSnapshots`' header comment exists precisely to stop positional ids resolving to the wrong voice — this reintroduces the same shape across a re-clustering boundary. (The stale-snapshot-across-re-key exposure predates this PR via the live-stop path; this PR makes it routine by snapshotting after every batch pass, and the skip guard makes it sticky.)

Rather than a proxy probe: refresh the snapshot on each pass instead of bailing, and use an explicit signal for "the live stop already handled this" (a flag set by `RecognisedSpeakerAssigner.finish`, or a check across all labelled speakers).

(Bugbot's "Snapshot skip blocks auto-tagging". Confirmed.)

### 5. The WAV-compression mitigation is backwards

The comment at `:843-844` says the URL is captured early so compression can't "delete the WAV or change the path". Capturing it early prevents neither. `TranscriptionService.swift:1040` starts `store.compressRecordingAudio(id:)` in a `Task` immediately after this callback returns; that method transcodes to `.m4a`, swaps `audioFileName`, and then does `try? FileManager.default.removeItem(at: src)` on the very `.wav` this URL points at. So the captured path either misses at `:847` or disappears while Python is mid-`sf.read`, and auto-tagging is silently dropped with no retry — for exactly the users who keep compression on, i.e. the default.

Resolving the URL *late* would be strictly better: `embedSpeakers` already handles non-WAV input by decoding to a temp WAV (`SpeakerDiarizer.swift:373-376`), so the `.m4a` works fine. Suggest resolving `store.audioURL(for:)` at use time inside the task (hop to the main actor for it), or re-resolving and retrying once when the captured path is gone.

Bugbot raised this on the earlier commit and the thread is marked resolved. It is not resolved — only the comment changed.

### 6. Every failure mode returns `[:]`, and nothing is logged

`SpeakerDiarizer.swift:469-470`: non-zero exit, empty stdout and a decode failure all collapse to an empty dictionary, and both callers treat that as "nothing to embed" and return. `diarize` in the same file throws `Error.diarizationFailed(stderr)` in the equivalent place (`:350-353`). The new inline script also emits no stderr diagnostics at all, unlike `diarize` (`:322`, `:326`, `:329`, `:340`). Combined with finding 5, the two most likely real-world failures — missing audio and a broken pyannote import — are indistinguishable from "no speakers matched", and a bug report would carry nothing usable.

CodeRabbit flagged this and its suggestion is sound: log through the existing redacted style. `SpeakerDiarizer.Error.logMessage(for:)` (`:88-96`) exists for exactly this and keeps the recording title out of the log.

Related, at `MilaApp.swift:781`: `error.localizedDescription` at `privacy: .private` renders the whole line as `<private>`, so the catch branch is undiagnosable. `bugbot-rules/no-user-content-in-logs.md` calls out over-redaction as its own bug and the fix is the same — `SpeakerDiarizer.Error.logMessage(for: error)` as `.public`.

### 7. Auto-tagging ignores the user's similarity setting

`:867` calls `profileStoreRef.match(embedding: emb)`, taking the default `threshold: 0.55`. This is the first production caller of `match` — until now only tests used it. The live path uses `aiSettings.speakerSimilarityThreshold` (`MilaApp.swift:1579` → `LiveSpeakerDiarizer.similarityThreshold`), user-configurable and defaulting to 0.55. A user who raises that slider to tighten recognition gets stricter live matching and an auto-tagger that still names people at 0.55. Auto-naming after the fact is where a tightened threshold matters most.

### 8. No tests

Nothing in the diff (2 files, +260/-8). This area is unusually well covered — `LiveSpeakerNamingTests`, `RecognisedSpeakerMergeTests`, `CrossRecordingVoiceIsolationTests`, `VoiceRecognitionGateTests`, `RecognisedSpeakerAssignerTests` — and several of those exist because of bugs in this exact shape (#204, #209). Findings 2, 3 and 4 are all testable at the `ObservedVoiceSnapshots` / `RecordingStore` / `SpeakerProfileStore` seam without any Python. A regression test for "batch auto-tag must not overwrite a name typed during the embed" would be the highest-value one.

---

## Verified as correct — noting these so nobody re-litigates them

- **Pipe-drain ordering is fine.** This was the thing I most wanted to check, given `.claude/rules/python-subprocess.md` and PR #15. `embedSpeakers` delegates to `SpeakerDiarizer.runPython` rather than driving `Process` itself, and that helper does it right (`SpeakerDiarizer.swift:149-152`):

  ```swift
  let stdoutRead = Task.detached { stdout.fileHandleForReading.readDataToEndOfFile() }
  let stderrRead = Task.detached { stderr.fileHandleForReading.readDataToEndOfFile() }

  process.waitUntilExit()
  ```

  Both pipes are drained in detached tasks before `waitUntilExit()`, the whole thing runs inside `Task.detached(priority: .userInitiated)` off the main actor, and cancellation reaches the subprocess. Reusing it rather than copying the pattern is the right call, and it's the reason this PR doesn't reintroduce the >25-minute hang.
- **stdout/stderr separation holds.** The new script writes JSON only to `sys.stdout` (`:459`) and the torch / speechbrain / pyannote chatter goes to stderr as always, read separately. Same shape as `diarize`. (Its total lack of stderr diagnostics is finding 6, not a separation problem.)
- **The two compatibility patches are present and correct** — the speechbrain `LazyModule.ensure_module` patch and the `torch.load` `weights_only=False` patch, both matching the existing copies in `diarize` and in `LiveSpeakerDiarizer.daemonScript`.
- **`getattr(pipeline, "_embedding", None)` and the `(1, 1, N)` tensor shape are right**, and not novel — they mirror `LiveSpeakerDiarizer.daemonScript` exactly, including the comment there explaining why the tensor and not a `{"waveform": …}` dict. Which also means the embeddings this produces live in the same space as the ones already in stored profiles; I checked that specifically because a mismatch would have made every comparison meaningless.
- **The feature gate is respected.** Every entry point guards on `voiceSettings.isConfigured` (the `isEnabled && diarizationReady` gate), not on `isEnabled`, and both async paths re-check it after the await before writing. That satisfies `.claude/rules/feature-gates.md`.
- **No `Recording.encode(to:)` change**, so no `MilaKit.StoredRecording` mirroring is needed and `StoredRecordingDriftTests` is not in play.
- **`privacy: .public` on `capturedRawID` at `:770` and `:801` is safe.** I chased this because `SpeakerProfileStore.seedEntries()` returns `id: p.name` — a person's name. But `LiveSpeakerDiarizer.seedPool` discards that and re-mints `String(format: "SPEAKER_%02d", …)`, so raw ids are always positional. No leak.
- **Chaining `onTranscriptionCompleted` is done correctly** — `existingCallback` is captured and invoked first at `:818`, and `:602` is the only other assignment, so nothing clobbers it.

## Rejected / downgraded bot findings

- **CodeRabbit, "batch `record` replaces richer live observations" (`:864`, still open).** Largely mitigated by the skip guard you added at `:826` — in the common live-stop case the hook bails before reaching `:864`. The residual is real but it is the single-speaker probe, which is finding 4; the fix belongs there, not at `:864`.
- **Bugbot, "Embed tasks ignore later revocation" (marked resolved).** Mostly addressed: `isConfigured` is now re-checked inside both `MainActor.run` blocks. The part still standing is small — a recording hard-deleted or trashed mid-embed still gets its embedding written into `voiceSnapshots` (`setSpeakerName` no-ops on a missing row, so nothing worse happens). Low priority; worth a line rather than a fix.
- **Bugbot, "Auto-tag races with WAV compression" (marked resolved).** Not resolved — see finding 5. The mitigation is in the wrong direction.
- **CodeRabbit's "Docstring Coverage 50%" pre-merge check.** Noise. Both new functions carry doc comments; `embedSpeakers` has a good one.
- **CodeRabbit's docstring nitpick is right, though:** `Pipeline.from_pretrained` instantiates the whole `SpeakerDiarization` pipeline, segmentation model included. The docstring at `SpeakerDiarizer.swift:359-367` says "loads ONLY the embedding model". The speed claim still roughly holds (you skip *running* segmentation and clustering, which is where the minutes go), but the sentence as written is wrong and will mislead the next reader.

## Smaller notes

- `sf.read(wav_path, dtype="float32")` at `:439` loads the entire recording into memory. A 3-hour 16 kHz mono capture is ~690 MB of float32, and this now runs after every completed transcription, concurrently with the compression pass reading the same file. `soundfile` can seek — reading only the requested spans would make this proportional to the segments rather than to the recording.
- The daemon moves the pipeline to MPS when available (`LiveSpeakerDiarizer.daemonScript`); the new script doesn't. Not a correctness issue, just speed left on the table.
- Both paths embed the *longest transcript segment*, not a VAD-trimmed utterance the way the live daemon does. A long whisper segment can straddle a speaker change or carry leading silence, so these embeddings are noisier than the live ones — which matters given they are merged into the same centroids at `sampleCount: 1`. Worth a sentence in the docstring so the next person knows the two sources aren't equivalent in quality.

## On #238

They overlap heavily and the maintainer needs to know the order. #238's branch (`eb962f3`) is a **child of this PR's head** (`ca0a51c`), and both PRs target `main`, so #238's diff already contains everything here plus the `onSpeakerUnnamed` / `subtractObservation` API this PR calls. Consequences:

- Merging **#236 alone breaks the build** (finding 1).
- Merging **#236 then #238** works; #238 reduces to just the `eb962f3` changes.
- Merging **#238 alone** subsumes #236 entirely, leaving this PR empty — but then #235 has no PR closing it.
- Findings 2 through 7 live in code that both PRs carry, so whichever lands second inherits them. Fixing them once, in whichever branch merges first, is the cheap path.

`eb962f3` also raises `ObservedVoiceSnapshots`' retention from 8 to 20 and touches the same file finding 3 wants an upsert added to, so that fix probably belongs in #238.

---

Overall: the shape is sound, and the subprocess handling — the part with the worst history in this repo — is correct. What's in the way is finding 1 (mechanical) and findings 2 and 3, which write to the user's speaker labels and voice profiles on paths that are ordinary rather than exotic. Given the reading-only constraint, please treat my sequences as hypotheses to confirm on a real build; they are derived from the code, but none of them has been observed running.

---

urisland:
Closing this in favour of **#253**, which lands your work — this is not a rejection, and I want to be precise about what is happening to it.

**Your commit is in #253, with you as the author.** `ca0a51c` was cherry-picked, not squashed or rewritten; `git log --format='%an'` on that branch shows you as the author of the first two commits, and the fixes sit on top as separate commits. That is deliberate: this is your contribution and the history should say so.

**Why the two PRs became one.** `eb962f3` (#238's head) turns out to be a *descendant* of `ca0a51c` — `git merge-base --is-ancestor ca0a51c eb962f3` passes — so the two PRs were one stack, not two independent changes. That made every merge order wrong on its own: #236 alone leaves `main` unbuildable, because `MilaApp.swift:792` assigns `store.onSpeakerUnnamed` and `:796` calls `subtractObservation`, and both of those are introduced by #238's commit. #238 alone builds but subsumes this PR entirely, leaving #235 with nothing closing it. Combining them is the only ordering that closes both issues and compiles at every commit. Nothing about that is a criticism of the split — it is an artefact of the branches sharing a base, and it is only visible once you check the ancestry.

**What this PR got right, specifically.** Reusing `SpeakerDiarizer.runPython` instead of hand-rolling a second `Process` is the thing I most wanted to check and the thing that would have been easiest to get wrong. This repo has a scar there: PR #15, where transcription hung at 100% on anything over ~25 minutes because pyannote's stderr filled a ~64 KB pipe buffer while the parent sat in `waitUntilExit()`. `runPython` drains both pipes in detached tasks *before* the wait, runs off the main actor, and propagates cancellation to the subprocess — and by delegating to it, `embedSpeakers` inherits all of that instead of reintroducing the deadlock. The stdout/stderr separation is clean, both compatibility monkey-patches (`torch.load` `weights_only`, speechbrain's `LazyModule.ensure_module`) are present and correct, and `getattr(pipeline, "_embedding", None)` with the `(1, 1, N)` tensor shape mirrors `LiveSpeakerDiarizer.daemonScript` exactly — which is what makes these embeddings live in the same space as the stored profiles rather than being quietly incomparable. I checked that last one specifically, because a mismatch there would have made every comparison meaningless while still looking like it worked.

The feature gate holds too: every entry point guards on `voiceSettings.isConfigured` rather than `isEnabled`, and both async paths re-check after the await, which is what `.claude/rules/feature-gates.md` asks for.

**What changed in #253.** The findings from my review here, plus #238's. The substantive ones were the ordering hazards across the `await` — the URL captured before compression can delete it, the name re-read after the embed, the snapshot upsert, the stale snapshot surviving a re-key — and `embedSpeakers` collapsing every failure into `[:]`, which made a broken pyannote import indistinguishable from "nobody matched". The embedding paths also moved out of `MilaApp.init` into an `OfflineVoiceEmbedder`, because closures in an `App` initializer are unreachable from a unit test and this code writes to voice profiles across a Python call; with the embed injected, all four race-shaped findings became ordinary unit tests. Full list in #253's description.

Thanks for this — the fast-embedding idea is a good one and the shape was sound.

---

PR #238, opened 2026-09-01
feat(speakers): un-naming / renaming corrects voice profiles + segment reassignment

### LINKED ISSUE #237: Un-naming / renaming a speaker should correct voice profiles
## Problem

When a speaker is auto-tagged incorrectly (false recognition), the stranger's voice embedding gets merged into the stored profile via weighted average. Today, un-naming (resetting to "Speaker A") or renaming (Alice → Bob) clears the display name but does not touch the profile. The profile stays polluted, leading to more false matches over time.

Additionally, when the diarizer splits one person into two speakers (SPEAKER_00 and SPEAKER_01), or mis-tags a single segment, there's no way to fix it without re-transcribing.

### Profile correction on un-name / rename

- **Un-naming** (reset to default): subtract the recording's observed embedding from the profile
- **Renaming** (Alice → Bob): subtract from Alice's profile, merge into Bob's
- Uses the existing `ObservedVoiceSnapshots` to look up what was merged

### Per-segment speaker reassignment

Right-click a segment to reassign it to a different speaker in the recording. Fixes diarization errors without re-transcribing.

### Merge speakers in a recording

When the diarizer splits one person into two speakers, merge all segments of one into the other. Equivalent to reassigning all segments + un-naming the source.

## Related

Builds on the voice recognition work from #96 / PR #204 and on-demand embedding from #235.

noamleibovitch:
Closes #237

## Summary

- **Profile correction on un-name/rename**: subtracts the recording's observed embedding from the old profile via `SpeakerProfileStore.subtractObservation`, reversing a false recognition's pollution
- **Rename handling**: renaming Alice to Bob subtracts from Alice's profile first, then merges into Bob's
- **Auto-merge on duplicate name**: naming a speaker with a name already used by another speaker in the same recording automatically merges all segments into that speaker
- **Split this line**: speaker label popover includes "Split this line" to move a single mis-tagged segment to a new unnamed speaker, which can then be named normally
- **Snapshot retention**: raised from 8 to 20 recordings so un-naming can still find the embedding to subtract
- **Review fixes**: re-check `isConfigured` after await in detached tasks, skip batch matching when live snapshot exists (prevents double-learn), capture audio URL before compression can race

Depends on #236 (on-demand voice embedding).

## Test plan

- [ ] Un-name an auto-tagged speaker ("Use default") — verify profile sample count decreases
- [ ] Rename Alice to Bob — verify Alice loses samples, Bob gains
- [ ] Name Speaker B the same as Speaker A — verify segments merge automatically
- [ ] Click speaker label → "Split this line" — verify segment becomes a new unnamed speaker
- [ ] Name the split speaker same as an existing one — verify auto-merge
- [ ] Live recording flow unchanged — name during recording, stop, verify profile created
- [ ] Batch transcription auto-tagging still works
- [ ] Opt out mid-recording — verify no embeddings written after opt-out

---

> [!NOTE]
> **Medium Risk**
> Changes how voice fingerprints are learned, reversed, and auto-applied across live, batch, and on-demand paths; mistakes could mis-tag speakers or corrupt profiles, though guards limit double-learning and opt-out races.
> 
> **Overview**
> **Voice profiles now stay aligned with speaker labels.** Clearing a name or renaming a speaker triggers `onSpeakerUnnamed`, which uses the per-recording snapshot to call `SpeakerProfileStore.subtractObservation` and undo that recording’s contribution (or deletes the profile if samples hit zero). Renames subtract from the old name before `onSpeakerNamed` merges into the new one. Snapshot retention rises from **8 → 20** recordings so corrections still find embeddings after more back-to-back captures.
> 
> **Naming and matching got richer paths.** If there’s no snapshot (e.g. old recording), naming kicks off **on-demand** `SpeakerDiarizer.embedSpeakers` on the longest segment in a detached task, with audio URL captured early and `isConfigured` re-checked after await. After batch transcription, a chained `onTranscriptionCompleted` hook embeds longest segments, stores snapshots, and auto-applies stored profile names—**skipped** when a live-stop snapshot already exists to avoid overwriting richer observations or double-learning.
> 
> **Transcript speaker editing:** assigning a name already used by another speaker in the same recording **merges** segments via `mergeSpeakers`; the detail view adds **“Split this line”** (`splitSegmentSpeaker`) to give one segment a new `SPEAKER_NN` id. `embedSpeakers` is new—a fast embedding-only Python path vs full diarization.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb962f3c28d68523cf836d54f5a5bc2ec945861b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Split an individual transcript line into a separate speaker identity.
  * Voice profiles now retain observations from up to 20 recordings.

* **Bug Fixes**
  * Improved voice embedding reliability during concurrent recording and compression.
  * Prevented existing live observations from being overwritten during batch matching.
  * Reversed profile learning when speaker assignments change, preserving recognition accuracy.
  * Rechecked voice-recognition settings before recording or applying extracted results.
  * Avoided duplicate profile learning for recordings with existing speaker observations.

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -725,15 +725,153 @@ struct MilaApp: App {
         // weight already stored on disk, and folding it back counts it twice.
         // This is the only place voice profiles are written, so a recognised
         // speaker merges exactly once per recording.
-        store.onSpeakerNamed = { recordingID, rawID, name in
+        let voiceLog = os.Logger(subsystem: "io.island.whisper.IslandWhisper", category: "VoiceProfile")
+        store.onSpeakerNamed = { [weak store, weak diarSettings] recordingID, rawID, name in
             guard voiceSettings.isConfigured else { return }
-            guard let observed = voiceSnapshots.observation(forSpeaker: rawID,
-                                                           in: recordingID) else { return }
-            profileStoreRef.updateProfile(
-                name: name,
-                embedding: observed.observedCentroid,
-                sampleCount: observed.observedCount
-            )
+            if let observed = voiceSnapshots.observation(forSpeaker: rawID,
+                                                         in: recordingID) {
+                // Snapshot exists (live or batch) — save immediately.
+                profileStoreRef.updateProfile(
+                    name: name,
+                    embedding: observed.observedCentroid,
+                    sampleCount: observed.observedCount
+                )
+            } else if let store {
+                // No snapshot (old recording, never re-transcribed).
+                // Extract the embedding on demand (~0.5s) from the
+                // speaker's longest segment in this recording.
+                guard let rec = store.recordings.first(where: { $0.id == recordingID }) else { return }
+                var best: (start: Double, end: Double)?
+                for seg in rec.segments where seg.speaker == rawID {
+                    let dur = seg.end - seg.start
+                    if best == nil || dur > (best!.end - best!.start) {
+                        best = (seg.start, seg.end)
+                    }
+                }
+                guard let segment = best else { return }
+                let pythonPath = diarSettings?.pythonPath ?? "/usr/bin/python3"
+                let capturedName = name
+                let capturedRecID = recordingID
+                let capturedRawID = rawID
+                // Capture the audio URL now, before compression can
+                // change the recording's audioFileName to .m4a.
+                let wavURL = store.audioURL(for: rec)
+                Task.detached(priority: .utility) {
+                    guard FileManager.default.fileExists(atPath: wavURL.path) else { return }
+                    do {
+                        let embeddings = try await SpeakerDiarizer.embedSpeakers(
+                            wavURL: wavURL, pythonPath: pythonPath,
+                            speakerSegments: [capturedRawID: segment])
+                        guard let emb = embeddings[capturedRawID] else { return }
+                        await MainActor.run {
+                            // Re-check after await: user may have opted
+                            // out while the embedding was being extracted.
+                            guard voiceSettings.isConfigured else { return }
+                            voiceLog.log("on-demand voice profile saved for \(capturedRawID, privacy: .public)")
+                            voiceSnapshots.record(
+                                [(id: capturedRawID, observedCentroid: emb,
+                                  observedCount: 1, profileName: nil)],
+                                for: capturedRecID)
+                            profileStoreRef.updateProfile(
+                                name: capturedName,
+                                embedding: emb,
+                                sampleCount: 1)
+                        }
+                    } catch {
+                        voiceLog.log("on-demand embedding failed: \(error.localizedDescription, privacy: .private)")
+                    }
+                }
+            }
+        }
+
+        // Reverse a profile merge when a speaker is un-named or renamed.
+        // Looks up the snapshot for what was merged when the speaker was
+        // originally named; if available, subtracts that observation from
+        // the old profile. If the snapshot has been evicted, the name is
+        // still cleared but the profile is left untouched (safe default).
+        store.onSpeakerUnnamed = { recordingID, rawID, previousName in
+            guard voiceSettings.isConfigured else { return }
+            if let observed = voiceSnapshots.observation(forSpeaker: rawID,
+                                                         in: recordingID) {
+                profileStoreRef.subtractObservation(
+                    name: previousName,
+                    embedding: observed.observedCentroid,
+                    sampleCount: observed.observedCount)
+            } else {
+                voiceLog.log("un-name: no snapshot for \(rawID, privacy: .public) — profile not adjusted")
+            }
+        }
+
+        // Batch voice matching: after any transcription completes, embed
+        // each speaker's longest segment and match against stored voice
+        // profiles. Uses embedSpeakers (embedding model only, ~0.5s)
+        // instead of the full pyannote pipeline (30-60min).
+        //
+        // Skipped when the recording already has a snapshot from the live
+        // stop path (`RecognisedSpeakerAssigner.finish`), which stores
+        // richer multi-sample observations and already names recognised
+        // speakers. Without this guard the batch path would overwrite
+        // those observations with single-segment extracts and double-
+        // learn by firing `onSpeakerNamed` a second time.
+        let existingCallback = svc.onTranscriptionCompleted
+        svc.onTranscriptionCompleted = { [weak store, weak diarSettings] rec, wasRetranscription in
+            existingCallback?(rec, wasRetranscription)
+            guard voiceSettings.isConfigured,
+                  !profileStoreRef.profiles.isEmpty,
+                  rec.speakerNames.isEmpty,
+                  rec.segments.contains(where: { $0.speaker != nil }),
+                  // Skip if the live stop already snapshotted this
+                  // recording — its observations are richer and it
+                  // already named recognised speakers.
+                  voiceSnapshots.observation(forSpeaker: rec.segments.first(where: { $0.speaker != nil })!.speaker!, in: rec.id) == nil,
+                  let store else { return }
+            // Build a map of each speaker's longest segment.
+            var longest: [String: (start: Double, end: Double)] = [:]
+            for seg in rec.segments {
+                guard let sp = seg.speaker else { continue }
+                let dur = seg.end - seg.start
+                if let existing = longest[sp] {
+                    if dur > (existing.end - existing.start) {
+                        longest[sp] = (seg.start, seg.end)
+                    }
+                } else {
+                    longest[sp] = (seg.start, seg.end)
+                }
+            }
+            guard !longest.isEmpty else { return }
+            let pythonPath = diarSettings?.pythonPath ?? "/usr/bin/python3"
+            // Capture URL before the detached task — compression runs
+            // concurrently and may delete the WAV or change the path.
+            let wavURL = store.audioURL(for: rec)
+            Task.detached(priority: .utility) {
+                guard FileManager.default.fileExists(atPath: wavURL.path) else { return }
+                let embeddings = (try? await SpeakerDiarizer.embedSpeakers(
+                    wavURL: wavURL, pythonPath: pythonPath,
+                    speakerSegments: longest)) ?? [:]
+                guard !embeddings.isEmpty else { return }
+                await MainActor.run {
+                    // Re-check after await: user may have opted out.
+                    guard voiceSettings.isConfigured else { return }
+                    // Store embeddings as snapshots so naming a speaker
+                    // later (via onSpeakerNamed) can save the voice
+                    // profile even for batch-transcribed recordings.
+                    let entries = embeddings.map { (rawID, emb) in
+                        (id: rawID,
+                         observedCentroid: emb,
+                         observedCount: 1,
+                         profileName: nil as String?)
+                    }
+                    voiceSnapshots.record(entries, for: rec.id)
+
+                    for (rawID, emb) in embeddings {
+                        if let profile = profileStoreRef.match(embedding: emb) {
+                            store.setSpeakerName(
+                                profile.name, forSpeaker: rawID,
+                                recordingID: rec.id)
+                        }
+                    }
+                }
+            }
         }
 
         // Auto-drop accidental short+empty captures (issue #61). A recording
--- a/Mila/Models/ObservedVoiceSnapshots.swift
+++ b/Mila/Models/ObservedVoiceSnapshots.swift
@@ -40,17 +40,17 @@ final class ObservedVoiceSnapshots {
         let profileName: String?
     }
 
-    /// How many recordings' snapshots to keep. Naming happens right after a
-    /// recording in the overwhelming majority of cases; a handful of slots
-    /// covers "record a few back-to-back, then go back and label them"
-    /// without letting embeddings accumulate for the life of the process.
+    /// How many recordings' snapshots to keep. Raised from 8 to 20 so
+    /// un-naming a speaker (to correct a false recognition) can still find
+    /// the snapshot and subtract it from the voice profile. Memory cost is
+    /// negligible (~1 KB per speaker per recording).
     private let limit: Int
 
     private var byRecording: [UUID: [String: Observation]] = [:]
     /// Insertion order, oldest first — the eviction queue.
     private var order: [UUID] = []
 
-    init(limit: Int = 8) {
+    init(limit: Int = 20) {
         self.limit = max(1, limit)
     }
 
--- a/Mila/Models/RecordingStore.swift
+++ b/Mila/Models/RecordingStore.swift
@@ -28,6 +28,12 @@ final class RecordingStore: ObservableObject {
     /// Used by voice recognition to save voice profiles.
     var onSpeakerNamed: ((_ recordingID: UUID, _ rawID: String, _ name: String) -> Void)?
 
+    /// Called when a speaker name is cleared or replaced. Provides the
+    /// recording ID, raw speaker ID, and the *previous* name that was
+    /// removed. Used by voice recognition to subtract the recording's
+    /// observed embedding from the old profile.
+    var onSpeakerUnnamed: ((_ recordingID: UUID, _ rawID: String, _ previousName: String) -> Void)?
+
     private let fileManager = FileManager.default
     /// `storeURL` and `foldersURL` move with `recordingsDirectory` on
     /// every `relocateRecordings` call. On the default path they live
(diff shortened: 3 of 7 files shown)


urisland:
Thanks for this — the core idea is right, and `subtractObservation` itself is a correct algebraic inverse of `updateProfile`'s weighted fold (read side by side: `SpeakerProfileStore.swift:344-346` vs `:288-289`; they diverge only at the `maxSampleCount` clamp, which the existing doc comment establishes is unreachable). Almost everything below is in the wiring around it, not in that function.

**How I reviewed this.** Static reading of `eb962f3` only. There is no Swift toolchain available to me, so nothing here was compiled or run and no claim below rests on execution. There is also no CI signal: the workflow runs on this PR sit in `action_required` because GitHub requires a maintainer to approve workflows for outside contributors, so nothing has been built or tested by the repo either. Line numbers refer to the files as they stand at the head commit.

## Blocking

### 1. Batch auto-match can destructively merge two speakers with no user action

`MilaApp.swift:867-870` loops matched embeddings and calls `store.setSpeakerName(profile.name, forSpeaker: rawID, …)`. `match` (`SpeakerProfileStore.swift:476-486`) returns the best profile above the threshold — nothing prevents two different raw IDs from resolving to the *same* profile. `setSpeakerName` then hits the new duplicate-name branch at `RecordingStore.swift:518-521` and calls `mergeSpeakers`, which rewrites `segments[i].speaker` in place (`:548-552`), persists, and regenerates the `.srt`.

Sequence: enable voice recognition, have at least one stored profile, transcribe a two-speaker file (this hook runs on every completed batch transcription that has speakers and no names yet). If both clusters land above 0.55 cosine against that one profile — not exotic for two similar voices — the first iteration names `SPEAKER_00` "Alice" and the second silently folds every `SPEAKER_01` segment into `SPEAKER_00`. The user never asked for a merge, there is no undo, and "Split this line" only unpicks it one line at a time. Because `embeddings` is a `Dictionary`, which cluster survives is not even stable between runs.

The same shape is reachable from the live path — `RecognisedSpeakerAssigner.swift:208` also loops `store.setSpeakerName(profileName, …)`.

Putting merge *inside* `setSpeakerName` means every automatic caller inherits it. I'd move merge out into an explicit user action (which is also what #237 asked for) and leave `setSpeakerName` as a pure label write.

### 2. Re-transcription leaves a stale snapshot, and this PR now protects it

`TranscriptionService.swift:968-973` clears `speakerNames` on any pass that produced segments, with the comment: *"The batch pass minted fresh speaker IDs (new clustering, new segment boundaries) with no alignment to the old ones, so any user-assigned names are now keyed to meaningless IDs — clear them rather than mislabel a different voice."*

The **embeddings** keyed to those same meaningless IDs are not cleared. Grepping `voiceSnapshots.` across `MilaApp.swift` gives sites at `647, 731, 771, 794, 826, 864` and none of them invalidate a recording; the only clearing anywhere is `clearOnOptOut`. That gap pre-dates this PR, but this PR widens it three ways at once:

- retention goes 8 → 20 (`ObservedVoiceSnapshots.swift:53`), so a stale snapshot survives 2.5× longer;
- the batch path now records snapshots for recordings that previously had none (`MilaApp.swift:864`), extending the hazard from live recordings to imported/batch ones;
- `MilaApp.swift:826` explicitly *skips* batch matching when a snapshot already exists, so the fresh embedding can never replace the stale one.

Sequence: record with diarization → stop (live snapshot stored under run-1's `SPEAKER_00`/`SPEAKER_01`) → right-click → Re-transcribe → names cleared and IDs re-keyed → name `SPEAKER_00` "Alice". `MilaApp.swift:731` finds the run-1 snapshot for `SPEAKER_00` and `updateProfile` writes run-1's `SPEAKER_00` embedding — after re-clustering, potentially a different person — into Alice's profile. That is precisely the failure mode `ObservedVoiceSnapshots`' header comment says the recording-id keying exists to prevent.

Fix shape: invalidate the recording's snapshot at the top of the completion hook whenever the pass re-keyed segments, and let the batch path re-record.

## Also confirmed

### 3. A second on-demand embed wipes the first one's snapshot

`ObservedVoiceSnapshots.record` (`:74-93`) *replaces* the whole per-recording dictionary via `byRecording.updateValue(observations, …)`. The on-demand path at `MilaApp.swift:771-774` passes exactly one entry.

Sequence: old recording with no snapshot. Name `SPEAKER_00` "Alice" → detached embed A starts. Name `SPEAKER_01` "Bob" → `observation(forSpeaker: "SPEAKER_01")` is still nil, so embed B starts too. A lands: `byRecording[rec] == ["SPEAKER_00": …]`. B lands: `byRecording[rec] == ["SPEAKER_01": …]`. Un-naming Alice now falls to `MilaApp.swift:801` ("no snapshot … profile not adjusted") and her profile keeps the pollution — the exact thing #237 exists to remove. `record` needs a merge variant for single-speaker writes.

(Cursor raised this and the thread shows resolved; the code at head is unchanged.)

### 4. The on-demand embed writes a profile for a name the user already cleared, unrecoverably

`MilaApp.swift:766-778`: the only re-check after the `await` is `voiceSettings.isConfigured`.

Sequence: old recording, name `SPEAKER_00` "Alice" → the detached task spawns Python and loads torch + the pyannote pipeline (seconds, not milliseconds) → the user immediately picks "Use default". `onSpeakerUnnamed` runs, finds no snapshot, logs and does nothing. The task then lands and does `record(...)` + `updateProfile("Alice", emb, 1)`. Alice's profile now carries an embedding for a speaker the user explicitly un-named — and un-naming again is a no-op, because `RecordingStore.swift:530-531` returns early when there is no previous name. The UI offers no route back.

Re-read `store.recordings…speakerNames[capturedRawID] == capturedName` inside the `MainActor.run` before writing.

The batch path has the mirror problem: the `rec.speakerNames.isEmpty` guard at `:823` is evaluated against the value captured before the embed, so a name typed by hand during the embed is overwritten by the auto-match at `:868` (and, per #1, possibly merged away).

### 5. The auto-merge subtracts from the old profile but never adds to the new one

CodeRabbit's finding, and it holds. `RecordingStore.swift:518-521` returns immediately after `mergeSpeakers`, so `onSpeakerNamed` never fires; `mergeSpeakers:555-557` does fire `onSpeakerUnnamed`. Rename `SPEAKER_01` from "Bob" to "Alice" while `SPEAKER_00` is already "Alice": Bob's profile loses the observation and Alice's gains nothing. The non-merge rename branch four lines below (`:522-528`) does subtract-then-add, so the two branches disagree. The snapshot is still keyed to the source raw ID after the merge, so firing `onSpeakerNamed?(recordingID, rawID, trimmed)` would resolve correctly.

### 6. "Split this line" recycles `SPEAKER_NN` ids and strands names

`RecordingStore.swift:573-577` picks the lowest id not used by any *segment*. `speakerNames` is not consulted, and orphaned entries are never cleaned up.

Sequence: segments `SPEAKER_00` ×3 and `SPEAKER_01` ×1; name `SPEAKER_01` "Bob"; click "Split this line" on that single `SPEAKER_01` segment. `used == {SPEAKER_00, SPEAKER_01}`, so it becomes `SPEAKER_02`; no segment carries `SPEAKER_01` any more, but `speakerNames["SPEAKER_01"] == "Bob"` survives. Now split any `SPEAKER_00` line: `used == {SPEAKER_00, SPEAKER_02}`, next free is `SPEAKER_01` — and that line renders as "Bob", in the transcript and in the regenerated `.srt`.

The stranded `ObservedVoiceSnapshots` entry for a split-away or merged-away id is the same hazard one level down: a recycled id resolves to the previous speaker's embedding.

Two cheap fixes: exclude keys of `speakerNames` from `used`, and no-op the split when the segment is already the only one with that speaker.

### 7. The compression race is not actually closed by capturing the URL early

`TranscriptionService.swift:1029` fires `onTranscriptionCompleted` and then immediately spawns `Task { await store.compressRecordingAudio(id:) }` (`:1037-1038`), which transcodes and then `try? FileManager.default.removeItem(at: src)` (`RecordingStore.swift:357`). The new hook captures the WAV URL at `MilaApp.swift:845` and checks `FileManager.default.fileExists(atPath: wavURL.path)` at `:847` from inside a `Task.detached(priority: .utility)` — lower priority than the compression task. Capturing the URL earlier does not stop the file being deleted; it only guarantees the check looks at the path that is about to vanish. When compression wins, batch matching silently does nothing.

Capture `rec.id` and resolve `store.audioURL(for:)` inside the task instead — `embedSpeakers` already handles non-WAV input at `SpeakerDiarizer.swift:376-380` via `AudioCompressor.decodeToTempWAV`, which also normalises to 16 kHz.

### 8. `embedSpeakers` discards every diagnostic it has

`SpeakerDiarizer.swift:469-470` returns `[:]` on a non-zero exit and swallows the decode error, and `result.stderr` is never read at all. `:434-437` writes `{}` and exits 0 when `pipeline._embedding` is missing. Compare `diarize` at `:352-355`, which throws `Error.diarizationFailed` carrying stderr. `.claude/rules/python-subprocess.md` asks for stderr to be logged specifically because pyannote internals have already broken this codebase twice (PRs #14, #15). As written, a pyannote release that renames `_embedding` turns both new features into silent no-ops with nothing in the log — and both call sites then `try?` / `?? [:]` on top of that.

## Bot findings I checked and am rejecting

- **CodeRabbit, "`mergeSpeakers` loses the label when the target is unnamed" (`RecordingStore.swift:555`).** Does not hold at this head. It was written against `ede627b`, which had a "Merge all into…" context menu; that UI is gone — `grep -c 'Merge all into\|Reassign to' Mila/Views/RecordingDetailView.swift` is **0** at `eb962f3`. The only caller is `RecordingStore.swift:519`, where `existingRaw` is by construction a key of `speakerNames`, so the target always has a name. The defensive branch is still worth adding, because `mergeSpeakers` is `internal` and the next caller need not hold that invariant — but it is not a live bug.
- **CodeRabbit, "the row context menu hides the transcript context menu."** Same reason: `SegmentRow` has no `.contextMenu` at head. The only `.contextMenu` in `RecordingDetailView.swift` is on the transcript `ScrollView` (line 368).
- **Cursor, "Merge confirm can drop the action" in `SpeakerNamePicker`.** Stale: `grep -c mergeTarget Mila/Views/SpeakerNamePicker.swift` is **0** at head; the alert went with the merge UI.
- **CodeRabbit's merge-risk line about "persistence failures leaving transcript and profile data out of sync."** There is something real here, but it is not new. `setSpeakerName` fires the callbacks — which `save()` the profile — before `persist()` (`RecordingStore.swift:526-536`), and `persist()`'s `Bool` return is discarded. `main` already does exactly that for `onSpeakerNamed`. The PR widens the pattern rather than introducing it, so I would not hold the PR on it.

## Nitpicks I'd actually take

- `MilaApp.swift:867` uses `match(embedding:)`'s default 0.55 and ignores `LiveAISettings.speakerSimilarityThreshold`, which the live path applies at `MilaApp.swift:1717` and which is user-visible in Settings (`SettingsView.swift:2086`). Given finding #1, a user who tightened that slider *because* of false matches still gets 0.55 here — and a false match now merges segments, not just mislabels them.
- `SpeakerDiarizer.swift:439` `sf.read(wav_path, dtype="float32")` loads the whole file. At 16 kHz mono float32 a three-hour recording is ~690 MB resident just to read a few seconds per speaker. `sf.info(...).samplerate` plus per-segment `start`/`stop` reads is a small change.
- `MilaApp.swift:826` force-unwraps `rec.segments.first(where: …)!.speaker!` inside the `guard` list. It is safe today only because the `contains` clause on the line above short-circuits first; reorder the clauses and the app traps. Hoisting it to a `let` before the `guard` costs nothing.

## Repo conventions

- **Tests.** No test file is touched — the diff against the merge base is seven files, all under `Mila/`, and `grep -rn 'subtractObservation\|mergeSpeakers\|splitSegmentSpeaker\|onSpeakerUnnamed' MilaTests` returns nothing. This is the most densely tested subsystem in the repo (`ObservedVoiceSnapshotsTests`, `CrossRecordingVoiceIsolationTests`, `LiveSpeakerNamingTests`, `RecognisedSpeakerAssignerTests`, `VoiceRecognitionGateTests`). Findings 3, 4, 5 and 6 are all unit-testable against `RecordingStore` + `SpeakerProfileStore` + `ObservedVoiceSnapshots` with no Python runtime involved. Please add them.
- The 8 → 20 retention change does **not** break `ObservedVoiceSnapshotsTests.test_retention_is_bounded_and_evicts_oldest_first` — it constructs `ObservedVoiceSnapshots(limit: 3)` explicitly. Checked.
- **`StoredRecording` drift: not applicable, verified.** `Recording.encode(to:)` is untouched; `splitSegmentSpeaker` and `mergeSpeakers` only write the existing `TranscriptSegment.speaker` field, so `StoredRecordingDriftTests` has nothing new to mirror in MilaKit.
- **`.claude/rules/feature-gates.md`: the gate holds.** `subtractObservation` carries the same `guard settings.isConfigured` as `updateProfile`, and both detached tasks re-check `voiceSettings.isConfigured` after their awaits. One consequence worth a line of documentation rather than a fix: with the feature off, un-naming no longer corrects a profile that is still on disk, and `clearOnOptOut` has already dropped the snapshot, so the correction is permanently unavailable for that recording.
- **`.claude/rules/python-subprocess.md`:** the drain rule is met — `embedSpeakers` reuses `runPython`, which reads both pipes in detached tasks before `waitUntilExit()` (`SpeakerDiarizer.swift:149-152`). The stderr-diagnostics half of the rule is not; see #8.

## Scope against #237

CodeRabbit's "Linked Issues" warning is fair, and this is a call for @urisland rather than something to go implement on the strength of a bot comment. #237 asks for three things. Profile correction on un-name/rename is here. The other two are not, as direct actions:

- *"Right-click a segment to reassign it to a different speaker in the recording"* — there is no reassign action. The nearest workflow is Split, then type the target speaker's name and rely on the auto-merge. If the target speaker is **unnamed**, that route does not exist at all: you must name them first, which itself writes a voice profile.
- *"Merge all segments of one into the other"* — reachable only as a side effect of typing a duplicate name, with no affordance and no confirmation. If neither speaker is named, you have to name one first.

Both existed as a "Reassign to…" / "Merge all into…" context menu in `ede627b` / `9697e08` and were removed after Cursor flagged an alert-presentation bug in the merge confirmation. Removing the feature rather than fixing the alert is what turns this into a scope question. Given finding #1 I think the right shape is the inverse of the current code: pull merge *out* of `setSpeakerName` into that explicit, confirmable action, and leave `setSpeakerName` as a pure label write.

---

urisland:
Closing this in favour of **#253**, which lands your work together with #236's — not a rejection, and the details matter here.

**Your commits are in #253, with you as the author.** `ca0a51c` and `eb962f3` were cherry-picked, not squashed into something of ours; `git log --format='%an'` on that branch shows you as the author of the first two commits, and the fixes are separate commits on top. This is your contribution and the history says so.

**Why the two PRs became one.** `eb962f3` is a descendant of `ca0a51c` — `git merge-base --is-ancestor ca0a51c eb962f3` passes — so despite the "Depends on #236" framing these were one stack, and this branch was a superset rather than a dependant. Every merge order had a problem: #236 alone leaves `main` unbuildable (it calls `onSpeakerUnnamed` and `subtractObservation`, both introduced here), while this PR alone builds but subsumes #236 and leaves #235 with nothing closing it. #253 closes both.

**What this got right.** `subtractObservation` is a correct algebraic inverse of `updateProfile`'s weighted fold — I read them side by side (`SpeakerProfileStore.swift:344-346` against `:288-289`) and they diverge only at the `maxSampleCount` clamp, which the existing doc comment establishes is unreachable. Deleting the profile when the remaining count hits zero, and deleting rather than persisting a non-finite centroid, are both the right calls: the alternative is a profile that silently poisons every future match. That function is the hard part of #237 and it is right. Almost everything I flagged was in the wiring around it, not in it.

Raising retention 8 → 20 was also correct reasoning — un-naming has to be able to find the snapshot — and I checked that it doesn't break `ObservedVoiceSnapshotsTests.test_retention_is_bounded_and_evicts_oldest_first`, which constructs its own `ObservedVoiceSnapshots(limit: 3)`.

**The one design change worth explaining.** I moved merge *out* of `setSpeakerName` rather than fixing it in place. Putting it there reads well from the popover — typing a name someone else already has plainly means "same person" — but `setSpeakerName` is also what both automatic namers loop over (`RecognisedSpeakerAssigner.finish` and the new post-pass auto-match). So two clusters that both cleared the similarity threshold against one stored profile would fold together with no user action, no confirmation and no undo, and which cluster survived depended on `Dictionary` iteration order. Two similar voices and one threshold is not an exotic input.

That also fixes the scope gap CodeRabbit's "Linked Issues" check flagged, and it was a fair flag: #237 asked for a direct per-segment reassign and a direct merge, and both had existed in `ede627b`/`9697e08` before being removed over the alert bug. #253 brings them back as "Move this line to…" and "Merge all these lines into…" in the speaker popover. The alert bug is avoided by shape rather than patched: the popover only *requests* the merge and `RecordingDetailView` — which the request does not dismiss — presents the confirmation, so there is no alert attached to a disappearing view. Removing the feature was the wrong trade, but the instinct that something was wrong with that alert was right.

I also took the merge-adds-to-the-target fix (CodeRabbit's finding, and it held: `mergeSpeakers` fired only `onSpeakerUnnamed`, so a Bob→Alice merge lost the observation entirely), the `speakerNames`-aware `used` set in `splitSegmentSpeaker`, the snapshot upsert, the re-read of the name after the embed, late URL resolution, stderr diagnostics in `embedSpeakers`, and the user's similarity threshold instead of the hardcoded 0.55. Full list in #253's description. The three bot findings I rejected as stale against this head — the `mergeSpeakers` unnamed-target case, the row context menu, and the `mergeTarget` alert — are recorded there too, with reasons.

Thanks for this. The design was sound and the algebra was right; what needed work was the wiring around it, which is the easiest part to get wrong and the hardest to see by reading.

---

Issue #246, opened 2026-09-01
Two flaky suites from #208 are still undiagnosed: MicrophoneConfigurationChangeTests and LLMSandboxDirectoryTests


urisland:
Successor to #208, which PR #245 closes. #245 fixed three of the five suites listed there with a diagnosed root cause and removed the CI blindness that made the rest undiagnosable. **These two were not diagnosed and are deliberately carried over here rather than being quietly closed with the parent.**

## What #245 actually fixed

One race, three suites: a test that adopts a *production* timeout longer than CI's `-default-test-execution-time-allowance 240` inverts which watchdog fires first. Under the allowance a stall is a normal, retryable assertion failure; over it, XCTest kills the test **host** — no message, nothing to retry, and the job prints a bare `Failing tests: …` line. That is the signature of the `aec65fc` main-red (`0 failures` reported next to a named failing test).

- `LLMRunnerTests` — the only test in the bundle spawning via `LLMRunner.run` with no `timeout:` (300 s default).
- `PostRecordingCoordinatorSendTests` — bounded neither wait: 600 s transcript wait, 300 s CLI.
- `ObsidianGitSyncerIntegrationTests` — its `runGit` helper had no deadline at all.

### 1. `MicrophoneConfigurationChangeTests`

Untouched by #245. Reported in #208 failing `test_a_storm_of_configuration_changes_is_bounded()` on a diff that touched only the MCP helper and log-privacy annotations.

Not a timeout-vs-allowance case — I checked: `bringUpOverride` means no real `AVAudioEngine`, so `stop()` cannot block on CoreAudio teardown, and every wait in the suite is already bounded (`waitUntil(3.0)`, `bringUpTimeout = 1.0`). Total runtime is 1-2 s.

The remaining hypothesis, **not** a diagnosis, is that the assertions are wall-clock shaped and a loaded runner can move them in either direction:

- `XCTAssertGreaterThan(posts, 10)` after a 0.6 s wall-clock burst at a 15 ms cadence — needs each iteration to stay under ~60 ms;
- `XCTAssertLessThanOrEqual(mic.restartCount, 2)` against a 1 s `minimumRebuildInterval`, where `Task.sleep` overshoot stretches the window the policy is judged over.

Deliberately not tuned: changing the numbers blind just relocates the threshold. This needs a real failure message first.

### 2. `LLMSandboxDirectoryTests`

[...]

PR #248, opened 2026-09-01
fix(speakers): pin the seed anchor weight and record why it is 3

### LINKED ISSUE #206: Tune the seed anchor weight for returning-speaker matching

### LINKED ISSUE #249: The seed anchor weight is an unexplained literal that no test observes
`LiveSpeakerDiarizer.seedPool` caps a returning speaker's stored `sampleCount`:

```swift
sampleCount: min(entry.sampleCount, 3),
```

That `3` is the seed anchor weight — the sole knob controlling how strongly a stored centroid anchors matching versus adapting to this session's acoustics. Two separate problems with how it sits in the code today, both independent of *what the value should be* (that is #206, and stays open).

## 1. The rationale is not in the code

The literal has no explanation at the site. `seedPool`'s doc comment says only that the count is "capped … so the centroid can adapt to current-session acoustics on the first few matches" — it does not say what the cap means arithmetically, why a cap is needed at all, or why the number is 3.

The reasoning does exist, in #206's thread: the confident-match fold in `assign` is a running mean, so over *k* matches it telescopes to `(n₀·c₀ + Σeⱼ)/(n₀ + k)`, which makes n₀ literally the stored voice's share of the centroid the next utterance is compared against. But a GitHub comment is not where someone reading `LiveSpeakerDiarizer.swift` will find it.

## 2. Nothing observes the value — a change to it is silently green

This is the part that can actually cause harm. `currentProfiles()` deliberately reports only the persisted pair (`observedCentroid` / `observedCount`), never the matching pair, so the seeded weight has **no direct read-out anywhere**. And every seeded fixture in `LiveSpeakerDiarizerPoolTests`, `RecognisedSpeakerMergeTests`, `CrossRecordingVoiceIsolationTests` and `RecognisedSpeakerAssignerTests` happens to use vectors far enough apart that the outcome does not depend on the weight.

Verified by replaying those fixtures through a float32 simulation of `seedPool` + `assign` + `cosineSimilarity` across n₀ ∈ {1, 2, 3, 4, 6, 8, 12, no-cap}: **every one passes unchanged at every value.** So today you could change 3 to 1, or delete the `min` entirely so a mature profile anchors at `VoiceProfile.maxSampleCount`, and the whole suite stays green — even though both are failures the arithmetic can prove:

* **n₀ = 0** — `(0·c₀ + e)/1 == e`, so the stored voice is discarded by its own first match and seeding affects only the first comparison of a recording.
* **no cap** — within-recording adaptation becomes arithmetically impossible, so a returning speaker whose acoustics changed never clears the threshold, never gets a confident match, `observedCount` stays 0, and `RecognisedSpeakerAssigner.finish` correctly refuses to auto-name them. Nothing is learned, so the profile can never adapt in a future recording either: the profile that most needs updating becomes the one that cannot be updated.

A regression that walks the value into either end is currently invisible.

## Intended fix

Name the constant and give it the derivation (what n₀ means, the resulting weight table, why both ends are wrong, and the asymmetry that keeps it small), then add a test file that pins the number, the arithmetic it implies, that the cap is a ceiling and not a floor, and the two behaviours the value is chosen for.

**No behaviour change** — the value stays 3.

## Relationship to #206

Deliberately narrow, and does **not** close #206. #206 asks whether 3 is the *right* value, and says up front that "unit tests can pin the arithmetic but cannot tell us whether recognition actually improves" — that needs multi-recording audio of the same speakers across capture setups, which nobody has. #206 also carries a second strand (a decayed or windowed mean for the persisted centroid) untouched by this.

This issue is only about the value being undocumented and unpinned, which is fixable today and worth fixing whatever the eventual number turns out to be — a future tuning PR then starts from a documented baseline with a test that makes the change visible.

urisland:
Closes #249
Refs #206

## What & why

`seedPool` capped a returning speaker's stored `sampleCount` at a bare `3`. That literal is the sole knob controlling how strongly a stored centroid anchors matching versus adapting to today's acoustics, and it was both unexplained at the code site and — the part that bites — **completely unobserved by any test**.

**Scope note, since this started as work on #206.** The sweep #206 asked for was carried out and written up in that issue's thread. Its conclusion: the value should stand. The arithmetic pins the ends but not the interior, and the failures either side are asymmetric in a way that argues for keeping the anchor small, which 3 already is. I re-derived that analysis against the current code and it checks out, including the weight table, which I reproduced by simulating `seedPool` + `assign` in float32.

But that verdict is *not* the same as answering #206, and **this PR does not close it** — see "Relationship to #206" below. What this PR closes is #249: the value being undocumented and unpinned, which is fixable today and worth fixing whatever the eventual number turns out to be.

### The rationale moves to the code

`seedAnchorWeight` replaces the literal and carries the derivation. The confident-match fold in `assign` is a running mean, so over *k* matches it telescopes exactly to `(n₀·c₀ + Σeⱼ)/(n₀ + k)`, which makes n₀ literally the stored voice's share of the centroid the next utterance is compared against:

| k  | n₀=1 | n₀=2 | **n₀=3** | n₀=6 | uncapped (40) |
|----|------|------|----------|------|---------------|
| 1  |  50% |  67% | **75%**  |  86% |  98% |
| 3  |  25% |  40% | **50%**  |  67% |  93% |
| 10 |   9% |  17% | **23%**  |  38% |  80% |

The doc comment records why `0` and "no cap" are both provably wrong, states the asymmetry that keeps the value small — a too-light anchor's damage is confined to one recording, a too-heavy anchor's persists across every future one — and is explicit that **none of it distinguishes 2 from 3 from 4**, pointing at #206 for the replay protocol that would.

### And it becomes observable

The value was unpinned in the strict sense: **nothing read it back.** `currentProfiles()` deliberately reports only the persisted pair, so the seeded weight has no direct read-out, and every seeded fixture in `LiveSpeakerDiarizerPoolTests`, `RecognisedSpeakerMergeTests`, `CrossRecordingVoiceIsolationTests` and `RecognisedSpeakerAssignerTests` uses vectors far enough apart that the outcome is identical for every weight from 1 to uncapped. I verified this by replaying those fixtures through a float32 simulation of `assign` across n₀ ∈ {1, 2, 3, 4, 6, 8, 12, no-cap}: **every one passes unchanged at every value.** Changing 3 to 1, or deleting the `min` so a mature profile anchors at `VoiceProfile.maxSampleCount` — which the constant now documents as provably wrong — left the whole suite green.

`SeedAnchorWeightTests` closes that. It uses a 2-D geometry embedded in 4-D vectors, so every similarity is the cosine of an angle difference and the margins are checkable by hand:

- **the constant is 3** — hardcoded, so it fails loudly and explains the bar for moving it;
- **the weight table** it implies, derived from the constant, so the doc comment cannot go stale;
- **the cap is a ceiling, not a floor** — a one-sample profile anchors at 1, not 3, so a barely-known voice does not steer matching as hard as a well-known one; plus the control that stored counts 3, 5 and 40 all anchor identically;
- **ten of this session's utterances must outweigh the anchor** — otherwise a returning speaker whose mic/room/health changed never clears the threshold, never gets a confident match, `observedCount` stays 0 and nothing is learned, so the profile that most needs updating is the one that cannot be updated. At the current weight the probe lands at 0.750; at a weight of 6 it would be 0.685 (borderline attach) and uncapped 0.456 (mints a duplicate speaker for the same person);
- **one utterance must not take the centroid halfway** — the mirror failure, where a single noisy or false-positive match half-steals the entry. Probe at 0.744 now, 0.623 at a weight of 1.

Each behavioural assertion discriminates by `observedCount`, not by the returned speaker id: a borderline attach returns the *same* id as a confident match, and only the fold is visible.

**No behaviour changes.** `min(entry.sampleCount, Self.seedAnchorWeight)` is the same 3 it was. Deliberately: per #206, changing it to another unmeasured number would only move the guess.

## Relationship to #206

**#206 stays open.** It asks whether 3 is the *right* value, and its own body sets the bar this PR cannot clear:

> Both need real multi-recording audio with the same speakers across sessions to evaluate — unit tests can pin the arithmetic but cannot tell us whether recognition actually improves.

That names precisely what this PR does and rules it insufficient, correctly. Two things remain undone there, and closing #206 would bury both:

1. **Choosing between 2, 3 and 4.** Nothing here distinguishes them; that needs a corpus nobody has.
2. **The decayed / windowed mean for the persisted centroid** — #206's second strand, which this PR does not touch at all.

So this is linked to #249 instead, which is the narrow, genuinely-finished piece: the weight is now documented and pinned, so a future tuning PR starts from a stated baseline with a test that makes any change visible. A comment on #206 records what landed and what is still open.

(CodeRabbit's advisory Linked Issues pre-merge check flagged the original `Closes #206` for exactly this reason. It was right.)

## How it was tested

No Swift toolchain in this environment, so **not built or run locally** — CI is the compiler here. What I did instead:

- Reimplemented `seedPool`, `assign` (both fold branches, `createThreshold`, the short-utterance and dimension guards) and `cosineSimilarity` in float32 and swept n₀ ∈ {1, 2, 3, 4, 6, 8, 12, no-cap}. This produced every expected value asserted in the new tests, and confirmed the margins are comfortable (the tightest is +0.044 on a 0.7 threshold, against float32 error of ~1e-7).
- Replayed the existing seeded fixtures through the same simulation to establish the "nothing catches a changed cap" claim above.
- Confirmed the new tests fail at the values they are meant to exclude, rather than merely passing at 3.

Scope note: the diff is confined to `LiveSpeakerDiarizer.swift` and one new test file, which neither #236 nor #238 touches.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — no Swift toolchain available; relying on CI
- [x] User-facing change is noted in the README changelog (if applicable) — n/a, no behaviour change

https://claude.ai/code/session_01PewyUzivK9LuGMezKJM2vN

---

> [!NOTE]
> **Low Risk**
> No production logic change beyond naming the existing cap; risk is limited to test maintenance if the weight is tuned later.
> 
> **Overview**
> Documents and test-pins the **seed anchor weight** (`n₀`) that caps how strongly a returning speaker’s stored centroid influences live matching in `seedPool` — **runtime behavior is unchanged** (still `3`).
> 
> `LiveSpeakerDiarizer` gains `static let seedAnchorWeight = 3` in place of the bare literal, with a long doc comment on the telescoped running-mean math, why `0` and uncapped anchors fail, and why the choice between 2/3/4 remains empirical (#206). `seedPool` now uses `min(stored, seedAnchorWeight)` and clarifies that the cap is a **ceiling, not a floor**.
> 
> **`SeedAnchorWeightTests`** adds geometry-based unit tests so changing or removing the cap can’t slip through silently: the constant stays 3, the implied share table matches the doc, young profiles anchor at true count, mature profiles bind at the cap, ten session folds must still allow adaptation (and avoid duplicate speakers), one fold must not half-steal the entry, and seeded weight must not inflate `observedCount` for persistence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd22ee59c34a660372d1b420da94fe641fec286d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


urisland:
bugbot run

---

urisland:
## `unit-and-ui-tests` is red on `e8be45d`, and none of it is this PR's

Recording this rather than re-running, because the failure is informative.

```
LLMRunnerTests.test_runner_spawns_child_in_isolated_empty_directory
    failed: caught error: "timedOut(seconds: 30)"          — took 38.095s
LLMSandboxDirectoryTests.test_runner_leaves_the_shared_directory_in_place
    failed: caught error: "timedOut(seconds: 30)"          — took 36.104s
LiveAIDeletedLineSessionTests.test_deleting_during_an_in_flight_tick_recovers_on_a_new_session
    XCTAssertTrue failed - Live AI wedged: the dropped tick left `inFlight` set   (×2 iterations)
```

**Why it is not this PR's.** The diff is two files: `LiveSpeakerDiarizer.swift`, where a literal `3` becomes `seedAnchorWeight = 3` at the same use site, and a new `SeedAnchorWeightTests.swift`. `min(entry.sampleCount, Self.seedAnchorWeight)` is the same expression it was, so the change is behaviour-neutral by construction. Nothing in it can reach `LLMRunner`'s subprocess path or `LiveAISession`'s tick loop. The new test file adds cases; it does not touch existing ones.

**What it actually is.** The first two are the #246 signature exactly: a bounded subprocess wait expiring, each overshooting its own 30s bound by 6-8s — the same shape as the three failures on #245's head (`timedOut(seconds: 120)` at 126.8s, `timedOut(seconds: 30)` at 36.3s, and the git one that carried `Switched to a new branch 'main'` on stderr and still failed to exit within 60s). A command that has already finished its work while the wait does not return.

The third is more interesting and I think the same cause. That test asserts a tick cannot leave `inFlight` set — the regression fixed in #242 by the `!Task.isCancelled` guard. That fix is in this branch's base (`e0eb6b0`, well after `aec65fc`). What the assertion cannot distinguish is a tick that wedged from a tick that was never *scheduled*: under a starved cooperative pool the observable state is identical — `inFlight` still set, later ticks coalescing behind it. So a timing-sensitive async test is a natural third victim of the same starvation, not a second bug.

If that reading is right, this red is a data point **for** #251's thesis rather than against this PR.

**Plan, and what I am deliberately not doing.** [#251](https://github.com/island-io/mila/pull/251) takes every blocking subprocess wait off the shared pool and is the fix for all three. Porting it into a two-file constant PR would make this a duplicate of it, so instead: let #251 land, merge `main` into this branch, and let CI re-run against the fixed base.

I have **not** spent the one permitted re-run. A green re-run here would only confirm what the diff already establishes, and would leave the same three tests just as likely to fail on the next PR. If #251 stalls, I will re-run this once and say so.

---

PR #251, opened 2026-09-01
fix(subprocess): take every blocking subprocess wait off the shared thread pool

### LINKED ISSUE #246: Two flaky suites from #208 are still undiagnosed: MicrophoneConfigurationChangeTests and LLMSandboxDirectoryTests

### LINKED ISSUE #250: Blocking subprocess waits are parked on the non-overcommit DispatchQueue.global() pool
Narrower successor to #246, carved out because it is diagnosed and #246's `MicrophoneConfigurationChangeTests` thread is not. #246 stays open for that one.

## The bug

Three places run a child process and wait for it, all with the same shape:

- `LLMRunner.executeProcess`
- `ProcessGitCommandRunner.runSync`
- the `runGit` helper in `ObsidianGitSyncerIntegrationTests`

Each parks **three blocking work items on `DispatchQueue.global()` per invocation** — two pipe readers sitting in `FileHandle.availableData` until EOF, and one thread sitting in `Process.waitUntilExit()` — and the caller then blocks a fourth pooled thread waiting on the resulting `DispatchGroup`. `LLMRunner.run` and `LLMRunner.diagnose` add that fourth one explicitly (`DispatchQueue.global(qos: .userInitiated).async { executeProcess(...) }`), and `ProcessGitCommandRunner.run` does the same on `.utility`.

The global concurrent queues are **non-overcommit**. libdispatch does not spin up a thread for a queued item just because the existing ones are blocked in the kernel; it notices the stall on a delay and grows the pool slowly, up to a hard cap on constrained threads. So each blocking item parked there makes the next one slower to be **scheduled at all**, and that delay is unbounded in the bad case.

This is not a new observation in the codebase — several comments already record "10–15s of dispatch latency on this CI image" as ordinary, and `ObsidianGitSyncerIntegrationTests.pipeDrainTimeout` was widened from 5s to 30s specifically to absorb it. That latency *is* the bug; it was being budgeted for rather than removed.

## The evidence

The #245 instrumentation caught it. `unit-and-ui-tests` on `f0b663f` went green with 3 failed executions:

```
LLMSandboxDirectoryTests.test_a_oneshot_run_does_not_sweep_a_concurrent_tick
  timedOut(seconds: 120)   — took 126.8s
LLMSandboxDirectoryTests.test_session_runs_share_the_same_directory_as_oneshot_runs
  timedOut(seconds: 30)    — took 36.3s
ObsidianGitSyncerIntegrationTests.test_sync_rebases_nonconflicting_remote_change_and_pushes
  "git checkout -b main did not exit within 60s and was killed
   (stderr so far: Switched to a new branch 'main'"   — took 65.9s
```

`Switched to a new branch 'main'` is **git's own success message**. On that invocation git ran, completed its work, and wrote to a pipe we were reading — and 65s later the item meant to observe its exit still had not run. Not a slow command, not a drain bug, not leftover repo state.

The overshoot arithmetic confirms it. Each failure exceeds its own bound by almost exactly the SIGTERM + SIGKILL grace of the code that reported it (`executeProcess`: 1 + 5 = 6s, observed 6.8s and 6.3s; `runGit`: 2 + 3 = 5s, observed 5.9s). In every case the exit observer never reported in **even after the child was killed** — which is what "the work item was never scheduled" looks like, and is not what "the command was slow" looks like.

## Fix

1. `BlockingWork.onDedicatedThread(named:)` — a named `Thread` per blocking wait. A dedicated thread cannot be starved and cannot starve anyone else, and if it leaks (a grandchild holding a pipe open forever) it costs one thread rather than a slot the whole process shares. No `DispatchQueue.global()` call sites should remain.
2. Observe the exit with `Process.terminationHandler` + a `DispatchSemaphore`, installed before `run()` — that costs no thread at all. Keep `waitUntilExit()` on a dedicated thread as a backstop, because `executeProcess` documents a macOS 26 reaping race where it never returns even after `SIGKILL`; the two paths fail independently.

## Also in scope

Two things sitting on the same code, both from #246 and both cheap once the above is in:

- **`LLMSandboxDirectoryTests` budgets.** `test_a_oneshot_run_does_not_sweep_a_concurrent_tick` gives its tick 120s but cannot release it until 60 + 60s of preceding waits have elapsed — zero margin **by construction**, which is why it was observed at 126.8s. Raising the tick's bound is the wrong fix: 120s is already half the 240s `-default-test-execution-time-allowance`, and crossing that swaps a retryable failure for the unretryable test-host kill #245 existed to eliminate. The sub-waits have to shrink instead. Separately, `test_session_runs_share_the_same_directory_as_oneshot_runs` spawns six children where four would assert strictly more.
- **`setenv` on the test host.** `ObsidianGitSyncerIntegrationTests` mutates the whole test-host process's `environ` (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `GIT_CONFIG_NOSYSTEM`) because `ProcessGitCommandRunner` inherits the host environment verbatim and offered no other way in. The host is a live SwiftUI app and `setenv` can `realloc` `environ` underneath a concurrent `getenv`/`posix_spawn`. The runner needs a per-invocation child `environment`.

## Not in scope

`MicrophoneConfigurationChangeTests` — still no failure message, still no mechanism. Stays on #246.

## How to judge it

The **`Failed test-case executions`** count in the `Summarise test failures` step, across several runs. A single green tick proves nothing against a ~20% historical flake rate.

### LINKED ISSUE #255: Flaky: test_retranscribe_after_audio_compressed_reads_current_file_not_stale_wav treats compressRecordingAudio as a barrier, and it isn't one
Caught as a recovered flake by the #245 failed-execution counter, on the `unit-and-ui-tests` run for #251 (job 99899172181, head `b2df3cb`). One failed execution, absorbed by retry; the job went green.

```
MilaTests/TranscriptionServiceTests.swift:855: XCTAssertTrue failed
    - Compression should have swapped the audio to .m4a
MilaTests/TranscriptionServiceTests.swift:857: XCTAssertFalse failed
    - The original .wav (stale snapshot's path) must be gone
MilaTests/TranscriptionServiceTests.swift:865: XCTAssertTrue failed
    - prepareForRetranscription must keep the compressed .m4a name
Test Case 'test_retranscribe_after_audio_compressed_reads_current_file_not_stale_wav'
    failed (0.124 seconds).
```

**0.124s, three wrong values, no timeout** — so this is not the #246 wedged-wait family (those overshoot a 30–120s bound). It is a real race, and the mechanism is visible in the source.

## Mechanism

The test's synchronisation step is this, with a comment that states the property it needs:

```swift
// (The first pass also auto-kicks a compression; awaiting here is idempotent
// and guarantees the on-disk file is the .m4a regardless of which won.)
await store.compressRecordingAudio(id: fixture.recording.id)
```

**That comment is wrong.** `RecordingStore.compressRecordingAudio` is not "wait for compression", it is "start one *unless one is already running*":

```swift
guard !compressingIDs.contains(id) else { return }   // RecordingStore.swift:322
compressingIDs.insert(id)
```

and the first pass kicks an unawaited one on completion:

```swift
Task { await store.compressRecordingAudio(id: compressID) }   // TranscriptionService.swift:1040
```

`waitForIdle()` does not cover that detached Task. So the test's `await` is a real barrier in only two of the three reachable states:

| state when the test calls it | result |
|---|---|
| auto-kick already finished | src is no longer `.wav`, returns; state is `.m4a` ✅ |
| auto-kick not started yet | the test's call does the work ✅ |
| **auto-kick mid-flight** | **the dedup guard returns immediately — the test awaits nothing** ❌ |

In the third state the test asserts against pre-compression state and fails. The 0.124s runtime corroborates it: the test never paid for an AAC encode.

The irony is that this test was itself written to make a compression race deterministic ("REGRESSION (flaky CI)" in its doc comment) — it just picked a call that dedups rather than one that joins.

## Not caused by #251

Recorded because it surfaced on that PR and the question was asked directly:

- `MilaTests/TranscriptionServiceTests.swift`, `Mila/Models/RecordingStore.swift`, `Mila/Audio/AudioCompressor.swift` and `Mila/Transcription/TranscriptionService.swift` are all **byte-for-byte identical to `main`** on that branch (`git diff main --stat` over the four is empty). The test was last touched in #189.
- `AudioCompressor` imports only Foundation / AVFoundation / OSLog — no `Process`, no `DispatchQueue`, nothing that routes through the `BlockingWork` change.

What #251 plausibly did is *perturb the timing*: freeing blocked threads off the shared pool changes when everything else gets scheduled, which is enough to move a latent race into the reachable window. That is a trigger, not a cause — the defect is in the test and is reachable on `main` today.

## Suggested fix

Do not widen a timeout and do not retry. Make the test wait for the observable post-condition instead of assuming the call was a barrier: kick compression, then poll (bounded) until the store's `audioFileName` ends in `.m4a` and the `.wav` is gone, and fail with a message naming the state if it doesn't. That is correct in all three states above.

The alternative — making `compressRecordingAudio` *join* an in-flight compression for the same id instead of returning — changes production semantics for callers that deliberately want "don't start a second one" (`QuickActionsController`, the post-completion hook), so it should not be done just to serve a test.

While there: the misleading comment should go with the fix, since it is what makes the current shape look correct.

urisland:
Closes #250
Closes #255

Part of #246 — threads 1, 3 and 4. **Deliberately does not close #246**: its `MicrophoneConfigurationChangeTests` thread is untouched and stays open there (see "What I left alone" below). #250 is the narrower issue this actually fixes; #255 is a second scheduling-shaped test flake this PR surfaced and fixes.

### The diagnosis

`LLMRunner.executeProcess`, `ProcessGitCommandRunner.runSync` and the `runGit` helper in `ObsidianGitSyncerIntegrationTests` each parked **three blocking work items on `DispatchQueue.global()` per invocation** — two pipe readers in `FileHandle.availableData` until EOF, one thread in `Process.waitUntilExit()` — and the caller then blocked a fourth pooled thread on the resulting `DispatchGroup`. `LLMRunner.run`/`.diagnose` and `ProcessGitCommandRunner.run` supplied that fourth one explicitly.

The global concurrent queues are **non-overcommit**: libdispatch will not spin up a thread for a queued item just because the existing ones are blocked in the kernel. It notices the stall on a delay and grows the pool slowly, up to a hard cap. So every blocking item parked there makes the next one slower to be **scheduled at all**, and that delay is unbounded in the bad case.

That is not a new observation here — it is the "10-15s of dispatch latency on this CI image" several comments in this repo already treat as ordinary, and it is why `pipeDrainTimeout` had to be widened from 5s to 30s in #245. The latency *was* the bug; it was being budgeted for rather than removed.

### The evidence it was the wait, not the command

The #245 instrumentation earned its keep. From `unit-and-ui-tests` on `f0b663f`:

```
git checkout -b main did not exit within 60s and was killed
  (stderr so far: Switched to a new branch 'main'
```

`Switched to a new branch 'main'` is **git's own success message**. git ran, completed its work, and wrote to a pipe we were reading — and 65s later the item meant to observe its exit still had not run.

The overshoot arithmetic seals it. Each of the three failures in that run exceeded its own bound by almost exactly the SIGTERM + SIGKILL grace of the code reporting it — `executeProcess` 1 + 5 = 6s (observed 6.8s, 6.3s), `runGit` 2 + 3 = 5s (observed 5.9s). In every case the exit observer never reported in **even after the child was killed**. That is a work item that was never scheduled, not a command that was slow.

### The fix

- New `Mila/Actions/BlockingWork.swift`: `onDedicatedThread(named:)`, a named `Thread` per blocking wait. Not starvable, and not a starver; a reader that leaks (a grandchild holding a pipe open forever) costs one thread instead of a slot the whole process shares. **No `DispatchQueue.global()` call sites remain in the repo** — `grep` finds only comments.
- The exit is observed by `Process.terminationHandler` + a `DispatchSemaphore`, installed before `run()`, which costs no thread at all. `waitUntilExit()` stays on a dedicated thread as a **backstop**, because `executeProcess` documents a macOS 26 reaping race where it never returns even after `SIGKILL` — the two paths fail independently, whichever notices first signals, and every wait is bounded so a double signal is harmless.
- `ProcessGitCommandRunner` gains a per-invocation child `environment`, which is what lets the Obsidian suite stop calling `setenv` on the test **host**. The host is a live SwiftUI app; `setenv` can `realloc` `environ` underneath a concurrent `getenv`/`posix_spawn`, and no amount of care inside one test class contains that.
- `.claude/rules/python-subprocess.md` gains the rule, next to the existing pipe-drain rule that BUGBOT already reads.

### What bounds the thread count

An invocation costs **four** threads for the child's lifetime (2 pipe readers, 1 `waitUntilExit()` backstop, 1 blocked caller), so the ceiling is 4 × concurrent **child processes** — and every spawning path is already bounded upstream, by limits that exist to bound *processes*, which is the scarcer resource:

| caller | bound | children |
|---|---|---|
| `RecordingSummarizer` backfill / regeneration | `maxConcurrent = 2` + `backfillQueue`; exists "so a 20-recording catch-up sweep doesn't fork 20 `claude -p` subprocesses" | 2 |
| `LiveAISession` ticks | one `inFlight` Task + coalescing | 1 |
| `PostRecordingCoordinator` | one in-flight handle | 1 |
| `RenameRecordingSheet` "Suggest" | one modal sheet, guarded by `isFetchingName` | 1 |
| `LLMSettings.diagnose` | one Settings panel, user-driven | 1 |
| `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`; `sync` issues commands sequentially | 1 |

Worst case — batch re-summarise during a live meeting with an Obsidian sync — is ~7 children, ~28 transient threads, against a per-process limit in the hundreds. The only path that could fan out is the one that is *explicitly* capped, at 2.

This is the **same** concurrency as before: the old code created the same work items, they just contended for a capped pool, which is what starved. The change moves where concurrency lands; it does not raise it. A cap inside `BlockingWork` would be a regression rather than a safeguard — limiting thread creation there re-queues blocked work behind a bound, which is the original bug in a smaller box. That is now an explicit "do not add a cap here" in the type's docs and in the rules file, pointing new batch callers at `RecordingSummarizer.maxConcurrent` as the pattern. The whole accounting lives in `BlockingWork`'s doc comment so it is reviewable in-tree.

### The two `LLMSandboxDirectoryTests` budgets

Local arithmetic, independent of load, and worth fixing on their own merits:

- **`test_a_oneshot_run_does_not_sweep_a_concurrent_tick`** gave its tick `timeout: 120`, but the barrier that lets the tick return is written only after `waitForFile(ready, timeout: 60)` and a `timeout: 60` one-shot — **60 + 60 = exactly the tick's own budget**, i.e. zero margin by construction. Observed failing with `timedOut(seconds: 120)` at 126.8s. The sub-waits shrink to 20 + 30; the tick's bound is **not** raised, because 120s is already half the 240s `-default-test-execution-time-allowance` and crossing that swaps a retryable failure for the unretryable host kill #245 existed to remove. A new `test_tick_budget_exceeds_the_waits_it_spans` asserts `tickTimeout > 2 × (tickReadyTimeout + spawnTimeout)` and `tickTimeout < 240`, so the next person to touch a number finds out in milliseconds rather than in two minutes.
- **`test_session_runs_share_the_same_directory_as_oneshot_runs`** spawned six children; the last two only re-ran the session/resume pair to look at argv. One echo script for all four runs cuts it to four spawns (~156s worst case rather than ~234s against a 240s allowance) while asserting **strictly more**. The fixture now emits `ARGC:$#` alongside the joined `$*`, and the test asserts `argc == oneShot.argc + 2` for both session modes *before* it looks at the uuid — a check that depends on no string matching and would catch a production path that stopped emitting `--session-id`/`--resume` (every Live AI tick would then silently start a fresh conversation instead of continuing the meeting). Plus two negative cases the old shape never had.

### Obsidian isolation

- Temp roots are named after the test that created them, and `setUp` — not `tearDown` — asserts every earlier root is gone, so a leak is attributed to the test that caused it. `tearDown` reports a *failed* removal rather than swallowing it with `try?`, which is the only way the last test in the class can report its own leak.
- `setenv`/`unsetenv` are gone; git-config isolation is per-invocation child environment for both `runGit` and the production runner.

### #255 — a second scheduling-shaped flake this PR surfaced

`RecordingStore.compressRecordingAudio(id:)` takes its `compressingIDs` guard before the first suspension: it is "start one unless one is running", not "wait for one". The completion path kicks an unawaited `Task { await store.compressRecordingAudio(id:) }` that `waitForIdle()` does not cover, so two tests using it as a barrier got one only when the auto-kicked transcode had already finished — or had not started. In the middle state they awaited nothing and asserted against still-`.wav` state (three wrong values in 0.124s: never paid for an encode).

`TranscriptionServiceTests` already had the right helper, `drainPostCompletionCompression`, whose doc comment spells the mechanism out; these two older tests were never migrated to it. They are now, so the only bare `compressRecordingAudio` call left in the bundle is inside the helper. No assertion is weakened. Not a production regression — the whole compression path is byte-for-byte identical to `main` on this branch and `AudioCompressor` is AVFoundation-only; this PR's scheduling change is what moved a latent race into the reachable window.

## What I left alone, and why

**`MicrophoneConfigurationChangeTests`.** It has still never produced a failure message, including in the green runs on this branch. #246 is explicit that its wall-clock-shaped assertions (`posts > 10` after a 0.6s burst at 15ms cadence; `restartCount <= 2` against a 1s `minimumRebuildInterval`) must not be tuned blind, and I agree: the burst window is genuinely load-bearing for the `restartCount` assertion, so "just use a fixed iteration count" would trade one wall-clock dependency for a worse one. It also spawns no subprocesses, so nothing here should move it either way — which makes its next occurrence a clean signal rather than a confounded one. It stays on #246.

**The 30s drain grace that can return an empty string from a successful run.** Still there in `executeProcess`. This PR removes the thing that made it *fire* (readers are now scheduled immediately) and hardens the caller that was blind to it, but the underlying return-success-with-no-output behaviour is a separate design question I did not want to change unvalidated.

## How it was tested

**Not tested locally — there is no Swift toolchain in this environment, so CI is the only compiler.** No claim is made beyond what CI reports.

On `b2df3cb`: `build-and-test` ✅, `unit-and-ui-tests` ✅, Cursor Bugbot ✅ (zero findings), both e2e jobs ✅, UI layout ✅. The `Summarise test failures` step reported **one** failed execution in the whole bundle — the compression test, now fixed in `35991bc` — and **none** of the #246 family: no `LLMSandboxDirectoryTests`, no `ObsidianGitSyncerIntegrationTests.test_sync_rebases…`, no `LLMRunnerTests`, no `LiveAIDeletedLineSessionTests`, and no timeout anywhere in the run.

One clean run is consistent with the fix working, not proof of it, against a ~20% historical rate. What raises it above luck is that the four tests failed as a group under load and the run is clean of all four at once, and the diagnosis predicted exactly which would stop and why. The **`Failed test-case executions`** count remains the metric. No test is skipped, quarantined or retried to get there.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — cannot be run here; CI is the compiler, and `build-and-test` is green
- [x] User-facing change is noted in the README changelog (if applicable) — n/a, no user-facing change

---

> [!NOTE]
> **Medium Risk**
> Touches core subprocess lifecycle (LLM CLI, git sync) and thread usage; behavior is intended to be equivalent but mis-tuned timeouts or semaphore handling could affect cancellation or output capture under edge cases.
> 
> **Overview**
> Fixes **issue #246**-style CI hangs where fast commands (e.g. `git checkout`) hit timeouts even though stderr already showed success — caused by parking pipe drains and `waitUntilExit()` on **non-overcommit** `DispatchQueue.global()`, so exit observers could sit unscheduled for tens of seconds.
> 
> **Production:** Adds `BlockingWork.onDedicatedThread(named:)` and moves all blocking subprocess work in `LLMRunner` and `ProcessGitCommandRunner` onto named dedicated threads. Child exit is signaled via **`terminationHandler` + bounded `DispatchSemaphore` wait** (installed before `run()`), with `waitUntilExit()` kept on a dedicated thread as a macOS 26 backstop. `ProcessGitCommandRunner` gains **per-invocation child `environment`** so tests can isolate git config without `setenv` on the SwiftUI test host. Documents the rule in `.claude/rules/python-subprocess.md`.
> 
> **Tests:** Hardens flakes exposed under load — Obsidian integration uses child env + same threading model; sandbox tests tighten timeout budgets and merge argv/cwd checks into four spawns; Live AI waits use `mustHappen` vs `quietWindow`; microphone storm test lowers the notification-count precondition; transcription tests **`drainPostCompletionCompression`** instead of a no-op `compressRecordingAudio` when auto-compression is in flight (**#255**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff8d2e3ee9300445351405b1e3fb8d1dd2de2de8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability and responsiveness when running AI and Git operations, especially under heavy or constrained system load.
  - Added safer timeout and process-cleanup handling to prevent operations from hanging indefinitely.
  - Improved Git synchronization cleanup and output handling, including more reliable completion detection.

- **Tests**
  - Expanded coverage for session isolation, timing behavior, temporary-directory cleanup, audio compression completion, and configuration-change bursts.

- **Documentation**
  - Clarified guidance for safely running blocking subprocess work and controlling concurrent operations.

### DIFF
--- a/.claude/rules/python-subprocess.md
+++ b/.claude/rules/python-subprocess.md
@@ -39,6 +39,50 @@ let stdoutData = stdout.fileHandleForReading.readDataToEndOfFile()  // too late
 let stderrData = stderr.fileHandleForReading.readDataToEndOfFile()
 ```
 
+## Never Park a Blocking Subprocess Wait on `DispatchQueue.global()`
+
+The global concurrent queues are **non-overcommit**: libdispatch will not spin
+up a thread for a queued work item just because the existing ones are blocked in
+the kernel. It notices the stall on a delay and grows the pool slowly, up to a
+hard cap. So every blocking item you park there makes the *next* one slower to
+be **scheduled at all** — and that delay is unbounded in the bad case.
+
+A subprocess wait is three blocking items if you write it the obvious way: two
+pipe readers parked in `availableData` until EOF, and one thread parked in
+`waitUntilExit()`. The caller then blocks a fourth waiting on the group. Under
+load this produced ordinary-looking timeouts on commands that had already
+succeeded (issue #246):
+
+```text
+git checkout -b main did not exit within 60s and was killed
+  (stderr so far: Switched to a new branch 'main'
+```
+
+`Switched to a new branch 'main'` is git's own success message. git ran,
+succeeded, and wrote to a pipe we were reading — and 65s later the work item
+meant to notice its exit still had not run. Not a slow command; a wait that
+never started.
+
+**Rules:**
+
+1. Observe the exit with `process.terminationHandler` plus a
+   `DispatchSemaphore`, installed **before** `process.run()`. That costs no
+   thread of ours at all.
+2. Anything that must genuinely block — a pipe reader, a `waitUntilExit()`
+   backstop, a synchronous `runSync` body — goes on a thread of its own via
+   `BlockingWork.onDedicatedThread(named:)` (`Mila/Actions/BlockingWork.swift`).
+   A dedicated thread cannot be starved, cannot starve anyone else, and if it
+   leaks (a grandchild holding a pipe open forever) it costs one thread rather
+   than a slot in a pool the whole process shares.
+3. Keep the `waitUntilExit()` backstop even with a `terminationHandler`: macOS
+   26 has a reaping race where `waitUntilExit` never returns even after
+   `SIGKILL`. The two paths fail independently; whichever notices first signals.
+   Only wait on the semaphore once, so a double signal is harmless.
+
+This applies to tests as much as to production code — the test host is one
+process, and a suite that parks blocking work on the shared pool degrades every
+other suite in the bundle.
+
 ## Known Compatibility Patches
 These monkey-patches are required as of pyannote.audio 3.x + PyTorch >= 2.6 + speechbrain:
 
--- a/.claude/rules/python-subprocess.md
+++ b/.claude/rules/python-subprocess.md
@@ -77,7 +77,21 @@ never started.
 3. Keep the `waitUntilExit()` backstop even with a `terminationHandler`: macOS
    26 has a reaping race where `waitUntilExit` never returns even after
    `SIGKILL`. The two paths fail independently; whichever notices first signals.
-   Only wait on the semaphore once, so a double signal is harmless.
+   Both may signal, and that is harmless: a `DispatchSemaphore` is only unsafe
+   to destroy while its value is *below* where it started, and every wait on it
+   is bounded (`wait(timeout:)`), including the further waits the
+   SIGTERM-then-SIGKILL escalation performs. Never add an unbounded
+   `semaphore.wait()` — that is the hang this whole section exists to remove.
+
+**Bounding the threads.** A dedicated thread per blocking wait means the thread
+count follows the number of concurrent CHILD PROCESSES, so that is what has to
+be bounded — and it already is, upstream, where it also bounds the thing users
+actually care about (`RecordingSummarizer.maxConcurrent = 2` exists so a
+20-recording catch-up sweep does not fork 20 `claude -p` processes). Do not
+"fix" this by capping thread creation inside `BlockingWork`: a cap there just
+re-queues blocked work behind a limit, which is the original bug. If you add a
+new caller that can spawn subprocesses in a batch, give the CALLER a concurrency
+limit.
 
 This applies to tests as much as to production code — the test host is one
 process, and a suite that parks blocking work on the shared pool degrades every
--- a/Mila/Actions/BlockingWork.swift
+++ b/Mila/Actions/BlockingWork.swift
@@ -45,6 +45,42 @@ import Foundation
 /// work — private serial queues target an *overcommit* root queue, so they
 /// always get a thread — but that is a subtle property to rely on, and it is
 /// invisible at the call site. `Thread` says what it means.
+///
+/// ## What bounds the thread count
+///
+/// The obvious objection to "a thread per blocking wait" is that it trades a
+/// bounded pool that starves for unbounded native threads that do not. It is
+/// worth being precise about why that is not what happens here.
+///
+/// A subprocess invocation costs **four** threads for its lifetime: two pipe
+/// readers, one `waitUntilExit()` backstop, and the caller blocked inside
+/// `executeProcess` / `runSync`. So the ceiling is four times the number of
+/// concurrent CHILD PROCESSES — and every path that spawns one is already
+/// bounded, upstream, by something that exists to bound *processes*, which is
+/// the scarcer resource:
+///
+/// | caller | what bounds it | concurrent children |
+/// |---|---|---|
+/// | `RecordingSummarizer` backfill / regeneration | `maxConcurrent = 2`, with `backfillQueue` holding the rest — added so a 20-recording catch-up sweep does not fork 20 `claude -p` processes | 2 |
+/// | `LiveAISession` ticks | one `inFlight` Task, with coalescing | 1 |
+/// | `PostRecordingCoordinator` | one in-flight handle | 1 |
+/// | `RenameRecordingSheet` "Suggest" | one modal sheet, guarded by `isFetchingName` | 1 |
+/// | `LLMSettings.diagnose` (Settings → Test) | one Settings panel, user-driven | 1 |
+/// | `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`, and `sync` issues its commands sequentially | 1 |
+///
+/// Even with every one of those in flight at once — a batch summary sweep
+/// during a live meeting while an Obsidian sync runs — that is ~7 children and
+/// ~28 threads, transient, against a per-process limit in the hundreds.
+///
+/// Note this is the SAME bound as before. The old code created exactly the same
+/// number of work items; the only difference is that they used to contend for a
+/// capped shared pool, which is what starved. The change does not raise
+/// concurrency, it moves where the concurrency lands.
+///
+/// **Do not add a cap here.** A limit on thread creation inside this type would
+/// re-queue blocked work behind a bound — which is precisely the failure this
+/// exists to remove. A new caller that can spawn subprocesses in a batch needs
+/// a concurrency limit of its own, like `RecordingSummarizer.maxConcurrent`.
 enum BlockingWork {
 
     /// Start `body` on a thread of its own, right now.
--- a/Mila/Actions/BlockingWork.swift
+++ b/Mila/Actions/BlockingWork.swift
@@ -1,0 +1,62 @@
+import Foundation
+
+/// Somewhere to park a **blocking** wait that is not
+/// `DispatchQueue.global()`.
+///
+/// ## Why this exists (issue #246)
+///
+/// Every subprocess we run — `LLMRunner.executeProcess`,
+/// `ProcessGitCommandRunner.runSync`, and the `runGit` helper in
+/// `ObsidianGitSyncerIntegrationTests` — used to hand three *blocking* work
+/// items to `DispatchQueue.global()` per invocation: two pipe readers parked
+/// in `FileHandle.availableData` until EOF, and one thread parked in
+/// `Process.waitUntilExit()`. The caller then blocked a fourth pooled thread
+/// waiting on the resulting `DispatchGroup`.
+///
+/// The global concurrent queues are **non-overcommit**: libdispatch will not
+/// spin up a thread for a queued item just because the existing ones are
+/// blocked in the kernel. It notices the stall on a delay and grows the pool
+/// slowly, up to a hard cap on constrained threads. So the more of these waits
+/// are in flight, the longer an item sits before it is *scheduled at all* —
+/// which is exactly the "10-15s of dispatch latency on this CI image" that
+/// several comments in this codebase already record as ordinary.
+///
+/// That latency is not a slow wait, it is a wait that has not started, and it
+/// is unbounded in the bad case. It produced this signature on CI:
+///
+/// ```text
+/// git checkout -b main did not exit within 60s and was killed
+///   (stderr so far: Switched to a new branch 'main'
+/// ```
+///
+/// `Switched to a new branch 'main'` is git's own success message. git ran,
+/// succeeded, and wrote to a pipe we were reading — and 65s later the item
+/// that was supposed to observe its exit still had not run. The two
+/// `LLMSandboxDirectoryTests` timeouts in the same run have the same shape:
+/// each overshot its bound by ~6s, which is precisely the SIGTERM + SIGKILL
+/// grace, i.e. the exit observer never reported in even after the child was
+/// killed.
+///
+/// A thread of our own cannot be starved and cannot starve anybody else, and
+/// if it does leak (a grandchild holding a pipe open forever) it costs one
+/// thread rather than one slot in a pool the whole process shares.
+///
+/// This is deliberately not a queue. A `DispatchQueue(label:)` would also
+/// work — private serial queues target an *overcommit* root queue, so they
+/// always get a thread — but that is a subtle property to rely on, and it is
+/// invisible at the call site. `Thread` says what it means.
+enum BlockingWork {
+
+    /// Start `body` on a thread of its own, right now.
+    ///
+    /// `name` shows up in a sample / crash report, which is the whole reason
+    /// to prefer this over `Thread.detachNewThread`: the point of the change
+    /// is to make "who is blocked on what" legible, and an unnamed thread is
+    /// not.
+    static func onDedicatedThread(named name: String,
+                                  _ body: @escaping @Sendable () -> Void) {
+        let thread = Thread(block: body)
+        thread.name = name
+        thread.start()
+    }
+}
--- a/Mila/Actions/LLMRunner.swift
+++ b/Mila/Actions/LLMRunner.swift
@@ -563,7 +563,10 @@ enum LLMRunner {
         let handle = ProcessHandle()
         return try await withTaskCancellationHandler {
             try await withCheckedThrowingContinuation { continuation in
-                DispatchQueue.global(qos: .userInitiated).async {
+                // A thread of its own: `executeProcess` blocks for the whole
+                // life of the child, and a pooled thread parked there starves
+                // everything else queued behind it (issue #246).
+                BlockingWork.onDedicatedThread(named: "io.island.mila.llm.run") {
                     do {
                         let outcome = try executeProcess(executable: executable,
                                                          arguments: arguments,
@@ -840,7 +843,9 @@ enum LLMRunner {
         let handle = ProcessHandle()
         return await withTaskCancellationHandler {
             await withCheckedContinuation { continuation in
-                DispatchQueue.global(qos: .userInitiated).async {
+                // See `run` above — `executeProcess` blocks, so it must not
+                // hold a shared pool thread (issue #246).
+                BlockingWork.onDedicatedThread(named: "io.island.mila.llm.diagnose") {
                     let elapsed: () -> TimeInterval = { Date().timeIntervalSince(start) }
                     do {
                         let outcome = try executeProcess(executable: executable,
@@ -1425,6 +1430,21 @@ enum LLMRunner {
         process.standardOutput = stdoutPipe
         process.standardError = stderrPipe
 
+        // Observe the exit WITHOUT parking a blocking wait on the shared
+        // `DispatchQueue.global()` pool (issue #246 — the reasoning is in
+        // `BlockingWork`). `terminationHandler` is driven by Foundation's own
+        // child-reaping source, so nothing of ours is blocked to notice the
+        // exit; it has to be installed BEFORE `run()`.
+        //
+        // The `waitUntilExit()` backstop further down is not redundant with
+        // it: this method already documents a macOS 26 reaping race where
+        // that call never returns even after `SIGKILL`, so the two paths can
+        // fail independently. Whichever notices first signals. Only one wait
+        // below ever succeeds, so a second signal is harmless (a semaphore is
+        // only unsafe to destroy while its value is BELOW where it started).
+        let exited = DispatchSemaphore(value: 0)
+        process.terminationHandler = { _ in exited.signal() }
+
         do {
             try process.run()
         } catch {
@@ -1439,8 +1459,8 @@ enum LLMRunner {
         // result the user no longer cares about.
         handle.attach(process)
 
-        // Read stdout/stderr eagerly on background queues so a chatty CLI
-        // can't deadlock by filling the OS pipe buffer while we waitUntilExit.
+        // Read stdout/stderr eagerly on background threads so a chatty CLI
+        // can't deadlock by filling the OS pipe buffer while we wait for exit.
         // Chunked reads into lock-guarded boxes (not one readDataToEndOfFile
         // into a captured var) so the bounded drain below can snapshot what
         // arrived so far without racing a still-blocked reader.
@@ -1449,7 +1469,13 @@ enum LLMRunner {
         let group = DispatchGroup()
         for (pipe, box) in [(stdoutPipe, outBox), (stderrPipe, errBox)] {
             group.enter()
-            DispatchQueue.global().async {
+            // A thread of its own, NOT the global pool: `availableData` blocks
+            // until EOF, and a pooled thread parked there is one the whole
+            // process cannot use (issue #246, see `BlockingWork`). It also
+            // means the documented "an orphaned grandchild still holds the
+            // pipe" leak below costs a thread instead of a pool slot.
+            let stream = pipe === stdoutPipe ? "stdout" : "stderr"
+            BlockingWork.onDedicatedThread(named: "io.island.mila.llm.drain.\(stream)") {
                 let handle = pipe.fileHandleForReading
                 while true {
                     let chunk = handle.availableData   // blocks until data or EOF
@@ -1460,21 +1486,23 @@ enum LLMRunner {
             }
         }
 
-        // Bounded wait — kill the process if it's still running at deadline.
-        let runningGroup = DispatchGroup()
-        runningGroup.enter()
-        DispatchQueue.global().async {
+        // Backstop for `terminationHandler`, on a thread of its own for the
+        // same reason as the readers — and so that a `waitUntilExit()` which
+        // never returns (see below) leaks one thread rather than a pool slot.
+        BlockingWork.onDedicatedThread(named: "io.island.mila.llm.reap") {
             process.waitUntilExit()
-            runningGroup.leave()
+            exited.signal()
         }
+
+        // Bounded wait — kill the process if it's still running at deadline.
         let deadline = DispatchTime.now() + .seconds(Int(timeout.rounded(.up)))
-        let timedOut = runningGroup.wait(timeout: deadline) == .timedOut
+        let timedOut = exited.wait(timeout: deadline) == .timedOut
         if timedOut {
             // SIGTERM first; if the CLI ignores it, hard-kill after a short
             // grace period so we don't read half-drained pipes or remove the
             // sandbox out from under a still-running child.
             process.terminate()
-            if runningGroup.wait(timeout: .now() + 1) == .timedOut {
+            if exited.wait(timeout: .now() + 1) == .timedOut {
                 kill(process.processIdentifier, SIGKILL)
                 // BOUNDED: `waitUntilExit` has been observed never returning
                 // even after a SIGKILL (macOS 26, sampled live — likely a
@@ -1482,7 +1510,7 @@ enum LLMRunner {
                 // either way, and this is the timedOut path so its exit
                 // status is already being discarded — don't hang the caller
                 // (and its awaiting continuation) on the obituary.
-                if runningGroup.wait(timeout: .now() + 5) == .timedOut {
+                if exited.wait(timeout: .now() + 5) == .timedOut {
                     // `error`: a child that outlives SIGKILL is a genuine
                     // anomaly (see the comment above), and it explains a
                     // truncated result — it must not need `--debug` to see.
--- a/Mila/Actions/ObsidianGitSyncer.swift
+++ b/Mila/Actions/ObsidianGitSyncer.swift
@@ -133,11 +133,30 @@ actor ObsidianGitSyncer {
 struct ProcessGitCommandRunner: GitCommandRunning {
     var timeout: TimeInterval = 120
 
+    /// Extra environment applied to the child `git` only, after the defaults
+    /// below and before `run()`. Empty in production.
+    ///
+    /// It exists so a test can isolate git's config discovery
+    /// (`GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, `HOME`) **per invocation**
+    /// instead of via `setenv` on the test host (issue #246). The host is a
+    /// live SwiftUI app: `setenv` can `realloc` `environ` underneath a
+    /// concurrent `getenv`/`posix_spawn` on another thread, which is a real
+    /// data race and not one the suite doing it can contain.
+    var environment: [String: String] = [:]
+
     func run(_ arguments: [String], in directory: URL) async -> GitCommandResult {
         let timeout = self.timeout
+        let environment = self.environment
         return await withCheckedContinuation { continuation in
-            DispatchQueue.global(qos: .utility).async {
-                continuation.resume(returning: Self.runSync(arguments, in: directory, timeout: timeout))
+            // A thread of its own, not `DispatchQueue.global()`: `runSync`
+            // blocks for the whole life of the child, and a pooled thread
+            // parked there starves whatever is queued behind it (issue #246,
+            // see `BlockingWork`).
+            BlockingWork.onDedicatedThread(named: "io.island.mila.git.run") {
+                continuation.resume(returning: Self.runSync(arguments,
+                                                            in: directory,
+                                                            timeout: timeout,
+                                                            environment: environment))
             }
         }
     }
@@ -181,7 +200,8 @@ struct ProcessGitCommandRunner: GitCommandRunning {
 
     private static func runSync(_ arguments: [String],
                                 in directory: URL,
-                                timeout: TimeInterval) -> GitCommandResult {
+                                timeout: TimeInterval,
+                                environment: [String: String]) -> GitCommandResult {
         guard let git = gitExecutable() else {
             return GitCommandResult(exitCode: -1, stdout: "",
                                     stderr: "git executable not found on PATH", timedOut: false)
@@ -214,6 +234,9 @@ struct ProcessGitCommandRunner: GitCommandRunning {
         env["LC_ALL"] = "C"
         env["LANG"] = "C"
         env.removeValue(forKey: "LANGUAGE")
+        // Caller overrides last, so a test can pin git's config discovery for
+        // this child without touching the host process's `environ`.
+        for (key, value) in environment { env[key] = value }
         process.environment = env
 
         let stdinPipe = Pipe()
@@ -222,6 +245,13 @@ struct ProcessGitCommandRunner: GitCommandRunning {
         process.standardInput = stdinPipe
         process.standardOutput = stdoutPipe
         process.standardError = stderrPipe
+
+        // Exit notification with no blocking wait on the shared pool — same
+        // change, same reason, as `LLMRunner.executeProcess` (issue #246).
+        // Must be installed before `run()`. The `waitUntilExit()` backstop
+        // below covers the macOS 26 reaping race the runner documents.
+        let exited = DispatchSemaphore(value: 0)
+        process.terminationHandler = { _ in exited.signal() }
 
         do {
             try process.run()
@@ -239,7 +269,10 @@ struct ProcessGitCommandRunner: GitCommandRunning {
         let drain = DispatchGroup()
         for (pipe, box) in [(stdoutPipe, outBox), (stderrPipe, errBox)] {
             drain.enter()
-            DispatchQueue.global().async {
+            // Dedicated threads, not the global pool: `availableData` blocks
+            // until EOF (issue #246, see `BlockingWork`).
+            let stream = pipe === stdoutPipe ? "stdout" : "stderr"
+            BlockingWork.onDedicatedThread(named: "io.island.mila.git.drain.\(stream)") {
                 let handle = pipe.fileHandleForReading
                 while true {
                     let chunk = handle.availableData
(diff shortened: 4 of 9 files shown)


urisland:
Pushed `b2df3cb`. All four items below. Verified each against the code before acting, per `.claude/rules/pull-requests.md`.

## 1. The red — Bugbot's mechanism is correct, and it is the harness, not production

**Verified, not taken on faith. It is the harness.**

`LLMTool.arguments` on this branch is **byte-for-byte identical to `main`** (`git diff main -- Mila/Actions/LLMSettings.swift` is empty), and its session branch is intact:

```swift
case .new(let id):    args.append(contentsOf: ["--session-id", id.uuidString])
case .resume(let id): args.append(contentsOf: ["--resume", id.uuidString])
```

The defect was in the `runEcho` helper this PR introduced: it took `lines.first { $0.hasPrefix("ARGV:") }`, but `ARGV:` is **not one line**. The composed prompt is a single argv element containing newlines, and `$*` preserves them, so the ARGV output spans several lines and a first-line-only parse yields exactly `-p x` — the truncation you and Bugbot both read off the message shape. Fixed in `ca56c4b` by splitting on the marker rather than on newlines.

So: no production regression. Live AI session continuity is unaffected. Worth having ruled out — thank you for insisting.

## 2. …but "capture the full argv" is not sufficient on its own, so `b2df3cb` goes further

Capturing all of `$*` stops the truncation, but `$*` has **already collapsed argument boundaries** by the time Swift sees it. So `argv` alone still cannot distinguish "the CLI was never given `--session-id`" from "our parse dropped it" — which is precisely the ambiguity that made the first failure read like a production regression. A string search over a joined blob is not a discriminating assertion about argument passing.

The fixture now emits a count taken **before** the join:

```
CWD:<pwd>          one line
ARGC:<count>       one line — $#, before any joining
ARGV:<joined>      NOT one line
```

and the test asserts the count *before* it looks at the uuid:

```swift
XCTAssertEqual(newSession.argc, oneShot.argc + 2, "a new session must add `--session-id <uuid>`: …")
XCTAssertEqual(resumed.argc,    oneShot.argc + 2, "a resumed session must add `--resume <uuid>`: …")
```

Nothing else differs between the four runs (same prompt, same transcript, no model, no extra args), so the delta is exactly 2. This check depends on no string matching at all, and it is the one that would actually catch the regression you were worried about. The `contains(uuid)` assertions stay as the "and it is the *right* uuid" half, along with the two negative cases.

I did consider NUL/record-separated framing with one arg per line. It cannot be done cleanly here — the arguments themselves contain newlines, so any line-oriented framing needs an escape scheme, and that is more parser to get wrong for no extra signal over "count + contents". The count is the part that closes the hole.

## 3. `python-subprocess.md:77-80` — correct, the rule misdescribed the code

CodeRabbit is right and it matters more than usual: a rules file that misdescribes the implementation actively misleads the next reader. The implementations do escalate — first wait at the caller's timeout, then `+1`/`+2` after SIGTERM, then `+5`/`+3` after SIGKILL. Reworded to what they do: **every** wait is bounded, duplicate signals are harmless (a semaphore is only unsafe to destroy while its value is *below* where it started), and never add an unbounded `wait()`.

## 4. `ObsidianGitSyncerIntegrationTests:92` — correct, the last test's leak was unreportable

`try?` swallowed a failed removal, and the ledger is only read by the *next* `setUp`, so a leak from the final test in the class had nowhere to surface. `tearDownWithError` now reports a failed removal — which still attributes correctly, since `tearDown` runs in the failing test's own context — and prunes the ledger on success. The `setUp` ledger stays as the cross-test net for the cases `tearDown` cannot see: a root that reappears, or a `tearDown` that never ran.

## 5. The thread-ceiling question — the honest answer, with numbers

This is the right thing to push on, and I do not think an artificial cap is the answer. Here is the accounting, now written into `BlockingWork`'s doc comment so it is reviewable rather than asserted in a PR thread.

An invocation costs **four** threads for the child's lifetime: two pipe readers, one `waitUntilExit()` backstop, and the caller blocked inside `executeProcess`/`runSync`. So the ceiling is 4 × concurrent **child processes** — and every path that spawns one is already bounded upstream, by limits that exist to bound *processes*, which is the scarcer resource:

| caller | what bounds it | children |
|---|---|---|
| `RecordingSummarizer` backfill / regeneration | `maxConcurrent = 2` with `backfillQueue` holding the rest — its comment says it exists "so a 20-recording catch-up sweep doesn't fork 20 `claude -p` subprocesses" | 2 |
| `LiveAISession` ticks | single `inFlight` Task + coalescing | 1 |
| `PostRecordingCoordinator` | single in-flight handle | 1 |
| `RenameRecordingSheet` "Suggest" | one modal sheet, guarded by `isFetchingName` | 1 |
| `LLMSettings.diagnose` (Settings → Test) | one Settings panel, user-driven | 1 |
| `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`, and `sync` issues its commands sequentially | 1 |

Your worst case — "batch re-summarise while a live meeting ticks and Obsidian syncs" — is therefore ~7 children and ~28 transient threads, against a per-process limit in the hundreds. The batch case, the only one that could plausibly fan out, is the one that is *explicitly* capped, at 2.

Two things worth stating plainly:

- **This is the same concurrency as before.** The old code created exactly the same number of work items per invocation. The difference is that they contended for a capped shared pool — which is what starved. The change moves where the concurrency lands; it does not raise it.
- **A cap inside `BlockingWork` would be a regression, not a safeguard.** Limiting thread creation there means re-queuing blocked work behind a bound, which *is* the original bug in a smaller box. If a future caller can spawn subprocesses in a batch, the limit belongs on that caller, like `RecordingSummarizer.maxConcurrent`. That is now written as an explicit "do not add a cap here" in the type's documentation and in the rules file.

If you would still rather have a hard ceiling, the place I would put it is a concurrency limit on `executeProcess` entry (bounding children, not threads) — say it and I will add it, but I would be adding a second cap behind existing ones rather than closing a gap.

---

urisland:
## Results on `b2df3cb` — and the compression flake is (2), pre-existing, with a mechanism

### 1. Did `build-and-test` compile?

**Yes.** `build-and-test` **success**, 14:50:12 → 14:55:12 (job 99899175475). First real compile signal on this branch, and it is clean.

### 2. Did `unit-and-ui-tests` pass, and did the flaky family stop failing?

**Passed** (job 99899172181, 14:43:30 → 14:49:55). The `Summarise test failures` step reports **exactly one** failed execution across the whole bundle, and it is the compression test below.

The point that matters: **none of the #246 family appear in that list.** No `LLMSandboxDirectoryTests` (including `:251`/`:253` and `test_runner_leaves_the_shared_directory_in_place`), no `ObsidianGitSyncerIntegrationTests.test_sync_rebases…`, no `LLMRunnerTests`, no `LiveAIDeletedLineSessionTests`. Not a timeout anywhere in the run.

Being honest about the strength of that: it is **one** clean run against a ~20% historical rate, so it is consistent with the fix working and not yet proof of it. What raises it above "one lucky run" is that the four tests failed as a *group* under load and the run is now clean of all four at once, and that the diagnosis predicted exactly which tests would stop failing and why. The number to keep watching is still the failed-execution count.

### 3. The compression failure: pre-existing, and I can point at the line

**It is (2), and here is the evidence rather than a judgement call.**

The test's synchronisation step is:

```swift
// (The first pass also auto-kicks a compression; awaiting here is idempotent
// and guarantees the on-disk file is the .m4a regardless of which won.)
await store.compressRecordingAudio(id: fixture.recording.id)
```

**That comment is false.** `compressRecordingAudio` is not "wait for compression", it is "start one *unless one is already running*":

```swift
guard !compressingIDs.contains(id) else { return }        // RecordingStore.swift:322
```

and the first pass kicks an unawaited one that `waitForIdle()` does not cover:

```swift
Task { await store.compressRecordingAudio(id: compressID) }   // TranscriptionService.swift:1040
```

So the `await` is a barrier in only two of three states — auto-kick finished, or not yet started. In the third, **auto-kick mid-flight**, the dedup guard returns immediately and the test awaits nothing, then asserts against pre-compression state. The **0.124s** runtime is the corroboration: the test never paid for an AAC encode.

**Does my change reach it?** No, on both of your criteria:

- `MilaTests/TranscriptionServiceTests.swift`, `Mila/Models/RecordingStore.swift`, `Mila/Audio/AudioCompressor.swift` and `Mila/Transcription/TranscriptionService.swift` are **byte-for-byte identical to `main`** on this branch — `git diff main --stat` over all four is empty. The test was last touched in #189.
- `AudioCompressor` imports only Foundation / AVFoundation / OSLog. No `Process`, no `DispatchQueue`, nothing routing through `BlockingWork`.

I won't claim immunity, though: freeing blocked threads off the shared pool changes *when* everything else gets scheduled, which is plausibly enough to move a latent race into the reachable window. That makes this PR a **trigger**, not a cause — the defect is in the test and is reachable on `main` today.

Filed as **#255** with the full mechanism and a suggested fix (wait for the observable post-condition — `.m4a` present, `.wav` gone — rather than treating a dedup guard as a barrier; not a widened timeout, not a retry). Following your instruction for case (2), I have **not** folded it into this PR: it is a different subsystem, it would muddy this PR's signal, and it would cost another 20-minute CI round while three PRs wait.

### 4. What bounds concurrent native threads

Answered in [the previous comment](https://github.com/island-io/mila/pull/251#issuecomment-5495634465) and now written into `BlockingWork`'s doc comment so it is reviewable in-tree rather than asserted in a thread. Short version:

An invocation costs **four** threads for the child's lifetime (2 pipe readers, 1 `waitUntilExit()` backstop, 1 blocked caller), so the ceiling is 4 × concurrent **child processes** — and every spawning path is already bounded upstream, by limits that exist to bound *processes*:

| caller | bound | children |
|---|---|---|
| `RecordingSummarizer` backfill / regeneration — your batch case | `maxConcurrent = 2` + `backfillQueue`; its comment says it exists "so a 20-recording catch-up sweep doesn't fork 20 `claude -p` subprocesses" | 2 |
| `LiveAISession` ticks | one `inFlight` Task + coalescing | 1 |
| `PostRecordingCoordinator` | one in-flight handle | 1 |
| `RenameRecordingSheet` "Suggest" | one modal sheet, guarded by `isFetchingName` | 1 |
| `LLMSettings.diagnose` | one Settings panel, user-driven | 1 |
| `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`; `sync` issues commands sequentially | 1 |

Your worst case — batch re-summarise during a live meeting with an Obsidian sync — is ~7 children, ~28 transient threads, against a per-process limit in the hundreds. The only path that could fan out is the one that is *explicitly* capped, at 2.

Two things worth stating plainly: this is the **same** concurrency as before (the old code created the same work items; they just contended for a capped pool, which is what starved — the change moves where concurrency lands, it does not raise it). And a cap inside `BlockingWork` would be a **regression, not a safeguard**: limiting thread creation there re-queues blocked work behind a bound, which is the original bug in a smaller box. That is now an explicit "do not add a cap here" in the type's docs and in `.claude/rules/python-subprocess.md`, pointing new batch callers at `RecordingSummarizer.maxConcurrent` as the pattern.

If you want a hard ceiling anyway, the right place is a concurrency limit on `executeProcess` entry — bounding children, not threads. Say the word and I'll add it, but it would be a second cap behind existing ones rather than closing a gap.

### Status

`build-and-test` ✅ · `unit-and-ui-tests` ✅ · `Cursor Bugbot` ✅ (success, zero findings, on this exact head) · `Linked issue` ✅. Remaining e2e/CodeQL jobs still queued behind the macOS pool.

---

urisland:
Pushed `67ab7ad`.

## `LiveAIDeletedLineSessionTests` — fixed on the principle, but the principle gives *two* answers

Your diagnosis is right and I've taken it. One correction to the prescription, which I'd rather flag than implement as written (`.claude/rules/pull-requests.md`: right in substance, wrong in shape).

**:233 and :237 are negative assertions.**

```swift
let wentIdle = await waitUntil(timeout: 1) { !session.isThinking }
XCTAssertFalse(wentIdle, "the cancelled tick cleared `isThinking` while …")

let startedAThird = await waitUntil(timeout: 1) { calls.value.count == 3 }
XCTAssertFalse(startedAThird, "`inFlight` was clobbered: …")
```

There the timeout is not a deadline for something to happen — it is a **window in which nothing must happen**. `waitUntil` returns the instant its condition holds, so a negative assertion spends its full budget only when it is **passing**. Raising those two to 30s would have added a minute to every green run for zero gain, and it would make a *spurious pass* more likely rather than less, since a longer window gives unrelated scheduling more chance to satisfy the condition. Same number, opposite meaning.

So the file now names both bounds and says which is which:

| | value | for | why that direction |
|---|---|---|---|
| `mustHappen` | **30s** | waits asserted `true` — :170, :181, :216, :224, :243, and `waitUntilIdle` | a genuine bug leaves the state wrong *forever*, so no amount of extra waiting turns a real failure green. Generous costs nothing and is never spent on a passing run. |
| `quietWindow` | **1s** | waits asserted `false` — :233, :237 | the budget is spent only when passing; widening slows every green run and weakens the check |

The default on `waitUntil` moves to `mustHappen`, so a negative caller has to opt into the short window explicitly rather than inheriting a bound meant for the other polarity. Both constants carry the reasoning in their doc comments, including an explicit "do not re-tighten these", so the next person has the argument rather than just the number.

Worst case per test after the change: 3 × 30s = 90s, comfortably inside the 240s allowance (the two tests with the most waits come to 60s and 92s).

**The `:182` message.** Worth the churn, yes — it cost time today. It asserted a diagnosis as fact ("Live AI wedged") when what had been observed was a wait expiring, and those are different findings. It now reports what was seen and what it would mean if real.

Nothing weakened, nothing deleted, no retries.

And agreed on scope: a test whose bound is tight enough that ordinary contention trips it is exactly #250's subject. Confirmed independently that this is not my change — the same test failed on #248's `e8be45d` and #252's `f0bb018`, neither of which carries this diff.

## Thread cap — answered, and it is in the tree

Apologies, I think we've been crossing in flight. This is answered in three places, all pushed:

1. **`Mila/Actions/BlockingWork.swift`**, in the type's doc comment — a "What bounds the thread count" section with the full table. That is the in-tree, reviewable version, and it is what I added to the file rather than only asserting it in a thread.
2. **The PR description**, section "What bounds the thread count".
3. **[This comment](https://github.com/island-io/mila/pull/251#issuecomment-5495634465)** (item 5) and **[this one](https://github.com/island-io/mila/pull/251#issuecomment-5495936854)** (item 4).

The short form, once more:

An invocation costs **4** threads for the child's lifetime — 2 pipe readers, 1 `waitUntilExit()` backstop, 1 blocked caller — so the ceiling is 4 × concurrent **child processes**, and every spawning path is already bounded upstream by a limit that exists to bound *processes*:

| caller | bound | children |
|---|---|---|
| `RecordingSummarizer` backfill / regeneration — your batch case | `maxConcurrent = 2` + `backfillQueue`; its comment says it exists "so a 20-recording catch-up sweep doesn't fork 20 `claude -p` subprocesses" | 2 |
| `LiveAISession` ticks | one `inFlight` Task + coalescing | 1 |
| `PostRecordingCoordinator` | one in-flight handle | 1 |
| `RenameRecordingSheet` "Suggest" | one modal sheet, guarded by `isFetchingName` | 1 |
| `LLMSettings.diagnose` | one Settings panel, user-driven | 1 |
| `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`; `sync` runs its commands sequentially | 1 |

Your exact scenario — batch re-summarise during a live meeting with an Obsidian sync — is **~7 children, ~28 transient threads**, against a per-process limit in the hundreds. The only path that could fan out is the one that is explicitly capped, at 2. There is no re-diarize path through `BlockingWork` at all: `SpeakerDiarizer` and `DiarizationBootstrap` use `Task.detached`, which this PR does not touch.

Two things I want on the record rather than buried:

- **This is the same concurrency as before.** The old code created the same number of work items per invocation; they contended for a capped pool, which is what starved. The change moves where the concurrency lands, it does not raise it. There is no new fan-out to cap.
- **A cap inside `BlockingWork` would be a regression, not a safeguard.** Limiting thread creation there re-queues blocked work behind a bound — the original bug in a smaller box. That is now an explicit "do not add a cap here" in the type's docs and in `.claude/rules/python-subprocess.md`, pointing new batch callers at `RecordingSummarizer.maxConcurrent` as the pattern to copy.

If you want a hard ceiling regardless, the correct place is a concurrency limit at `executeProcess` entry — bounding **children**, not threads, so it also bounds the CPU and memory a fan-out would cost. Say so and I'll add it in this PR; I've held off only because it would be a second cap sitting behind existing ones rather than closing a gap, and I did not want to add unvalidated production behaviour to a flake fix.

---

urisland:
Pushed `ff8d2e3`.

# What bounds concurrent native threads — the answer, up front

**Nothing new needs to. Every path that spawns a child process is already bounded upstream, and this PR does not raise the concurrency — it only moves where it lands.**

It is in the tree, not just in a thread: `Mila/Actions/BlockingWork.swift`, the `## What bounds the thread count` section of the type's doc comment. Also in `.claude/rules/python-subprocess.md` under **Bounding the threads**. Sorry for the runaround — I answered in the PR body and three comments and clearly never got it in front of you.

Four threads per child (2 pipe readers, 1 `waitUntilExit()` backstop, 1 blocked caller), so the ceiling is 4 × concurrent **children**:

| caller | bound | children |
|---|---|---|
| `RecordingSummarizer` backfill/regeneration | `maxConcurrent = 2` + `backfillQueue` — exists "so a 20-recording catch-up sweep doesn't fork 20 `claude -p` subprocesses" | 2 |
| `LiveAISession` ticks | one `inFlight` Task + coalescing | 1 |
| `PostRecordingCoordinator` | one in-flight handle | 1 |
| `RenameRecordingSheet` "Suggest" | one modal sheet, `isFetchingName` | 1 |
| `LLMSettings.diagnose` | one Settings panel | 1 |
| `ProcessGitCommandRunner` via `ObsidianGitSyncer` | an `actor`; `sync` runs commands sequentially | 1 |

Batch re-summarise + live meeting + Obsidian sync = **~7 children, ~28 transient threads**, against a per-process limit in the hundreds. The only path that can fan out is the one explicitly capped at 2. **Re-diarize does not go through `BlockingWork` at all** — `SpeakerDiarizer` and `DiarizationBootstrap` use `Task.detached`, untouched here.

Two points I want on the record:

- **Same concurrency as before.** The old code created the same work items per invocation; they contended for a capped pool, which is what starved. No new fan-out exists to cap.
- **A cap inside `BlockingWork` would be a regression, not a safeguard** — limiting thread creation re-queues blocked work behind a bound, the original bug in a smaller box. Now an explicit "do not add a cap here" in the docs, pointing new batch callers at `RecordingSummarizer.maxConcurrent`.

If you want it **enforced** rather than documented, the right shape is a `DispatchSemaphore` at `executeProcess` entry bounding **children** (blocking a dedicated thread there is safe — that is what they are for). I have deliberately not added it: it is unvalidated production behaviour, it would sit behind existing caps rather than close a gap, and an unbounded acquire is the very thing the new rule forbids. Say the word and I will open it as its own PR with its own issue, where it can be reviewed on its merits instead of riding a flake fix.

---

# The Microphone precondition

Taken, and #246's stated precondition is met: six failure messages across two PRs on different runners, yielding 6–10 against a bar of "> 10". That is not a mystery needing diagnosis — the assertion sits on the value the system produces.

**The backwards comment is fixed.** #253's agent is right and it matters: `Task.sleep(15ms)` is a floor, so a loaded machine makes each iteration cost *more* and **fewer** fit in the fixed window. The old comment claimed a slow runner "stretches the number of posts" — the opposite — and would have sent the next reader hunting in the wrong direction. Nominal yield is 40; CI sees 6–10, i.e. ~4–6× overhead.

**Only the bound without timing meaning moves.** I took the separation as described:

- **`burstWindow` (0.6s) — untouched**, now named, with its constraint written down: it must stay under the 1s `minimumRebuildInterval` or "at most one rebuild" stops being the right expectation and the policy gets judged over a window it was never given.
- **`posts > 10` → `posts > 2 * maximumExpectedRebuilds`.** The precondition only has to establish that enough notifications arrived for "paced" and "one rebuild per notification" to be *visibly different outcomes*. At twice the rebuild bound they already are: the bug would have produced `posts` rebuilds, and `posts` is at least double what is allowed. So the bar is now derived from the assertion it supports, rather than from the yield the loop happens to produce. That is the difference between this and the blind tuning #246 forbids — and it is why I did not pick a new magic number. The comment says explicitly not to raise it back toward the nominal yield, and why.

I did **not** use the "wait until enough notifications arrive" shape from your first message: that would extend the window past 0.6s and break the very property the window is load-bearing for. Your third message reaches the same conclusion; flagging it only so the reasoning is on the record.

`restartCount <= 2` is unchanged. It now reads through the same `maximumExpectedRebuilds` constant as the precondition, so the two cannot drift apart — which is how the old pair got into this state.

Margin check: the bar is now 4, against a worst observation of 6. Tripping it would need ~10× scheduling overhead on a nominal-15ms loop, at which point the whole bundle is in trouble anyway.

**On the "your threads made it worse" hypothesis** — noted as withdrawn, and I agree #253's occurrence is the cleaner data point: that PR touches none of this code, spawns no subprocesses, and failed at 10/9/10 regardless. I have not spent time on it.

---

## Status on `ff8d2e3`

Previous head `67ab7ad` had all four subprocess tests, `LiveAIDeletedLineSessionTests` and the compression test green — three diagnoses confirmed in one run — with `build-and-test` ✅ and Bugbot ✅. This push adds only the Microphone precondition, so it needs a fresh `unit-and-ui-tests` and a fresh Bugbot pass on this head. Holding all further pushes until both report.

---

Issue #201, opened 2026-08-27
MCP transcript search reads one sidecar per recording on every call


urisland:
Deferred from the CodeRabbit review of #183, where the reviewer itself rated it acceptable at the current scale.

## The cost

`MilaStoreReader.searchTranscripts` calls `namedTranscript` for every non-trashed recording, and that reads the recording's `.txt` sidecar from disk and formats the whole transcript. One `search_transcripts` tool call therefore performs **N synchronous file reads plus N full-transcript formats**, where N is the total recording count. The `limit` argument cannot reduce it, because relevance ranking has to score every candidate before it can take the top few.

At a few hundred recordings this is fine. It scales linearly with a store that only grows.

## Why it wasn't fixed in #183

The obvious fix — cache transcript text keyed by recording id and file mtime — conflicts with two properties that PR deliberately established:

1. **`MilaStoreReader` is stateless on purpose.** Its own documentation says every call re-reads from disk because "freshness beats caching for a live assistant". A cache changes that contract.
2. **The whole MilaKit chain is now `Sendable`, verified under `StrictConcurrency`.** A mutable cache would need a lock or an actor, which either undoes that or moves the type into async-only territory — a real design change, not a tweak.

So this wants its own change with a considered approach, not a cache rushed into a review round on an unrelated PR.

## Options worth weighing

- **An mtime-keyed cache behind an actor**, accepting that reads become `async`.
- **A small on-disk search index** the app maintains on save, which also makes search cheap for the app's own UI rather than only for MCP.
- **Bound the work**: search titles first and only read transcripts for a capped candidate set, accepting slightly worse recall for a large store.
- **Do nothing until it bites**, and record the ceiling in `MilaStoreReader`'s doc comment so the next reader knows the cost is deliberate.

## Trigger

Worth acting on when a real store passes a few hundred recordings, or if `search_transcripts` starts showing up as slow in practice. Until then the honest thing is that the current behaviour is a known, documented trade rather than an oversight.

PR #252, opened 2026-09-01
perf(mcp): relevance search scores from raw segments instead of rendering every transcript

### LINKED ISSUE #241: search_transcripts still renders every transcript when sorting by relevance
## What's left over from #201

#201 (fixed in #225) bounded most of what one `search_transcripts` call costs:

- diarized recordings no longer pay a wasted `.txt` sidecar read each (`fallback:` became an `@autoclosure`, so the discarded read never happens);
- `sort: created_at` now stops at the `limit`-th hit, because date order *is* the traversal order;
- a non-matching recording no longer pays the per-line split, count and snippet pass.

**`sort: relevance` — the default — is still linear in the size of the store.** Ranking by match count cannot know the top few before it has scored them all, so every non-trashed recording is still rendered once per search, plus one sidecar read per *undiarized* one. #225 shrank the constant; it did not remove the linear term, and the doc comment on `searchTranscripts` says so rather than implying the problem is solved.

This is filed so that remaining ceiling is tracked somewhere other than a code comment, now that #201 is closed.

## Why the easy fixes were already ruled out

Both were weighed in #201/#225 and rejected for reasons that still hold — worth restating so they don't get re-proposed:

- **An mtime-keyed transcript cache** breaks two properties #183 established deliberately. `MilaStoreReader` is stateless on purpose ("every call re-reads from disk — the store is small, the app's writes are atomic, and freshness beats caching for a live assistant"), and for a store whose rows can be **trashed** between calls a cache is a deleted-data-stays-deleted hazard. It also needs a lock or an actor: the first reintroduces exactly the shape `MilaKitSendabilityTests.test_the_whole_data_source_chain_is_sendable` exists to prevent, the second makes every read `async` and drags `MilaDataSource` and all four tool handlers with it.
- **Capping the candidate set** ("search titles first") silently drops the best hit in a large store. That is a worse answer, not a cheaper one.

## What would actually settle it

An on-disk index — the option #201 also weighed. It is the only shape that gives sublinear relevance ranking without either caching mutable state in a type that is required to stay `Sendable` and stateless, or degrading the answer.

Worth being honest that this may never need doing: the linear term only bites in the tens of thousands of recordings, and nobody has reported it. Two things would make it real, and either is a better trigger than a hunch:

- a measured `search_transcripts` latency complaint from an actual store, or
- a store size in the field that makes the arithmetic obvious.

Until then this is a documented ceiling, not a defect. Filed as `enhancement`, not a bug.

urisland:
Closes #241

## What & why

`search_transcripts` with the default `sort: relevance` rendered the full speaker-named transcript of **every** non-trashed recording, scanned it, and threw all but `limit` of them away.

Ranking by match count genuinely has to score every candidate — the issue and `CLAUDE.md` are both right that capping the candidate set is a worse answer, not a cheaper one, and this PR does not cap anything. But **scoring never needed the rendering.**

### The argument the fix rests on

`TranscriptFormatter.plainText` builds a transcript by joining pieces with exactly three separators: `": "` after a speaker label, `" "` between segments of one turn, `"\n"` between turns. Every one of them contains whitespace, and case/diacritic folding neither deletes a character nor turns a non-whitespace one into whitespace.

So **a run of the query with no whitespace in it cannot straddle two pieces**: wherever it occurs in the rendered text, it occurs inside a single `"<label>:"` or inside a single segment's text — and a segment's text is trimmed before it goes in, so an occurrence in the trimmed text is an occurrence in the raw string `recordings.json` already decoded, and vice versa.

`rawScore` uses that twice:

- **whole query is whitespace-free** (the single-token case): its occurrences are partitioned across the segments, so summing the per-segment counts *is* the per-line count. Exact, not an estimate.
- **anything else** (a phrase): the longest whitespace-free run of the query is a substring of it, so every occurrence of the query contains an occurrence of that run. No run anywhere ⇒ no match anywhere. That rules out the bulk of a store for a phrase query without building anything.

Two cases deliberately still render, and both are principled rather than lazy:

- **a recording with no speaker labels** — its transcript *is* the `.txt` sidecar, so answering means reading the file either way. Deferring the render there would read the same file twice for the hits that come back, which is the exact shape #201 removed.
- **a query a speaker label carries** — a label appears once per **turn**, and the flat segment list does not describe turns. Rare enough to pay for a rendering.

### Before / after

Per `search_transcripts(sort: relevance)` over a store of N recordings, M of which are returned (M ≤ `limit`):

| | before | after |
|---|---|---|
| transcripts rendered | N | ≤ M — and only the hits that carry a snippet |
| strings built per non-matching diarized recording | per-segment trims, per-turn buffers, per-turn formatted lines, one joined transcript | none — the scan runs over the segment strings the JSON decode already produced |
| sidecar reads | one per undiarized recording | unchanged (one per undiarized recording) |
| candidates scored | N | N |

The scan itself and the undiarized sidecar read are the irreducible part — those *are* the text — so the relevance path is still linear in the store and a store in the tens of thousands still wants the on-disk index #201 weighs. What is gone is the per-candidate rendering, which is what the issue title is about.

`sort: created_at` is untouched: it still stops at the `limit`-th hit.

## How it was tested

There is no Swift toolchain in this environment, so **nothing was compiled or run locally** — CI (`swift test` in `Packages/MilaKit`, which the CI job runs before anything else) is the verification.

New tests in `MilaStoreReaderTests`, all in the style of the existing #201 ones — they count work, because discarded work is invisible in the results:

- `test_scoring_from_raw_segments_agrees_with_the_rendered_count` — the load-bearing one. Compares the cheap count against an independent oracle that counts per line over `namedTranscript`, for 12 queries over a store built to hit every case the argument reasons about: a turn collapsed out of two segments, an untrimmed segment, a renamed speaker, a raw `SPEAKER_NN` label, an empty segment inside a turn, a segment with no speaker, a sidecar-only recording, a labels-free one, diacritics, case, a phrase, an all-whitespace query.
- `test_relevance_renders_only_the_transcripts_it_returns` — 23 candidates, 2 returned, **2 rendered** (against the unfixed reader: 23), via a new internal `TranscriptRenderer` seam that mirrors `SidecarReader`.
- `test_relevance_finds_the_best_hit_when_it_sorts_last_in_traversal` — the anti-capping guard. The highest-scoring recording is the *oldest*, i.e. the last thing the traversal reaches, and `limit: 1` must still return it.
- `test_a_phrase_query_straddling_two_segments_is_still_found` — the phrase exists only across the space the renderer inserts, so the raw scan may rule a recording IN but never OUT.
- `test_a_query_matching_a_speaker_label_is_still_counted_per_turn` — falls back to rendering and reports 2 for a speaker who holds two of three turns.
- `test_a_title_only_match_renders_nothing` — a proven-empty transcript needs no snippet and so no render, even for a returned hit.

The existing sidecar-read assertions are untouched and still meaningful: reads are unchanged everywhere (`0` for diarized stores, `limit` for the bounded date search, one per undiarized candidate, `["Live.txt", "Live.txt"]` for the trashed-filter test).

Docs updated: the `searchTranscripts` doc comment, `docs/mcp.md`'s "What a search costs", and the `CLAUDE.md` MCP bullet (which documented the old trade-off as deliberate) — plus a new bullet recording that `rawScore`'s shortcut is a proof about whitespace and what widening it would break.

## Checklist

- [ ] Builds locally (`make build`) and tests pass (`make test`) — no Swift toolchain available here; CI is the compiler
- [x] Follows the conventions in `CLAUDE.md` (no cache, still stateless and `Sendable`, MilaKit still dependency-free, trashed filter still in `MilaStoreReader`, tool logic untouched)
- [x] User-facing change is noted in the README changelog (if applicable) — internal perf, no user-visible behaviour change

---

---

> [!NOTE]
> **Medium Risk**
> Search ranking and match counts depend on subtle whitespace/formatting invariants; regressions would change MCP search results without compile-time signal, though extensive equivalence tests mitigate that.
> 
> **Overview**
> **Relevance-ranked `search_transcripts` no longer builds a full named transcript for every recording.** Scoring still walks the whole non-trashed library (no candidate cap), but diarized rows are matched via new `rawScore` logic over decoded segment text, with formatted transcripts produced only for returned hits that need snippets—at most `limit` renders instead of one per candidate (#241).
> 
> The fast path is intentional, not heuristic: because `TranscriptFormatter.plainText` only joins pieces with whitespace-containing separators, whitespace-free query runs cannot straddle pieces, so single-token queries get exact counts from segments and phrase queries use the longest whitespace-free run to rule recordings out. Undiarized recordings and speaker-label queries still go through the render path so counts stay identical to before.
> 
> `MilaStoreReader` splits **score** vs **hit** assembly (`ScoredCandidate`), adds an injectable `TranscriptRenderer` test seam (mirroring `SidecarReader`), and updates `CLAUDE.md`, `docs/mcp.md`, and doc comments. New `MilaStoreReaderTests` count renders and compare raw vs rendered match counts across edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4800873a2dd05a979c300076fc67ce988de41f29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


urisland:
## `unit-and-ui-tests` red on `f0bb018` — the #246 flake, not this PR

Recording rather than re-running.

```
LLMRunnerTests.test_runner_surfaces_nonzero_exit_code                      timedOut(30s)  37.256s
LLMSandboxDirectoryTests.test_runner_leaves_the_shared_directory_in_place  timedOut(30s)  36.118s
LiveAIDeletedLineSessionTests.test_deleting_during_an_in_flight_tick_recovers_on_a_new_session
ObsidianGitSyncerIntegrationTests.test_sync_rebases_nonconflicting_remote_change_and_pushes
    "git checkout -b main did not exit within 60s (stderr so far: Switched to a new branch 'main'"  65.958s
```

**Why none of it is this PR's.** The diff is four files: `Packages/MilaKit/Sources/MilaKit/MilaStoreReader.swift`, its test file, `CLAUDE.md` and `docs/mcp.md`. MilaKit is dependency-free by construction and is not linked by `LLMRunner`, `ObsidianGitSyncer` or `LiveAISession`. There is no path from a relevance-scoring change to a subprocess wait.

**What it is.** All four are the signature on #246/#250: a bounded wait expiring on a subprocess that has already finished its work, each overshooting its own bound by 6-8s. The Obsidian one is the clearest — `Switched to a new branch 'main'` is git's own success message, captured on stderr, and the process still did not exit within 60s. That is the second time this session that exact string has appeared in a failure.

**The systemic point.** This is now the third PR in fifteen minutes taken down by the same four tests — #248, #252 and (differently) #251's own base. Four concurrent PRs each running the full bundle is turning a ~20% per-run flake into something close to a guarantee that *something* goes red. [#251](https://github.com/island-io/mila/pull/251) takes every blocking subprocess wait off the shared thread pool and is the fix for all of it; until it lands, no PR here can be relied on to go green on its own merits.

So: no re-run spent, no change pushed. Once #251 merges, `main` gets merged into this branch and CI re-runs against a fixed base. A green re-run now would prove nothing and would leave the next PR in exactly the same position.

---

Issue #265, opened 2026-09-04
Deleting a Whisper model orphans its ~1.1GB CoreML encoder


celarent7:
## Problem

Surfaced during review of #263 (sticky model deletion, PR pending on branch `worktree-sticky-model-deletion`).

`ModelManager.delete(_:)` (`Mila/Transcription/ModelManager.swift`) only removes the model's `<name>.bin` file. The sibling CoreML encoder directory (`<name>-encoder.mlmodelc`) — 1,174,466,438 bytes for `ivritLarge`, 1,173,393,014 bytes for `openaiTurbo` — is left on disk. `ensureCoreMLInstalled()` skips re-checking it because it's gated on `isInstalled(model)` being true.

The Settings → Models delete confirmation dialog tells the user this frees the `.bin`'s size (~2.9GB for ivritLarge), but the ~1.1GB CoreML encoder silently survives — the user does not actually reclaim the disk space they were told they would, and there's no UI path to see or remove the orphan.

This was always possible, but harmless before #263 (the `.bin` came back on the next launch and the CoreML pair became consistent again, self-correcting). With deletion now sticking, the orphan is permanent until the model is manually re-downloaded.

## Suggested fix

`delete(_:)` should also remove `coreMLDirectory(for: model)` if it exists, so a full delete actually reclaims all the disk space the confirmation dialog implies, and the `.bin` + `.mlmodelc` pair stays consistent.

Issue #264, opened 2026-09-04
Deleting both Whisper models permanently breaks transcription with a misleading error


celarent7:
## Problem

Surfaced during review of #263 (sticky model deletion, PR pending on branch `worktree-sticky-model-deletion`).

`MilaApp.ensureDefaultModelsInstalled()` unconditionally calls `modelManager.setSelected(WhisperModel.ivritLarge)` on every app launch (`Mila/App/MilaApp.swift`), regardless of whether that model is installed, declined, or even the model the user actually picked in Settings → Models.

Combined with #263's fix (deletion now sticks instead of silently re-downloading), this creates a dead end:

1. User deletes both default models via Settings → Models.
2. Every subsequent launch force-selects `ivritLarge` again (now declined, so it isn't re-downloaded).
3. `ModelManager.model(for:)`'s fallback chain returns `ivritLarge` for every language.
4. Every recording attempt fails with `TranscriptionService`'s generic "Whisper model is still downloading. Try again once it's ready." — which is never true again, since nothing is downloading.
5. The only recovery (Settings → Models → Download) isn't hinted anywhere in that error message.

Before #263, this was harmless because the missing model silently came back on the next launch. After #263, deleting both models is a permanent, confusing dead end.

## Suggested fix

- Stop unconditionally calling `setSelected(WhisperModel.ivritLarge)` on every launch — only set a default when there is no persisted selection yet (`ModelManager.init` already restores a persisted selection correctly on its own).
- Make the "still downloading" error message in `TranscriptionService` conditional on `ModelManager.isDeclined(_:)` so it correctly says something like "You deleted this model — download it again in Settings → Models" instead of implying a download is in progress.

PR #269, opened 2026-09-04
fix(models): sticky model deletion, plus reselect and test-crash fallout

### LINKED ISSUE #263: Deleting a downloaded Whisper model doesn't stick — comes back on next launch
## Problem

Settings → Models → Delete removes the model's `.bin` file, but the app's launch-time bootstrap (`MilaApp.ensureDefaultModelsInstalled()`) only checks whether the file is present on disk. It has no way to tell "never downloaded" apart from "downloaded, then deleted on purpose," so the very next app launch — including right after every update, since an update is just a fresh process start — silently re-downloads the model (a multi-GB file).

## Root cause

- `Mila/App/MilaApp.swift`, `ensureDefaultModelsInstalled()`: unconditionally re-downloads both default models (`ivrit-ai large-v3`, ~3GB, and `openai large-v3-turbo`, ~1.6GB) whenever `ModelManager.isInstalled(_:)` returns false, with no persisted concept of user intent.
- `Mila/Transcription/ModelManager.swift`: `delete(_:)` only removed the file; nothing recorded that the removal was deliberate.

## Fix

Persist a `declinedModelNames: Set<String>` in `ModelManager` (`UserDefaults` key `model.declinedNames`). `delete(_:)` marks a model declined; `download(_:)` (an explicit request) clears it. `ensureDefaultModelsInstalled()` now skips declined models.

Out of scope for this fix (tracked separately, not fixed here):
- `ensureDefaultModelsInstalled()` still downloads both catalog models by default regardless of the user's selected language.
- The same function unconditionally calls `modelManager.setSelected(WhisperModel.ivritLarge)` on every launch, overwriting the user's Settings → Models selection.

### LINKED ISSUE #266: ensureDefaultModelsInstalled() re-selects a deleted model on every launch
## Problem

Follow-up to #263, which explicitly flagged this as out of scope:

> The same function unconditionally calls `modelManager.setSelected(WhisperModel.ivritLarge)` on every launch, overwriting the user's Settings → Models selection.

Even after #263's fix stopped the deleted model from being re-downloaded, `MilaApp.ensureDefaultModelsInstalled()` still unconditionally calls `modelManager.setSelected(WhisperModel.ivritLarge)` on every launch. If the user deleted `ivrit-ai large-v3` via Settings, the persisted `selectedModelName` UserDefaults value gets silently reset back to it on every restart, regardless of what the user actually wants selected.

## Root cause

`Mila/App/MilaApp.swift`, `ensureDefaultModelsInstalled()`: the `setSelected` call has no `isDeclined` guard, unlike the download loop right below it which does.

## Impact

Currently masked because every reader of `selectedModel()` re-checks `isInstalled` — except `RecordingDetailView.swift`'s `activeTranscriptionModelName`, which reads `selectedModel()?.displayName` with no installed-check, so it could show a deleted model's name.

### LINKED ISSUE #267: Unit test run crashes the app process at exit (GGML_ASSERT in ggml_metal_rsets_free)
## Problem

Any `xcodebuild test` invocation of `MilaTests` reliably crashes the hosting `Mila` process at exit with:

```
GGML_ASSERT([rsets->data count] == 0) failed
```

in `ggml_metal_rsets_free` (`ggml-metal-device.m:612`), reached via `abort()` -> `ggml_abort` -> `~vector<unique_ptr<ggml_metal_device>>` -> `__cxa_finalize_ranges` -> `exit()` -> `_XCTestMain`.

Test results themselves are unaffected (xcodebuild reports pass/fail correctly before the crash), but it litters `~/Library/Logs/DiagnosticReports/` with a crash report on every test run and risks masking a real crash if CI/tooling ever alerts on crash-report count.

## Root cause

`MilaTests` is an app-hosted unit test bundle (`TEST_HOST` = `Mila.app`, `project.yml`), so every unit test run boots the real app, including the `.task { prewarmDefaultModel() }` launch hook, which creates a real whisper.cpp/Metal (`ggml_metal_device`) context.

`prewarmDefaultModel()`'s existing test guard (`isRunningUITests`, keyed off the `--uitests` launch argument) only covers XCUITest — it does not catch this app-hosted *unit* test case, since no such launch argument is passed there. So prewarm runs for real under plain unit tests too.

When the test run finishes, `XCTestMain` calls `exit()` directly, bypassing `NSApp.terminate()` -> `applicationShouldTerminate` -> `MilaAppDelegate.gracefulShutdown()` — the method that exists specifically to free the whisper/Metal context safely before libc++ static destructors run (see its doc comment, which already documents this exact crash from real end-user quit flows). Since `gracefulShutdown()` never runs here, the async ggml-metal rsets-init dispatch block races the static destructors and the assert fires.

## Fix

See linked PR — broadens the test-hosting guard (renamed `isRunningUnderXCTest`) to also detect `NSClassFromString("XCTestCase") != nil`, which is reliably true for app-hosted unit tests. (`XCTestConfigurationFilePath`, the usual env-var signal, was checked empirically and found present-but-empty on this toolchain for this hosting mode, so it isn't usable here.)

### LINKED ISSUE #268: ModelManager tests write to the real app's UserDefaults domain instead of an isolated suite
## Problem

`ModelManager` reads/writes `selectedModelName` and `model.declinedNames` via `UserDefaults.standard` directly, unlike other settings models in this codebase (`DiarizationSettings`, `HotkeySettings`), which accept an injectable `UserDefaults` instance per this repo's own `CLAUDE.md` convention:

> For verification/setup state that should survive app restarts, persist a `verified` flag alongside the verified parameter values

and the "Environment Objects" / DI pattern used throughout `MilaTests`.

Because of this, any test that constructs its own `ModelManager` and calls `delete(_:)` / `setSelected(_:)` / `download(_:)` — `ModelManagerTests`, `DictationControllerTests`, `TranscriptionServiceTests` — was writing directly into the real, on-disk `io.island.whisper.IslandWhisper` preferences domain (the same one the actually-installed Mila.app on a developer's machine uses), instead of an isolated store. `ModelManagerTests` worked around this with a save/restore-around-`.standard` hack in `setUp`/`tearDown` rather than real isolation, and the other two files did not guard against it at all.

Confirmed by direct observation: after running the affected test suites, `defaults read io.island.whisper.IslandWhisper model.declinedNames` and `selectedModelName` showed values written by the test run.

## Fix

See linked PR — added an injectable `defaults: UserDefaults = .standard` parameter to `ModelManager`, and updated the three affected test files to pass an isolated `UserDefaults(suiteName:)` per the `HotkeySettingsTests` pattern.

celarent7:
## Summary

- Persist a `declinedModelNames` set in `ModelManager` so deleting a model via Settings sticks across relaunch instead of silently re-downloading (Closes #263).
- Guard `ensureDefaultModelsInstalled()`'s `setSelected(ivritLarge)` call against the declined state too, so a deleted model's selection doesn't come back on every launch either (Closes #266).
- Fix a `GGML_ASSERT` crash at process exit on every unit test run: `prewarmDefaultModel()`'s test-hosting guard only checked the `--uitests` XCUITest flag, missing that `MilaTests` is an app-hosted unit test bundle that also runs launch-time `.task`s for real (Closes #267).
- Give `ModelManager` an injectable `UserDefaults`, matching the `DiarizationSettings`/`HotkeySettings` DI pattern, so tests stop writing into the real app's on-disk preferences domain (Closes #268).

## Test plan

- [x] `MilaTests/ModelManagerTests` (14/14 pass)
- [x] `MilaTests/DictationControllerTests` (3/3 pass)
- [x] `MilaTests/TranscriptionServiceTests` (39/39 pass)
- [x] Confirmed no `GGML_ASSERT`/crash report and no `prewarm: kicking off` log line after the fix
- [x] Confirmed `defaults read io.island.whisper.IslandWhisper model.declinedNames` no longer gets written by test runs

https://claude.ai/code/session_01FNNpYS4agJ1gmgHjMRSFWt

---

> [!NOTE]
> **Medium Risk**
> Changes first-launch model bootstrap and persisted selection/download behavior for users who delete models; risk is mitigated by unit tests and XCTest guards, but regressions could affect onboarding or model settings.
> 
> **Overview**
> **Deleted models stay deleted.** `ModelManager` now persists a `declinedModelNames` set (`model.declinedNames` in `UserDefaults`): `delete(_:)` marks a model declined, `download(_:)` clears that flag, and launch-time `ensureDefaultModelsInstalled()` skips auto-download and the unconditional `setSelected(ivritLarge)` when a default model was declined.
> 
> **App-hosted unit tests no longer run production launch work.** `isRunningUITests` becomes `isRunningUnderXCTest`, also true when `XCTestCase` is loaded (`MilaTests` as `TEST_HOST`). That guard skips `prewarmDefaultModel()` and `ensureDefaultModelsInstalled()`, avoiding Metal/whisper teardown noise and mutations to real preferences or network during `xcodebuild test`.
> 
> **Tests isolate `ModelManager` prefs and downloads.** `ModelManager` accepts injectable `UserDefaults` and optional `URLSession` protocol classes; several test suites use dedicated suites instead of `UserDefaults.standard`, with new coverage for declined-state persistence and download clearing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38f580b1495e74fc7da1c5cb2fa507a440329f29. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleted transcription models are no longer automatically selected or downloaded again.
  - Model deletion preferences now persist between app launches.
  - Downloading a previously deleted model explicitly restores it and clears its declined status.
  - Declining one model does not affect the availability or download behavior of other models.

### DIFF
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -2044,7 +2044,9 @@ struct MilaApp: App {
         guard !remoteTranscriptionSettings.isActive else { return }
         modelManager.setSelected(WhisperModel.ivritLarge)
         for model in [WhisperModel.ivritLarge, WhisperModel.openaiTurbo] {
-            if !modelManager.isInstalled(model) && modelManager.downloads[model.name] == nil {
+            if !modelManager.isInstalled(model)
+                && !modelManager.isDeclined(model)
+                && modelManager.downloads[model.name] == nil {
                 modelManager.download(model)
             }
         }
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -2050,6 +2050,15 @@ struct MilaApp: App {
     /// same `URLSession` so they don't actually saturate the network, and
     /// the in-app banner shows whichever is currently selected.
     private func ensureDefaultModelsInstalled() {
+        // Skip entirely under XCTest hosting (unit or UI) — same rationale
+        // as `prewarmDefaultModel()`: `MilaTests` is app-hosted, so this
+        // `.task` runs for real during ordinary unit test runs too. Without
+        // this guard it calls `setSelected`/`download` on the production
+        // `ModelManager`, which is backed by `UserDefaults.standard` (see
+        // `MilaApp.body`'s `ModelManager(modelsDirectory:)` call) — every
+        // `xcodebuild test` invocation would mutate the real app's declined
+        // models / selected model and could kick off real network downloads.
+        guard !isRunningUnderXCTest else { return }
         // Skip the multi-GB local model downloads + CoreML prep entirely for
         // users who've opted into the remote backend — they don't need any
         // local weights. (The Models tab still offers manual downloads, and
--- a/Mila/App/MilaApp.swift
+++ b/Mila/App/MilaApp.swift
@@ -2011,23 +2011,36 @@ struct MilaApp: App {
     /// button, and any too-eager Record press during the window finds
     /// the model already loaded.
     private func prewarmDefaultModel() {
-        // Skip prewarm under UI tests. No UI test relies on a launch-time
-        // warm — the transcription E2E tests pass `--ui-test-tiny-model-path=`
-        // and load on demand — so warming the (potentially multi-GB) default
-        // model here just burns CPU/IO on every UI-test launch and creates a
-        // whisper/Metal context that XCTest never frees (it exits via
-        // `exit()`, bypassing `gracefulShutdown()`), which is what prints the
-        // ggml-metal teardown backtrace at process exit.
-        guard !isRunningUITests else { return }
+        // Skip prewarm under any XCTest hosting (UI or unit tests). No test
+        // relies on a launch-time warm — the transcription E2E tests pass
+        // `--ui-test-tiny-model-path=` and load on demand — so warming the
+        // (potentially multi-GB) default model here just burns CPU/IO on
+        // every test launch and creates a whisper/Metal context that XCTest
+        // never frees (it exits via `exit()`, bypassing `gracefulShutdown()`),
+        // which is what prints the ggml-metal teardown backtrace at process
+        // exit. This used to be guarded by `--uitests` alone, which covers
+        // XCUITest, but `MilaTests` is an app-hosted unit test bundle
+        // (`TEST_HOST` = Mila.app in project.yml) — SwiftUI's `.task`
+        // modifiers, prewarm included, run there too, just without that flag.
+        guard !isRunningUnderXCTest else { return }
         // No local weights to warm when the remote backend is selected.
         guard !remoteTranscriptionSettings.isActive else { return }
         transcription.prewarm(language: languageSettings.current.rawValue)
     }
 
-    /// True when the process was launched by an XCUITest run. Keyed off the
-    /// `--uitests` / `--ui-test-*` launch arguments the UI test targets pass.
-    private var isRunningUITests: Bool {
+    /// True when the process is hosting an XCTest run — either an XCUITest
+    /// (keyed off the `--uitests` / `--ui-test-*` launch arguments the UI
+    /// test targets pass) or an app-hosted unit test bundle (`MilaTests`),
+    /// which loads `XCTestCase` directly into this process via
+    /// `BUNDLE_LOADER`/`TEST_HOST` — no launch argument marks that case, so
+    /// `XCTestCase`'s presence is the only reliable signal.
+    /// `XCTestConfigurationFilePath`, the usual env-var check for "is this
+    /// process under XCTest," is present but empty for app-hosted unit
+    /// tests on this toolchain — verified empirically, not a documented
+    /// contract — so it can't be used here.
+    private var isRunningUnderXCTest: Bool {
         CommandLine.arguments.contains { $0 == "--uitests" || $0.hasPrefix("--ui-test") }
+            || NSClassFromString("XCTestCase") != nil
     }
 
     /// Pre-download the two default models on first launch:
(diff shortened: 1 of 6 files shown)


urisland:
Thank you for this, @celarent7 — this was a genuinely good piece of work, and most of what ships is yours.

Relayed to **#276** on an `island-io` branch. That is purely mechanical: every workflow here is stuck at `action_required` (outside-contributor approval), so this PR has never been through our CI and cannot get there from a fork. Your five code commits are cherry-picked unchanged and still carry you as author — `git log --format='%an'` on #276 shows that — and the squash commit will carry `Co-Authored-By` so the attribution survives the merge.

What stood out:

**The `NSClassFromString("XCTestCase") != nil` detection in #267.** This is the good find. The `--uitests` flag looked like the test guard, and the natural conclusion from "prewarm ran anyway" is that something is wrong with the flag. Instead you worked out that `MilaTests` is app-hosted (`TEST_HOST` = Mila.app), that SwiftUI's launch `.task`s therefore run for real under plain unit tests, and that *nothing at all marks that case* — so the only reliable signal is the test framework being loaded into the process. That is the correct diagnosis of a crash that had been dismissed as test-runner noise.

**And you wrote down the negative result.** Recording in the code comment that `XCTestConfigurationFilePath` — the obvious env-var check, the one every StackOverflow answer reaches for — was checked empirically and found *present but empty* for this hosting mode, and flagging it as empirical rather than a documented contract, is the part that saves the next person a wasted afternoon. Findings like that almost never get written down. Please keep doing it.

**The declined-state design is right at the seam.** Keying per model so declining one cannot affect the other, and treating `download(_:)` as the thing that clears it — an explicit download request *is* the user asking for the model back, wherever it comes from — means the state can't drift from intent. `lastDownloadErrors` in the same file had already learned the per-model lesson the hard way; you got there first time.

Two things to flag, both in the open:

- **#276 also carries #264 and #265**, which were deliberately left out here. Both were harmless *because of* the bug you fixed — the model came back on the next launch, which quietly repaired the orphaned 1.17 GB CoreML encoder and the deleted-both-models dead end alike. Once deletion sticks, both become permanent, and a stable release is being cut immediately after this lands. So it ships as one PR covering all six. Your #266 guard was the right first half of the selection fix; #276 adds the rest (only establish a default when there is no persisted selection at all).
- **`docs/plans/2026-09-04-sticky-model-deletion.md` is not relayed.** It predates your last three commits and the shipped code contradicts it where it matters — it specifies `UserDefaults.standard` with a save/restore hack and "no new `UserDefaults` suite", which is the opposite of the DI you ended up building for #268. A plan in `docs/plans/` describing the design that was rejected would mislead the next reader. The reasoning is preserved in the commit messages and in #276's description. No reflection on the work; the code moved past the plan, which is what should happen.

One change to your code itself: the `lazy var session` initializer now says `self.sessionProtocolClasses` explicitly, since the same closure already needs `self` for `delegate: self` and I had no toolchain to settle the implicit-`self` question locally.

Closing in favour of #276.

---

Issue #280, opened 2026-09-07
Recording pins the main thread at 60–75% CPU: RecordingSession level/elapsed publishes re-evaluate the App scene


urisland:
## Symptom

Mila sits at **60–75% CPU for the whole of a recording** even with the remote transcription backend selected — i.e. with nothing local to transcribe. Two independent observations:

- A user's diagnostic report (1.9.3-beta.4, macOS 26.6.2, Mac15,10): a 71-minute mic recording on the remote backend, "CPU at 100%".
- Reproduced live on 1.9.3 (build 62), meeting-mode recording on the remote backend: `top` shows `Mila` at 61–75% steadily, plus the diarization daemon `python3.11` at ~16% (expected — that one is the live speaker-embedding work).

## What the process sample shows

A 5 s `sample` of the recording Mila puts **~81% of main-thread samples inside SwiftUI's view-graph update** (`NSHostingView.beginTransaction` → `ViewGraphRootValueUpdater.updateGraph` → `GraphHost.flushTransactions` → `AG::Subgraph::update`). Underneath it, per pass:

- `AppGraph.sceneList(namespace:flush:)` and `AppKitWindowController.updateRootView(...)` — the **App scene** is being re-evaluated, and `MilaApp.body.getter` is in the stacks;
- `OutlineListCoordinator.update(...)` / `outlineView(_:configureCell:for:)` — the sidebar's NSOutlineView reconfiguring;
- `AppDelegate.makeMainMenu(commandsList:)` and `PlatformItemList.Item.update` — the **main menu** being rebuilt.

Nothing audio-, VAD-, network- or diarizer-related registers; those threads are parked in `read` / `__workq_kernreturn`.

## Root cause

`RecordingSession` is one of ~35 `@StateObject`s on `MilaApp`. A `@StateObject` subscribes the owning body to `objectWillChange` — for an `App`, that means **every publish re-evaluates `MilaApp.body` and re-diffs every scene** (window root view, `.commands`, `Settings`). The App body reads none of the session's properties, so all of that work is wasted.

`RecordingSession` publishes at audio-buffer cadence:

- `micLevel` — once per mic tap buffer (4096 frames @ 48 kHz ≈ 12 Hz),
- `systemLevel` — once per ScreenCaptureKit audio buffer (meeting mode, up to ~50 Hz),
- `elapsed` — 5 Hz.

[...]

PR #284, opened 2026-09-07
test(ci): count App-scene re-evaluations during a recording so a per-buffer @StateObject publish fails the build

### LINKED ISSUE #283: CI: count App-scene re-evaluations during a recording so a per-buffer @StateObject publish fails the build
## Problem

#280 — a `@StateObject` on `MilaApp` publishing once per audio buffer, so the App scene re-diffed at ~12–50 Hz and a recording pinned the main thread at 60–75% — shipped through every existing check. Nothing in the pipeline logs "the App body ran", the unit tests exercise objects in isolation (where a publish is free), and the UI E2Es assert content, not cost. The regression class is invisible until a user opens Activity Monitor.

We need CI to catch the next one, and it needs to be **as deterministic as possible**: a CPU-percentage threshold on a shared macOS runner is a coin toss (the same work reads 8% idle and 30% under load), which is exactly the kind of flake that gets retried past.

## Proposal

Count, don't time.

- **`AppSceneChurnProbe`** (DEBUG-only, no-op shell in release): `MilaApp.body` bumps a counter on every evaluation, and `MilaApp.init` registers every App-level `@StateObject` by property name with a type-erased `objectWillChange`. New App-level objects are picked up by pattern, no per-object wiring.
- **`AppSceneChurnTests`** (app-hosted `MilaTests`, gated on `MILA_APP_CHURN_E2E=1`): starts a fake recording on the host app's real `QuickActionsController`, pumps 12 s of the English speech fixture through the real `RecordingSession.consumeMic` path at real time (so a 5 Hz timer regression reads as 60 publishes, not 5), then asserts:
  1. no App-level object published more than 4× per second of recording — and names the one that did;
  2. `RecordingSession` published 0 times mid-pump (state transitions only);
  3. `MilaApp.body` evaluated within the same budget;
  4. main-thread CPU fraction < 50% as a *backstop* only (bug: 60–75%; fix: single digits), with the failure message pointing at the counts.
- **`QuickActionsController.discardFakeRecordingForTesting()`**: ends the fake recording through `session.cancelAll()` — the teardown `wireLiveAIPipeline` already handles for sleep/lock/quit — so the test writes nothing to the host's store and shows no rename sheet.
- CI-only because the host app is the developer's real one: on a dev machine the same run would boot their diarization runtime and fire their configured LLM. `unit-and-ui-tests` sets the gate.

Follow-up to #280 / #282.

urisland:
Closes #283

## What & why

#280 shipped through every existing check: a `@StateObject` on `MilaApp` published once per audio buffer, the App scene re-diffed at ~12–50 Hz, and a recording pinned the main thread at 60–75% for its whole duration. Nothing in the pipeline logs "the App body ran", unit tests exercise objects in isolation (where a publish is free), and the UI E2Es assert content, not cost. This adds a CI check for that class of regression — and keeps it **count-based**, because a CPU-percentage threshold on a shared macOS runner is a coin toss.

- **`AppSceneChurnProbe`** (`Mila/App/`, DEBUG-only — an empty shell in release): `MilaApp.body` bumps a counter on every evaluation; `MilaApp.init` registers every App-level `@StateObject` by property name with a type-erased `objectWillChange`. New App-level objects only need a line in that list.
- **`AppSceneChurnTests`** (app-hosted `MilaTests`, gated on `MILA_APP_CHURN_E2E=1`): starts a fake recording on the host app's *real* `QuickActionsController`, pumps 12 s of `en_meeting_notes.wav` through the real `RecordingSession.consumeMic` path at real time (so a 5 Hz timer regression reads as 60 publishes, not 5), then asserts:
  1. no App-level object published more than 4× per second of recording — the failure names it;
  2. `RecordingSession` published 0 times mid-pump (state transitions only);
  3. `MilaApp.body` evaluated within the same budget;
  4. main-thread CPU fraction < 50% as a backstop only (bug: 60–75%; fix: single digits), with the message pointing at the counts.
- **`QuickActionsController.discardFakeRecordingForTesting()`**: ends the fake recording through `session.cancelAll()` — the teardown `wireLiveAIPipeline` already runs for sleep/lock/quit — so the test writes nothing to the host's store and shows no rename sheet.
- **CI-only on purpose.** The host app is the developer's real one: on a dev machine the same run would boot their diarization runtime and fire their configured LLM for 12 s. `unit-and-ui-tests` sets the gate; `make test` skips it.

## How it was tested

- `make project` + `xcodebuild build-for-testing`: app and both test bundles compile (the probe's registry is built from the init-time locals, so a missing name fails the build).
- The test itself was **not run locally** (it is CI-gated for the reason above; and running app-hosted tests here would launch a second Mila while the reporter's own was recording). It runs in this PR's `unit-and-ui-tests` job — please read that job's log for the `AppSceneChurn:` line, which prints the per-object counts and the measured CPU fraction so the budgets can be judged against a real runner.
- Expected on `main` with #282: `session=0`, body evaluations in the low tens for 12 s, CPU fraction well under 0.5. Expected on the pre-#282 tree: `session≈140` and the test fails naming `session`.

## Checklist

- [x] Builds locally (`make build`) and tests pass (`make test`) — compile verified; the new test is CI-gated and runs in this PR's checks
- [ ] User-facing change is noted in the README changelog (if applicable) — not user-facing

---

> [!NOTE]
> **Low Risk**
> Test and DEBUG-only instrumentation plus CI enforcement; production behavior changes only the XCTest-hosted Live AI capability override so app-hosted tests can wire the live pipeline.
> 
> **Overview**
> Adds a **CI-only regression guard** for the #280 class of bug: an App-level `@StateObject` publishing on every audio buffer, forcing `MilaApp.body` to re-diff the whole scene at mic cadence.
> 
> **`AppSceneChurnProbe`** (DEBUG-only shell in release) counts `MilaApp.body` evaluations and registers each App-level `@StateObject` by name with type-erased `objectWillChange`. **`AppSceneChurnTests`** runs inside the real app host when `MILA_APP_CHURN_E2E=1`: fake recording via `QuickActionsController`, 12 s of real-time fixture audio through `RecordingSession.consumeMic`, then asserts per-object publish counts (≤4/s), zero mid-pump `session` publishes, body-eval budget, and a loose main-thread CPU backstop. **`discardFakeRecordingForTesting()`** tears down via `cancelAll()` without saving or finalizing.
> 
> **`MilaApp`** registers probe targets in DEBUG, calls `noteBodyEvaluation()` from `body`, and treats XCTest-hosted runs like UI tests for the Live AI hardware gate so the live pipeline wires on Air-class CI runners.
> 
> **GitHub Actions** sets both `MILA_APP_CHURN_E2E` and `TEST_RUNNER_*`, then **fails the job** if the churn test did not **pass** (so an env-gated `XCTSkip` cannot green-wash), and posts the `AppSceneChurn:` report to the step summary. **`CLAUDE.md`** documents the test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fb4d5a5ddcaae089e06c1528d264e81b3fc2a80. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/ios-tests.yml
+++ b/.github/workflows/ios-tests.yml
@@ -78,6 +78,14 @@ jobs:
             -resolvePackageDependencies
 
       - name: Run unit tests
+        env:
+          # Opts in MilaTests/AppSceneChurnTests: it drives the HOST APP's real
+          # recording controller for ~12 s of fixture audio and asserts a
+          # publish budget per App-level object (the #280 class of bug — a
+          # @StateObject on MilaApp publishing per audio buffer). CI-only
+          # because on a developer's machine the same run would boot their
+          # diarization runtime and fire their configured LLM.
+          MILA_APP_CHURN_E2E: "1"
         run: |
           set -o pipefail
           # `-retry-tests-on-failure -test-iterations 3`: the subprocess-spawning
--- a/.github/workflows/ios-tests.yml
+++ b/.github/workflows/ios-tests.yml
@@ -84,8 +84,14 @@ jobs:
           # publish budget per App-level object (the #280 class of bug — a
           # @StateObject on MilaApp publishing per audio buffer). CI-only
           # because on a developer's machine the same run would boot their
-          # diarization runtime and fire their configured LLM.
+          # diarization runtime and fire their configured LLM. Set in both
+          # forms: xcodebuild forwards `TEST_RUNNER_`-prefixed vars into the
+          # test process with the prefix stripped, and the plain form is what
+          # the audio-loopback job relies on for the same bundle — the step
+          # after the run checks the test actually PASSED rather than skipped,
+          # so a delivery regression cannot hide behind a green job.
           MILA_APP_CHURN_E2E: "1"
+          TEST_RUNNER_MILA_APP_CHURN_E2E: "1"
         run: |
           set -o pipefail
           # `-retry-tests-on-failure -test-iterations 3`: the subprocess-spawning
@@ -120,6 +126,32 @@ jobs:
             -default-test-execution-time-allowance 240 \
             -maximum-test-execution-time-allowance 480 \
             test 2>&1 | tee build/unit-tests.log | tail -200
+
+      # The App-scene churn budget (#283) is gated on an env var, and a gated
+      # test that XCTSkips leaves the job green — which is exactly how a
+      # per-buffer @StateObject publish (#280) would ship again. Require the
+      # test to have PASSED, and surface its one-line report (per-object
+      # publish counts + main-thread CPU fraction) so the budgets can be read
+      # against a real runner without downloading the log.
+      - name: Verify the App-scene churn test ran
+        run: |
+          log=build/unit-tests.log
+          line=$(grep -E "^Test Case '-\[MilaTests\.AppSceneChurnTests test_live_recording_stays_within_the_app_scene_publish_budget\]' (passed|failed|skipped)" "$log" | tail -1 || true)
+          report=$(grep -E "^AppSceneChurn: " "$log" | tail -1 || true)
+          echo "result: ${line:-<not found>}"
+          echo "report: ${report:-<none>}"
+          {
+            echo "### App-scene churn budget (#283)"
+            echo
+            echo '```'
+            echo "${line:-test case line not found in log}"
+            echo "${report:-no AppSceneChurn: report line}"
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+          case "$line" in
+            *"' passed"*) ;;
+            *) echo "::error::AppSceneChurnTests did not pass (skipped or missing) — is MILA_APP_CHURN_E2E reaching the test host?"; exit 1 ;;
+          esac
 
       # Issue #208: with `-retry-tests-on-failure`, a green job can sit on top
       # of a pile of failed executions — one run logged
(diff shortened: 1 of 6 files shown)


urisland:
Bugbot's three findings, checked against the code and addressed in 0fb4d5a:

1. **CI env var never reaches the test** — plausible: the repo itself uses both conventions for this bundle (`MILA_PIPELINE_E2E` plain in the loopback job, `TEST_RUNNER_MILA_COREML_E2E` in ci.yml). Rather than pick one, the gate is now set in both forms, and a new step requires the test case line to read `passed` (a skip or a missing line fails the job) and prints the `AppSceneChurn:` report into the step summary. That step is the real fix: the silent-skip trap is what would let #280 ship again.
2. **Live pipeline skipped on Air runners** — confirmed: `wireLiveAIPipeline`'s `.recording` branch gates on `isLiveAIAvailable`, and only the `--ui-test-*` launch args forced the non-Air capability. The override now also applies when `XCTestCase` is loaded in the process (app-hosted unit tests); XCUITest app processes do not load it, so UI-test behaviour is unchanged. The test skips with a reason instead of failing if the pipeline still isn't wired — and the step above turns that skip red.
3. **Discard returns before pipeline teardown** — confirmed: `cancelAll()` only publishes `.idle`; the observer then awaits `transcribeNow()` and the diarizer drain. The seam now waits (bounded, ≤10 s) for the observer's last act, clearing `session.onLiveSamples`, before returning.

